PR #485, opened 2015-03-25
Avoid hardcoded wine64 path in msxml6 recipe.

slackner:
Commit https://github.com/Winetricks/winetricks/commit/84906b01114330595c37abbeba89c0b450df9a98 introduced a hardcoded `wine64` path, which would break the installation if a non-default wine version is used.

Unfortunately the `msxml6.dll` recipe needs a 64-bit `msiexec` to work properly. We do that by spawning 32-bit `start`, which then starts 64-bit `msiexec` (using the `C:\Windows\sysnative` path). This only works as long as the `$WINE` environment variable really specifies the 32-bit wine version, but I assume there are a lot more recipes depending on that... and its better than doing path manipulations! ;)

### DIFF
--- a/src/winetricks
+++ b/src/winetricks
@@ -355,6 +355,16 @@ w_try_7z()
     w_try 7z "$@"
 }
 
+w_try_msiexec64()
+{
+    if test "$W_ARCH" != "win64"
+    then
+        w_die "bug: 64-bit msiexec called from a $W_ARCH prefix."
+    fi
+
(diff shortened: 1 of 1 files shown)


slackner:
v2: Move code to execute 64-bit msiexec into a helper function.

abzalov:
Thanks, so much better. I did not know how to run 64-bit msiexec.

austin987:
Merged, thanks for working on this @slackner.

PR #488, opened 2015-03-28
winetricks: add binkw32 verb

slackner:
Some games do not ship a binkw32.dll (somehow required for video rendering?), having a winetricks recipe for that makes it easier to test such bugs. One example is: https://bugs.wine-staging.com/show_bug.cgi?id=155

Some comments on the patch:

* The changes to `w_download_to` are required since the webserver is not happy with winetricks passing the accept-encoding header, and wget returns a gzip compressed result. The replacement `*gzip*` -> `*:*gzip*` was done to make sure it doesn't accidentially match the filename instead of the file type. **Needs testing on MacOS/BSD!**

* Metadata information is based on the file metadata.

* I know that `down-dll.com` is probably not really the best mirror, but the checksum matches the expected one, and it doesn't require any cookies. If it breaks we might have to find a better alternative, but that could happen with any mirror, so ...

### DIFF
--- a/src/winetricks
+++ b/src/winetricks
@@ -707,9 +707,9 @@ w_download_to()
         # Also affects ttf files on github
         _W_filetype=`which file 2>/dev/null`
         case $_W_filetype-$_W_file in
-        /*-*.exe|/*-*.ttf)
+        /*-*.exe|/*-*.ttf|/*-*.zip)
             case `file "$_W_file"` in
-            *gzip*) mv "$_W_file" "$_W_file.gz"; gunzip < "$_W_file.gz" > "$_W_file";;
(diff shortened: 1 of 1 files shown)


austin987:
@dankegel

slackner:
I have tested that the output of the `file` command is identical on FreeBSD, and the changes works as expected. Austin also asked me to run the `tests/winetricks-test` script, but it seems like it is not compatible with FreeBSD yet.... lot of errors like:

```
time: illegal option -- -
usage: time [-al] [-h | -p] [-o file] utility [argument ...]
sort -k -k3.1: Invalid argument
...
```

All tests are counted as failures, although they work fine when executed manually directly with winetricks. I've tested a couple of recipes which download `.zip` files, and they still work fine.

PR #491, opened 2015-03-29
physx: Update to version 9.14.0702

ghost:
Now should be fine.


austin987:
You now have two commits, 71f0350 which has the change (and changes the version), then ed86af7 reverting that change.

You want one commit, without WINETRICKS_VERSION being changed (the copyright change is fine, of course).

austin987:
https://github.com/Winetricks/winetricks/commit/178813ea6b4297a45cfec3e332a868075d0369df

PR #498, opened 2015-04-21
Fix a broken test in non-bash shells

ScoreUnder:
```
$ winetricks dotnet45
Executing w_do_call dotnet45
/usr/bin/winetricks: 1968: test: dotnet20: unexpected operator
error: dotnet45 conflicts with dotnet20sp2, which is already installed.
```
It doesn't appear to have any adverse affects (it is reporting the conflict correctly), but the test fails. I have modified the check to use the exit code of grep -q (indicating the presence of matches), rather than testing for matches using test.

### DIFF
--- a/src/winetricks
+++ b/src/winetricks
@@ -2037,7 +2037,7 @@ w_conflicts()
 {
     for x in $2
     do
-        if test $(grep -w "$x" "$WINEPREFIX/winetricks.log")
+        if grep -qw "$x" "$WINEPREFIX/winetricks.log"
         then
             w_die "error: $1 conflicts with $x, which is already installed."
         fi


ScoreUnder:
As an aside, there are a lot of suspicious unquoted variables around the code. I don't know how worried I should be as I'm not very familiar with the code base.

austin987:
What shell are you using?

ScoreUnder:
dash

ScoreUnder:
For the record, the source of the problem:

```
score@kirisame ~ % grep -w dotnet20 .osu-wine/winetricks.log 
dotnet20
dotnet20
```

So it ends up calling "test dotnet20 dotnet20" (since the process substitution is un-quoted, it is subject to word splitting and this confuses "test"). Grep's exit code indicates whether or not a match was found, so the use of a bare "grep -q" was mostly because I thought it was more elegant than just putting quotes around the process substitution.

austin987:
Thanks for the patch / sorry for the delay in accepting. I wanted to try reproducing myself, but was unable to in Fedora's dash (0.5.8-1.fc22), or ksh (20120801-24.fc22) or mksh (50f-1.fc22).

In any case, the patch works and simplifies the code, and fixes things for you, so no reason not to commit :)

ScoreUnder:
Sorry if it was difficult to test - it could be reproduced by duplicating lines in winetricks.log, but I don't know how it got like that in the first place.

austin987:
Ah, yes, now I can reproduce it, thanks for the tip.

PR #501, opened 2015-05-26
dotnet20 should now install on 64bit systems

grepwood:


grepwood:
Update: you can pull, but it still doesn't work. The installer thinks that dotnet20 is already installed. At least it's something.

austin987:
There's not really a reason to merge this hack if the installer doesn't work.

Until upstream is fixed, I don't see this changing anytime soon.

PR #507, opened 2015-06-24
Updated steam sha1sum

ruseman:
I updated the checksum used for SteamInstall.exe.  I'd consider this one much more important, as 'winetricks steam' breaks without it.


austin987:
FYI, you should also add a comment with the date of the new sha1sum. I've fixed this in commit fb44775ff4e1dd56ad928ee3cdbded492824fa49

ruseman:
Ok, I'll be sure to do that in the future

PR #528, opened 2015-08-02
Don't delete CACHE and reuse same folder as winetricks

davispuh:
There's really no reason to delete cache as winetricks will verify that checksums match and redownload if not. Also reuse same folder as winetricks itself, there's no point in keeping multiple copies of files around and you can always manually delete cache if there's any need for it.

### DIFF
--- a/tests/winetricks-test
+++ b/tests/winetricks-test
@@ -3,8 +3,8 @@
 
 # Override this if you want to put the work area on a different disk
 WINE_PREFIXES=${WINE_PREFIXES:-$HOME/winetrickstest-prefixes}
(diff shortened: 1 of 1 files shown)


austin987:
The use of a different cache is intentional, the tests are meant to catch not only winetricks problems, but broken/changed URLs. Reusing the existing cache defeats that purpose.

davispuh:
Then I suggest another test, like `winetricks-test downloads` which would test if URLs have been changed, etc.
It's much more common to run test multiple times to check if different code changes didn't break something and redownloading everything every time is stupid.

austin987:
Sure, patches welcome.

PR #533, opened 2015-08-08
Add download tests and delete cache only for those

davispuh:
This is basically update to #528 
Implementation isn't very pretty because had to copy some code from winetricks*, but it works fine, will test all links  (220 + 40 manual) and no need to run wine tests itself to check downloads.

\*I recommend splitting winetricks in multiple files and in makefile concatenate it together so it's more manageable, like apps/dlls in different files and could reuse code, wouldn't need to copy like this...

### DIFF
--- a/src/winetricks
+++ b/src/winetricks
@@ -11257,7 +11257,7 @@ w_metadata algodoo_demo games \
 
 load_algodoo_demo()
 {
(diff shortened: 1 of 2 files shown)


austin987:
I merged 7c19b6e and 5778a1f, thanks. I'm going to write a different solution for download tests (using check and test goals, with check only running the quick tests and test running all the tests, with a wiped cache).

PR #536, opened 2015-08-10
Add "Managed DirectX" verb

kakurasan:
Install "Managed DirectX" which is required by some .NET applications (e.g. Bve trainsim).
This can be installed by "winetricks directx9", but other DLLs are also installed/overridden.

Note: Bve trainsim also requires "dotnet35" and "d3dx9_36".

### DIFF
--- a/src/winetricks
+++ b/src/winetricks
@@ -6490,6 +6490,86 @@ load_mdac28()
 
 #----------------------------------------------------------------
 
+w_metadata mdx dlls \
+    title="Managed DirectX" \
+    publisher="Microsoft" \
+    year="2006" \
+    media="download" \
+    file1="../directx9/directx_feb2010_redist.exe" \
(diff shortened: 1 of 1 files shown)


austin987:
Have you tested this with Bve trainsim/any other games? Given that your recipe differs from Focht's, I'm curious how well it works with different .Net versions (looks like it should, but it's nice to know ;)).

kakurasan:
At least, Bve trainsim (.NET 3.5) and Gray Matter Demo configuration tool (.NET 2.0) work. Unfortunately, I don't know/have any other programs that use Managed DirectX.

Note: Managed DirectX (1.1) is not compatible with .NET 4.0 - http://blogs.msdn.com/b/chuckw/archive/2010/12/09/directx-and-net.aspx

austin987:
Two versions is good for me, thanks for the patch and for testing.

PR #537, opened 2015-08-11
Replace 'w_override_dlls native' with 'w_override_dlls native,builtin' where possible.

slackner:
So far Winetricks recipes often install only 32-bit libs. Using the first method for
overriding builtin dlls effectly disables the library on x86_64, whereas the second
method should work as expected (native on 32-bit, builtin as fallback on 64-bit).

A simple example code to load an overriden library returns `0x0` before this change
on x86_64, and the correct address of the builtin dll after the change.

At two places there was a comment that using `native,builtin` would not work.
I decided to skip replacing there, until it has been manually checked, if this is (still) correct.

### DIFF
--- a/src/winetricks
+++ b/src/winetricks
@@ -4270,7 +4270,7 @@ helper_d3dx9_xx()
       w_try_cabextract -d "$W_SYSTEM32_DLLS" -L -F "$dllname.dll" "$x"
     done
 
-    w_override_dlls native $dllname
+    w_override_dlls native,builtin $dllname
 }
 
 helper_win2ksp4()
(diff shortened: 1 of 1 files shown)


qwertychouskie:
Has this PR been forgotten?

austin987:
I had provided feedback to @slackner on IRC, but it wasn't included here.

'winetricks -q -v --verify dotnet40' is broken by this patch. There may be other failures, as well that need to be checked.

qwertychouskie:
It seems @slackner has abandoned this PR, as there have been no commits since Aug. 10, 2015, and no other indication that he is working on the problems.

PR #539, opened 2015-08-15
d3dx10: Install d3dx10_43.dll

kakurasan:
"February 2010" redistributable package doesn't contain d3dx10_43.dll.
Use "June 2010" version instead.


austin987:
Thanks.

Issue #543, opened 2015-09-09
Please, update Picasa checksum


rodoviario:
There are new versions of Picasa39 released by Google that takes in account the new Google Photos. The checksum doesn't match anymore.

If you try to install the new picasa39.exe using wine it works very well (even in the about box says Picasa linux). The only missing thing is the option to connect to your google account.

PR #544, opened 2015-09-11
Update sha1sum for picasa39 per #543

ShannonVanWagner:


austin987:
Please put the previous sha1sum as a comment with the previous date and the new sha1sum and the current date like we do for other verbs (flash, etc.).

PR #547, opened 2015-09-21
Update sha1sum for picasa39 for #543

ShannonVanWagner:
I've added the old sha1sum into a comment, please review and accept if you like it! Thanks!


austin987:
Merged with minor edits, thanks.

PR #557, opened 2015-10-09
More apps support

hillwoodroc:


austin987:
I'd like to avoid making winetricks a huge installer for every application under the sun. Dlls are fine, as they replace missing/incomplete parts of wine. Applications are added on a case by case basis, when they're either useful to a lot of people (Steam) or useful for testing Wine bugs (Firefox).

Aside from that, these app should be a single commit, and there are whitespace/indent issues.

If you really want to get these in, I'd like to see a convincing reason for each application (as a single commit). If they're super popular or used a lot by wine/linux users, that would be a good start.

Issue #559, opened 2015-10-18
vcrun2012/vcrun2013 need 64-bit support


austin987:
Mentioned here https://bugs.winehq.org/show_bug.cgi?id=39181#c3

vcrun2012 also lacks it.

austin987:
Fixed.

nyov:
Sadly the fix in #560 does not work as the 32bit installer fails and winetricks terminates before ever trying the 64bit download.

dufferzafar:
Yeah, I'm facing this issue as well.

PR #560, opened 2015-10-19
vcrun2012, vcrun2013: Add x64 support, closes issue #559

kakurasan:
Install also 64bit DLLs if wineprefix is 64bit.


austin987:
Thanks @kakurasan

Issue #561, opened 2015-10-19
Digital Editions new version?


Thomas-O:
Please can you add support for a more recent version of Adobe's Digital Editions?

I have happily used version 1.7.2 for many years but in the last few months I have noticed that it no longer works with an increasing number of e-book stores. A while back I read something about Adobe's implementation of a new DRM scheme to be included in versions 3.0+. This new scheme will demand that the end-user continually upgrades their copy of Digital Editions to the latest version. I don't know if the failures are as a result of stores upgrading their back-end software to match Adobe's newer versions. But it might be. And my fear is that without a working version of Digital Editions 4.0 or 4.5 then in a year or two from now it might again be impossible to download a DRM e-book to a Linux desktop.

Thanks for your time.

Thomas

austin987:
I'll take a look at some point, but would appreciate if someone else uploaded a patch for review.

austin987:
Fixed by https://github.com/Winetricks/winetricks/commit/b395dd86988303060dff518809ced08ffddb01aa

eli-schwartz:
???

I thought the reason this was stuck at version 1.7 is because the newer versions don't work!
Did you check it? Have you gotten it to work? If so, I'd love to know your secret...

Because sure enough, I tried installing adobe_diged into a fresh wineprefix, and Adobe Digital Editions instantly crashes.

ArchLinux latest
wine 1.7.54-1
winetricks-git 20150826.r27.gb395dd8-1


Please revert this. Better to have a version of ADE which doesn't support the new super-fancy DRM :wink: than a version of ADE which doesn't work at all.

eli-schwartz:
As for the DRM...

Adobe supports the new DRM system introduced in ADE 3.0 as well as the old system.
Bookstores have to specifically enable the new DRM. Many haven't, if only because there are no E-Ink ereaders out there that run a new enough version of RMSDK to support the new DRM, and people will start getting angry if they can't use their ereaders to read their books...

AFAIK the popular bookstores (at least in the US) still use the old system, so ADE 1.7 still works fine.
Kobo, Google Play, etc.
(B&N of course stopped allowing you to download their books with ADE at all.)


I was under the impression that it was a very rare bookstore which would enforce the new DRM.

austin987:
This launched for me without crashing. I didn't try loading any books.

What crash are you seeing / how to reproduce?

eli-schwartz:
Well, I take it all back.
[...]

PR #568, opened 2015-11-06
Offer both old and new versions of Adobe Digital Editions.

eli-schwartz:
The new version has features that might be undesirable, especially to anyone who absolutely must use an old E-Ink ereader. ;)
I chose to use adobe_diged4 for the new version, since the old version is known to be reliable, does not require winbind/samba which may or may not be installed, and also because I subjectively feel it should be the default. :wink:

Also provide a warning when installing ADE 4.5 if winbind is not detected, since the application fails to actually run without it.

These changes are in reference to comments made in #561

### DIFF
--- a/src/winetricks
+++ b/src/winetricks
@@ -9282,6 +9282,26 @@ load_abiword()
 #----------------------------------------------------------------
 
 w_metadata adobe_diged apps \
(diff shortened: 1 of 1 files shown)


austin987:
Looks great, thanks for the patches.

PR #571, opened 2015-11-15
Add "sound=pulse" verb

kakurasan:
Set sound driver to PulseAudio.

Note: PulseAudio support is available since Wine 1.7.55.


austin987:
Thanks, I meant to do this last week and forgot about it.

PR #573, opened 2015-11-17
Add hidewineexports setting

lucianposton:
Add winetricks settings to toggle hiding wine exports feature available in wine staging. See https://github.com/wine-compholio/wine-staging/blob/master/patches/ntdll-Hide_Wine_Exports/0001-ntdll-Add-support-for-hiding-wine-version-informatio.patch

### DIFF
--- a/src/winetricks
+++ b/src/winetricks
@@ -18735,6 +18735,42 @@ _EOF_
 
 #----------------------------------------------------------------
 
+w_metadata hidewineexports=enable settings \
+    title="Enable hiding wine exports from applications (wine-staging)"
+w_metadata hidewineexports=disable settings \
+    title="Disable hiding wine exports from applications (wine-staging)"
+
+load_hidewineexports()
+{
+    # Wine exports some functions allowing apps to query the wine version and
+    # information about the host environment. Using these functions, some apps
+    # will intentionally terminate if they can detect that they are running in
+    # a wine environment.
+    #
+    # Hiding these wine exports is only available in wine-staging.
+    # See https://bugs.winehq.org/show_bug.cgi?id=38656
+    case $arg in
+        enable)
+            local registry_value="\"Y\""
+            ;;
+        disable)
+            local registry_value="-"
+            ;;
+        *) w_die "Unexpected argument, $arg";;
+    esac
+
+    cat > "$W_TMP"/set-wineexports.reg <<_EOF_
+REGEDIT4
+
+[HKEY_CURRENT_USER\Software\Wine]
+"HideWineExports"=$registry_value
+
+_EOF_
+    w_try_regedit "$W_TMP"/set-wineexports.reg
+}
+
+#----------------------------------------------------------------
+
 w_metadata hosts settings \
     title_uk="Додати порожні файли у C:\windows\system32\drivers\etc\{hosts,services}" \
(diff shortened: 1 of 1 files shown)


austin987:
@slackner @DarkPlayer

DarkPlayer:
Might make sense, especially if you want to add a verb for http://warspear-online.com/en/home

slackner:
The patch looks good and works as expected. There is also a checkbox in winecfg though, so a winetricks recipe is not strictly required. @lucianposton: Is there any specific reason you need this recipe?

slackner:
>Might make sense, especially if you want to add a verb for http://warspear-online.com/en/home

Usually games are out of the scope of winetricks.

lucianposton:
> There is also a checkbox in winecfg though, so a winetricks recipe is not strictly required. @lucianposton: Is there any specific reason you need this recipe?

It's one of the winecfg settings I configure for each new prefix I create. Automating it through winetricks just makes that easier.

Besides the app DarkPlayer mentioned, I don't know of any others that specifically use the hidden procedures (except one that just queries and displays various system info). I do it as a precaution to hide wine/linux host info.

austin987:
It's okay in principle.

Is there a related wine bug? If so, w_workaround_bug() should be used (to warn/explain what his happening). If the bug is a WONTFIX, please add (wine-staging) to the verb description.

DarkPlayer:
The bug is https://bugs.winehq.org/show_bug.cgi?id=38656 which was closed as invalid since the application checks this on purpose. You can consider it as WONTFIX.

austin987:
Sure. I'd like to the verb labelled as wine-staging so if someone asks why it doesn't on plain wine I can point to that :).

Please include that bug in a brief comment

lucianposton:
I updated the branch with the two changes, added "wine-staging" to the verb title and added a comment referencing the wine bug.

src/winetricks
```diff
@@ -18735,6 +18735,42 @@ _EOF_
 
 #----------------------------------------------------------------
 
+w_metadata hidewineexports=enable settings \
+    title="Enable hiding wine exports from Windows (wine-staging)"
```
austin987: This should probably be something like 'Windows applications' or just 'applications'

src/winetricks
```diff
@@ -18735,6 +18735,42 @@ _EOF_
...
+w_metadata hidewineexports=disable settings \
+    title="Disable hiding wine exports from Windows (wine-staging)"
```
austin987: Also here.

src/winetricks
```diff
@@ -18735,6 +18735,42 @@ _EOF_
...
+
+load_hidewineexports()
+{
+    # Wine exports some procedures allowing apps to query the wine version and
```
austin987: 'some procedures' > 'some functions'

src/winetricks
```diff
@@ -18735,6 +18735,42 @@ _EOF_
...
+    # information about the host environment. Using these procedures, some
+    # apps will intentionally terminate if they can detect that they are run
```
austin987: 'run' > 'running'

austin987:
Looking better, thanks. I noticed a few other things while reviewing. Once those are fixed I think it should be good to merge.

lucianposton:
I pushed those requested changes.

austin987:
Thanks, merged.

PR #574, opened 2015-11-19
Correct carriage return conversion on w_ahk_do()

kmiksi:
I am not sure about the need of this addiction on ahk script, but at this time it seams to be unused.


austin987:
I assume Dan put it there for Windows support, but that's only speculation.

In any case, passes make check, and doesn't hurt, so why not?

PR #587, opened 2016-01-02
Update QQ,  Picasa and Chinese translation

hillwoodroc:


austin987:
51d3bbf looks fine, I added in line comments on the other commits

PR #591, opened 2016-01-09
Update QQ, Picasa and Chinese translation

hillwoodroc:
Instead of requests#587(https://github.com/Winetricks/winetricks/pull/587)

### DIFF
--- a/src/winetricks
+++ b/src/winetricks
@@ -16505,29 +16505,24 @@ load_popfs()
 #----------------------------------------------------------------
 
 w_metadata qq apps \
- title="QQ 7.1 (Chinese chat app)" \
+ title="QQ 8.0(Chinese chat app)" \
  publisher="Tencent" \
  year="2015" \
  media="download" \
- file1="QQ7.1.exe" \
+ file1="QQ8.0.exe" \
  file2="QQ.tar.gz"\
- installed_exe1="$W_PROGRAMS_X86_WIN/Tencent/QQ/Bin/QQ.exe" \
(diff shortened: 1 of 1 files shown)


austin987:
FYI, you can use git push --force to force update a commit that you've already pushed (not considered a good practice for published repos, but for pull requests it's preferable over closing/opening multiple requests).

src/winetricks
```diff
@@ -16546,12 +16543,23 @@ load_qq()
         w_try mv QQ/腾讯QQ.desktop ~/.local/share/applications/wine/Programs/腾讯软件/QQ
         w_try mv QQ/48x48/QQ.png ~/.local/share/icons/hicolor/48x48/apps
         w_try mv QQ/256x256/QQ.png ~/.local/share/icons/hicolor/256x256/apps
-        echo Exec=env WINEPREFIX=$WINEPREFIX $WINE C:\/Program\\ Files\/Tencent\/QQ\/Bin\/QQ.exe >> $HOME/.local/share/applications/wine/Programs/腾讯软件/QQ/腾讯QQ.desktop
+        echo Exec=env WINEPREFIX=$WINEPREFIX $WINE \"$W_PROGRAMS_X86_WIN\\Tencent\\QQ\\bin\\QQScLauncher.exe\">> $HOME/.local/share/applications/wine/Programs/腾讯软件/QQ/腾讯QQ.desktop
```
austin987: $WINEPREFIX / $WINE should be quoted (or alternatively, the entire line)

hillwoodroc:
In fact, showing these three commits was my intention. I think this way could clearly show every changes. Ok, if you think this was not a good action, I will fix later.

PR #593, opened 2016-01-11
Adapt to new Debian /usr/bin/wineserver[-development].

jre-wine:
Hi

Debian provides /usr/bin/wineserver since wine 1.8-2. So the existing workarounds for finding the wineserver are only needed for older Debian versions.
However for wine-development the wineserver will be /usr/bin/wineserver-development. I added that to the list and the binary detection code.

Thanks a lot
jre


austin987:
Thanks @jre-wine

PR #602, opened 2016-02-08
Self-update functionality for winetricks.

saulotoledo:
With this patch it is possible to use "winetricks --self-update|--selfupdate" to automatically update the script with the last online version. Also, it is possible to rollback an update by running "winetricks --update-rollback".

This patch is a proposal and can be improved. Let me know if it is useful and if you want some adjustments.

### DIFF
--- a/src/winetricks
+++ b/src/winetricks
@@ -614,6 +614,68 @@ w_dotnet_verify()
             dn_status=$?
 }
 
+# Checks if the user can run the self-update/rollback commands
+winetricks_check_update_availability()
+{
+    # Prevents the development file overwrite:
+    if test -d "../.git"
+    then
+        w_warn "You're running in a dev environment. Please make a copy of the file before running this command."
+        exit;
+    fi
+
+    # Checks read/write permissions on update directories
+    if ! (test -r $0 && test -w $0 && test -w ${0%/*} && test -x ${0%/*})
+    then
+        w_warn "You don't have the proper permissions to run this command. Try again with sudo or as root."
+        exit;
+    fi
+}
+
+winetricks_selfupdate()
+{
+    winetricks_check_update_availability
+
+    _W_filename="${0##*/}"
+    _W_rollback_file="${0}.bak"
+    _W_update_file="${0}.update"
+
+    _W_tmpdir=${TMPDIR:-/tmp}
+    _W_tmpdir="`mktemp -d "$_W_tmpdir/$_W_filename.XXXXXXXX"`"
+
+    w_download_to $_W_tmpdir https://raw.githubusercontent.com/Winetricks/winetricks/master/src/winetricks
+    w_try mv $_W_tmpdir/$_W_filename $_W_update_file.gz
+    w_try gunzip $_W_update_file.gz
+    w_try rmdir $_W_tmpdir
+
+    w_try cp $0 $_W_rollback_file
+    w_try chmod -x $_W_rollback_file
+
+    w_try mv $_W_update_file $0
+    w_try chmod +x $0
+
+    w_warn "Update finished! The current version is '`$0 -V`'. Use 'winetricks --update-rollback' to return to the previous version."
+
+    exit;
+}
+
+winetricks_selfupdate_rollback()
+{
+    winetricks_check_update_availability
+
+    _W_rollback_file="${0}.bak"
+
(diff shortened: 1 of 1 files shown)


qwertychouskie:
Has this been forgotten?  I would love a self-update command.

saulotoledo:
@qwertychouskie not really. I had a personal situation to handle, but I will submit an update to the patch  probably next weekend.

austin987:
On Mar 4, 2016 8:18 PM, "Saulo S. Toledo" <notifications@github.com> wrote:
>
> @qwertychouskie not really. I had a personal situation to handle, but I
will submit an update to the patch probably next weekend.

Hi Saulo,

A thought about keeping the old version around. As I said,  /tmp means it
will likely be lost after a reboot. Instead, why not rename to
winetricks.bak, in its current directory, along with removing executable
permission. That way it stays around, and can easily be restored, but
prevents polluting $PATH

saulotoledo:
Hello @austin987 and @qwertychouskie, sorry for the delay, here is the update.

@austin987, I've tried to follow your comments. A brief summary:
- The distro versioning problem: Instead of a self containing winetricks version, the distro's install files could contain a download/update script for winetricks. If (for example) /usr/bin/winetricks does not exists, download it; if it exists, run the self-update command. I know that this is a more complex problem, since that some distros freeze versions, but old versions of winetricks aren't a good idea anyway...

- The tmp dir handling: It was really complex. Now the code is simpler. I am still using the tmp dir to download winetricks since that w_download_to() trusts on $W_CACHE, that is initialized on winetricks_init(), but it doesn't make sense to run winetricks_init() before the update. But there is a major problem that we need to solve and I ask for your help:
    - When I run w_download_to(), _W_cache="$W_CACHE/$_W_packagename" (line 687) results on "(empty)//tmp/mydir". The download works just by a side effect, but this needs to be changed. I've used w_download_to() to avoid using wget.

- The gunzip at code: this is another side effect of using w_download_to();

- I've removed the --selfupdate alias for --self-update;

- The rollback file: Initially I wasn't worried about to store the rolback version after the machine reboot, but store the .bak file for some more time seems a better idea. I've done as you asked for;

- I've added a warning message for users who run unprivileged and do not have the proper permissions to update winetricks. I've also added a message to avoid to override the dev version.

austin987:
Please use w_warn() instead of special echo's.

winetrics_check_update_availability < missing a k in winetricks

Also, please squash to 1 commit.

saulotoledo:
Hello again @austin987 !
I've done as you asked for. Let me know if you want some more updates.

austin987:
I'll try to review this week, been busy.

src/winetricks
```diff
@@ -614,6 +614,68 @@ w_dotnet_verify()
...
+    _W_filename="${0##*/}"
+    _W_rollback_file="${0}.bak"
+    _W_update_file="${0}.update"
+
+    _W_tmpdir=${TMPDIR:-/tmp}
+    _W_tmpdir="`mktemp -d "$_W_tmpdir/$_W_filename.XXXXXXXX"`"
+
+    w_download_to $_W_tmpdir https://raw.githubusercontent.com/Winetricks/winetricks/master/src/winetricks
+    mv $_W_tmpdir/$_W_filename $_W_update_file.gz
+    gunzip $_W_update_file.gz
```
austin987: Please add w_try around shell commands (here and below).

Looks fine other than that.

saulotoledo:
Thank you for the review @austin987 !
I've added another commit with the changes. Could you confirm them? I will squash the commits again when you approve it.

austin987:
@saulotoledo Yes, looks good. Please squash.

saulotoledo:
@austin987 Done!

austin987:
Thanks

saulotoledo:
Thank you for the merge @austin987 !
I want to suggest one more thing to be checked next:

> When I run w_download_to(), _W_cache="$W_CACHE/$_W_packagename" (line 687) results on "(empty)//tmp/mydir". The download works just by a side effect, but this needs to be changed. I've used w_download_to() to avoid using wget.

Since this is a side effect, I think that it is something that need to be checked. I have avoided touching in this function because I believe it is something to someone with more experience with the project's code. My suggestion is to open a new issue for this.

austin987:
Please open a new issue.

PR #617, opened 2016-03-03
Fixing the wording

birdie-github:
"app" has a very special meaning nowadays and it has nothing to do with Wine

secondly, this option contains certain Windows libraries

### DIFF
--- a/src/winetricks
+++ b/src/winetricks
@@ -2475,7 +2475,7 @@ winetricks_prefixmenu()
          ;;
     *)   _W_msg_title="Winetricks - choose a wineprefix"
          _W_msg_body='What do you want to do?'
-         _W_msg_apps='Install an app'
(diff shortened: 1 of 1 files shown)


austin987:
libraries are under the dll category, not app

birdie-github:
Erm, then what is that?
libraries

austin987:
I assume you're referring to the checkboxes?

vc*express / vc*trial are Microsoft Visual Studio (well, express aka versions). They're the MS C++ compilers, not the redistributable libraries (those are vcrun20xx). Not an end-user library.

wme9 is windows media encoder, for encoding audio/video files. Also not an end-user library.

birdie-github:
Thanks for the explanation!

austin987:
No problem, thanks for the patches!

FYI, if you're curious about something, might be easier to email me or find me on IRC ;)

PR #620, opened 2016-03-04
Typos

birdie-github:
Still it's an "application" not an "app".

(Fixed the Russian translation accordingly - ostensibly there should be no libraries).


austin987:
Thanks for catching/fixing this.

PR #629, opened 2016-03-30
Warn Russian IPs before trying to download from web archive

birdie-github:
There are already several bug reports in regard to the inability to download files from web.archive.org. Let's warn people why downloads might fail.

### DIFF
--- a/src/winetricks
+++ b/src/winetricks
@@ -143,6 +143,8 @@ XDG_CACHE_HOME="${XDG_CACHE_HOME:-$HOME/.cache}"
 
 W_PREFIXES_ROOT="${WINE_PREFIXES:-$XDG_DATA_HOME/wineprefixes}"
 
+MYCOUNTRY=""
+
 # For temp files before $WINEPREFIX is available:
 if [ -x "`which mktemp 2>/dev/null`" ]
 then
@@ -768,6 +770,7 @@ w_download_to()
             w_die "Downloading $_W_url failed"
         fi
         # Download from the Wayback Machine on second try
+        winetricks_dl_warning
         _W_url="https://web.archive.org/web/$_W_url"
     done
 
@@ -2340,6 +2343,28 @@ _EOF_
 }
 
 #---- Private Functions ----
+
+winetricks_dl_warning() {
+    case $LANG in
+    ru*) _W_countrymsg="Скрипт определил, что ваш IP адрес принадлежит России. Если во время загрузки файлов вы увидите ошибки несоответствия сертификата, скачайте файлы вручную или используйте VPN."
+      ;;
+    *)  _W_countrymsg="Your IP address has been determined to belong to Russia. If you encounter a certificate error while downloading, please use VPN or download files manually."
+      ;;
+    esac
+
+    # Lookup own country via IP address only once (i.e. don't run this function for every download invocation)
+    if [ -z "$MYCOUNTRY" -a -x "`which geoiplookup 2>/dev/null`" ]
+    then
+        ip_address=`wget http://ipinfo.io/ip -q -O -`
+        export MYCOUNTRY=`geoiplookup "$ip_address" | awk '/Country Edition/{print $4}' | sed 's/,//'`
+        if [ "$MYCOUNTRY" = "RU" ]
+        then
+           w_warn "$_W_countrymsg"
+        fi
+    else
+        export MYCOUNTRY="Unknown"
+    fi
+}
 
 winetricks_get_sha1sum_prog() {
     # Mac folks tend to not have sha1sum, but we can make do with openssl


qwertychouskie:
Along with a warning, it might be nice to detect if `tor` is installed, and if so, prefix download commands with `torify`, as seen in this article: http://linuxaria.com/howto/how-to-anonymize-the-programs-from-your-terminal-with-torify

austin987:
An interesting idea. I think I'd prefer an opt-in option for now, as users in some countries may not want to inadvertently start tor traffic (e.g., they may be okay with downloading winetricks on a particular connection, but not using tor).

That should be a separate discussion/bug, however.

src/winetricks
```diff
@@ -143,6 +143,8 @@ XDG_CACHE_HOME="${XDG_CACHE_HOME:-$HOME/.cache}"
 
 W_PREFIXES_ROOT="${WINE_PREFIXES:-$XDG_DATA_HOME/wineprefixes}"
 
+MYCOUNTRY=""
```
austin987: Variables should be PREFIXED with W_

src/winetricks
```diff
@@ -2341,6 +2344,28 @@ _EOF_
...
+      ;;
+    *)  _W_countrymsg="Your IP address has been determined to belong to Russia. If you encounter a certificate error while downloading, please use VPN or download files manually."
+      ;;
+    esac
+
+    # Lookup own country via IP address only once (i.e. don't run this function for every download invocation)
+    if [ -z "$MYCOUNTRY" -a -x "`which geoiplookup 2>/dev/null`" ]
+    then
+        ip_address=`wget http://ipinfo.io/ip -q -O -`
+        export MYCOUNTRY=`geoiplookup "$ip_address" | awk '/Country Edition/{print $4}' | sed 's/,//'`
```
austin987: Rather than depending on geoiplookup, you could use http://ipinfo.io/..

ip="$(wget http://ipinfo.io/ip -q -O -)"
country="$(wget -q -O - http://ipinfo.io/${ip} | grep country | cut -d: -f2| tr -d "," | tr -d \" )"

src/winetricks
```diff
@@ -2341,6 +2344,28 @@ _EOF_
...
+        if [ "$MYCOUNTRY" = "RU" ]
+        then
+           w_warn "$_W_countrymsg"
+        fi
+    else
+        export MYCOUNTRY="Unknown"
```
austin987: If you're going to name this variable country, it should support all countries, not just Russia. Using the ipinfo solution above also solves this..

That could then be used for anything else country specific that may be needed in the future.

src/winetricks
qwertychouskie: Hard-coding a Wget dependency is probably not a good idea, winetricks is designed to be able to use `aria2c`, `wget`, or `curl`.

birdie-github:
It was a general idea, Austin ;-)

Actually your solution with ipinfo.info is better but then we'll depend on this particular website, while `geoiplookup` is more or less independent and futureproof. But then even my hack relies on some website to establish user's IP address. Unfortunately, to my knowledge, there's no general solution for discovering one's IP address without using 3d parties.

What I didn't understand is "it should support all countries, not just Russia" - right now my hack doesn't discriminate any other countries - it just checks against Russia as only this country is known to block wayback archive.

austin987:
@birdie-github 
my concern with geoiplookup is that it's an extra dependency, which many people won't have installed. Winetricks relies on plenty of other third parties by design, hence the frequent updates ;)

My comment about not just Russia is that you should set the variable ($MYCOUNTRY) for any known country. Currently, it sets Russia or Unknown.

src/winetricks
```diff
@@ -2341,6 +2344,28 @@ _EOF_
 
 #---- Private Functions ----
 
+winetricks_dl_warning() {
+    case $LANG in
+    ru*) _W_countrymsg="Скрипт определил, что ваш IP адрес принадлежит России. Если во время загрузки файлов вы увидите ошибки несоответствия сертификата, скачайте файлы вручную или используйте VPN."
+      ;;
+    *)  _W_countrymsg="Your IP address has been determined to belong to Russia. If you encounter a certificate error while downloading, please use VPN or download files manually."
```
qwertychouskie: The message should be changed from:
"Your IP address has been determined to belong to Russia. If you encounter a certificate error while downloading, please use VPN or download files manually."
to:
"Your IP address has been determined to belong to Russia. If you encounter a certificate error while downloading, please make sure tor is installed and relaunch with the --torify option, or use a VPN to download files manually."
becaue of https://github.com/Winetricks/winetricks/issues/630.  (Also, the Russian translation should be updated accordingly.)

qwertychouskie:
Any updates on this PR?

birdie-github:
I cannot figure out how to push a new patch.

Here's the relevant commit: https://github.com/birdie-github/winetricks/commit/bae38d6204a7a4f0d3c4d26be8ec925d5268865f

Proper full name country resolution is not there because it's quite complicated.

birdie-github:
@austin987 

Ping.

austin987:
@birdie-github I'm going to refactor it a bit. If you don't see it committed within a week, please ping me again.

austin987:
@birdie-github, this should be fixed now, the important commit is https://github.com/Winetricks/winetricks/pull/629#issuecomment-266390835

Please reopen/file an issue if it's not working for you.

PR #639, opened 2016-05-06
src/release.sh: create local signed release tarball

jre-wine:
From the discussion at
http://lists.alioth.debian.org/pipermail/pkg-wine-party/2016-May/005549.html

See https://wiki.debian.org/Creating%20signed%20GitHub%20releases

Currently the  tarball is created bit-by-bit identical.
There might be issues if github recreates the tarballs later
and changes their gzip or git archive implementations.

Unfortunately github currently doesn't provide remote shell
access, so the signature needs to be uploaded in their
webinterface.

### DIFF
--- a/src/release.sh
+++ b/src/release.sh
@@ -39,4 +39,13 @@ git tag -s -m "winetricks-${version}" ${version}
 git push
 git push --tags
 
+# create local tarball, identical to github's generated one
+git archive --prefix="winetricks-${version}/" -o "../${version}.tar.gz" "${version}"
+
+# create a detached signature of the tarball
+gpg --armor --detach-sign "../${version}.tar.gz"
+
+echo "Go to https://github.com/winetricks/winetricks/releases/edit/${version}"
+echo "and upload the signature ../${version}.tar.gz.asc."
+
 exit 0


austin987:
Dear lord that's a terrible process

jre-wine:
It could be worse, but I'm not happy with it either. I may close it, and we wait forever^Wuntil github offers shell access. Up to you, if you are willing to take this manual step hurdle.

jre-wine:
To avoid any doubt: you don't have to follow the instructions from the debian wiki link. Just follow the three step instructions given by release.sh.

austin987:
I'm not happy about it, but I can. Do you want this done for the last release as well?

I probably won't merge/do it until next week though, fyi

jre-wine:
Thanks a lot! I'm indeed not happy with the manual steps either, it's a burden for you and prone to error/be forgotten. Still it's more secure than only the signed tags (which are already a great step forward). And as we saw on another hosting site that for some time bundled adware in opensource software, it definitely is a good idea to sign locally.

I hope (TM) the Debian release will be out soon, so no need to sign the current release from my side.

josephbisch:
There is an API that GitHub has that supports [uploading release assets](https://developer.github.com/v3/repos/releases/#upload-a-release-asset), so you could potentially avoid needing to use the web interface if you have a script that uploads the signature. I'll take a look at writing such a script if I get a chance.

You could also delete the existing archives that GitHub generates and replace them with the archives generated by the release.sh script to avoid any potential issues if GitHub recreates their tarball in such a way that breaks the signature.

josephbisch:
I put together a Python 3 script that can be called in the following way to upload an asset to a GitHub release without needing to use the web interface.

```
python3 github-api-releases.py  ../{version}.tar.gz.asc Winetricks winetricks ${version}
```

For now, I just stuck the script in a temporary repo that I used to test the script:
https://github.com/josephbisch/test-releases-api/blob/master/github-api-releases.py

Make sure that you set the `GITHUB_TOKEN` environment variable with a personal access token that you can generate through the settings on the GitHub web interface. I gave mine the repo scope, which worked, but I'm not sure if there is a more specific scope that can be given that would still work in this case.

You can also get a token using curl:
```
curl -u username --data '{"scopes":["repo"], "note": "test"}' https://api.github.com/authorizations
```
where username is your GitHub username. Curl will prompt you for your GitHub password. The value of the note should be replaced with something descriptive, because it will show up in your settings on GitHub when you look at your tokens. You then take the token from the response and use that as the `GITHUB_TOKEN`.

austin987:
On May 7, 2016 11:29 PM, "Joseph Bisch" <notifications@github.com> wrote:
>
> I put together a Python 3 script that can be called in the following way
to upload an asset to a GitHub release without needing to use the web
interface.
>
> python3 github-api-releases.py  ../{version}.tar.gz.asc Winetricks
winetricks ${version}
>
> For now, I just stuck the script in a temporary repo that I used to test
the script:
>
https://github.com/josephbisch/test-releases-api/blob/master/github-api-releases.py
>
> Make sure that you set the GITHUB_TOKEN environment variable with a
personal access token that you can generate through the settings on the
GitHub web interface. I gave mine the repo scope, which worked, but I'm not
sure if there is a more specific scope that can be given that would still
work in this case.
>
> You can also get a token using curl:
>
> curl -u username --data '{"scopes":["repo"], "note": "test"}'
https://api.github.com/authorizations
>
> where username is your GitHub username. Curl will prompt you for your
GitHub password. The value of the note should be replaced with something
descriptive, because it will show up in your settings on GitHub when you
look at your tokens. You then take the token from the response and use that
as the GITHUB_TOKEN.

Thanks a lot, Joseph! I'll review tomorrow or Monday.

Do you intend to keep the repo this script is on active? I will keep a copy
in winetricks regardless, but want to know if I should link to your repo or
not.

Thanks,
-Austin

austin987:
On May 8, 2016 1:46 AM, "Austin English" <austinenglish@gmail.com> wrote:
>
> On May 7, 2016 11:29 PM, "Joseph Bisch" <notifications@github.com> wrote:
> >
> > I put together a Python 3 script that can be called in the following
way to upload an asset to a GitHub release without needing to use the web
interface.
> >
> > python3 github-api-releases.py  ../{version}.tar.gz.asc Winetricks
winetricks ${version}
> >
> > For now, I just stuck the script in a temporary repo that I used to
test the script:
> >
https://github.com/josephbisch/test-releases-api/blob/master/github-api-releases.py
> >
> > Make sure that you set the GITHUB_TOKEN environment variable with a
personal access token that you can generate through the settings on the
GitHub web interface. I gave mine the repo scope, which worked, but I'm not
sure if there is a more specific scope that can be given that would still
work in this case.
> >
> > You can also get a token using curl:
> >
> > curl -u username --data '{"scopes":["repo"], "note": "test"}'
https://api.github.com/authorizations
> >
> > where username is your GitHub username. Curl will prompt you for your
GitHub password. The value of the note should be replaced with something
descriptive, because it will show up in your settings on GitHub when you
look at your tokens. You then take the token from the response and use that
as the GITHUB_TOKEN.
>
> Thanks a lot, Joseph! I'll review tomorrow or Monday.
>
> Do you intend to keep the repo this script is on active? I will keep a
copy in winetricks regardless, but want to know if I should link to your
repo or not.
>
> Thanks,
> -Austin

Actually,  could you put a license on that script, along with your
copyright? Preferably LGPL to match wine/winetricks, but it's your decision.

josephbisch:
I'll keep the repo active. I added the copyright and license info for LGPL. I also added a description to the README, which describes how to use the script like I did in the comment here.

Also, I realized that currently winetricks isn't making "real" GitHub releases. When you push a tag, it automatically shows up on the releases tab of the web ui, but you can't upload assets via the API. I explain the situation in detail in [this commit](https://github.com/josephbisch/test-releases-api/commit/71bfd23a18cb6db52d65e5bfd7672d5381ec1248), but as the commit shows, I was able to just add a function to create a new release based on the existing tag.

austin987:
Finally looked at this. Script doesn't work for me, I sent a PR: https://github.com/josephbisch/test-releases-api/pull/1

After that, it fails on:
+ python3 src/github-api-releases.py ../20160627.tar.gz.asc Winetricks winetricks 20160627
Traceback (most recent call last):
  File "src/github-api-releases.py", line 96, in <module>
    ret = upload_asset(path, owner, repo, tag)
  File "src/github-api-releases.py", line 51, in upload_asset
    token = os.environ['GITHUB_TOKEN']
  File "/usr/lib/python3.5/os.py", line 725, in __getitem__
    raise KeyError(key) from None
KeyError: 'GITHUB_TOKEN'

austin@debian-laptop:~/src/winetricks$ echo $GITHUB_TOKEN
.......................censored..............

austin987:
Anyway, because I forgot to comment out git push, this is sorta committed :), but doesn't work yet (https://github.com/Winetricks/winetricks/pull/639#issuecomment-228941334)

josephbisch:
Thanks for the PR. I looked at how it fails after the PR issue is resolved. Could it be possible that you didn't export $GITHUB_TOKEN from the terminal before running release.sh? If so, I should improve the script to display a message asking the script's user if GITHUB_TOKEN has been exported instead of assuming the env variable is always there.

austin987:
That's why I did:
austin@debian-laptop:~/src/winetricks$ echo $GITHUB_TOKEN
.......................censored..............

to show that the variable is set (it's a real token, not ...censored..., of course).

josephbisch:
I just never saw if you ran `export GITHUB_TOKEN`, so I wasn't sure if you did or not. It wasn't clear to me if you ran `echo $GITHUB_TOKEN` in the same terminal as the script that calls github-api-releases.py, which, if you never exported GITHUB_TOKEN, would be the obvious explanation for why the echo worked, but the Python script isn't able to find the env variable.

I guess it could be possible there is a bug with Python and os.environ. Otherwise, I'm not sure what the issue may be. I don't think it can be an issue with the GitHub upload script, because either os.environ finds the environment variable or not. I don't think I messed that up, because there wasn't anything for me to mess up with respect to that part.

austin987:
Thanks @josephbisch. Not sure what I did wrong, I had it in my ~/.bashrc, then ran . ~/.bashrc before testing.

In any case, I'm getting further now. Current issue is:
```
python3 src/github-api-releases.py ../20160709.tar.gz.asc Winetricks winetricks 20160709
Traceback (most recent call last):
  File "src/github-api-releases.py", line 96, in <module>
    ret = upload_asset(path, owner, repo, tag)
  File "src/github-api-releases.py", line 73, in upload_asset
    content_type = magic.from_file(path)
AttributeError: module 'magic' has no attribute 'from_file'
```

That was fixed by:
```
diff --git a/src/github-api-releases.py b/src/github-api-releases.py
index 2ac195f..05e15dc 100644
--- a/src/github-api-releases.py
+++ b/src/github-api-releases.py
@@ -70,7 +70,7 @@ def upload_asset(path, owner, repo, tag):
     with open(path) as f:
         contents = f.read()
 
-    content_type = magic.from_file(path)
+    content_type = magic.detect_from_filename(path)
 
     headers = {'Content-Type': content_type, 'Authorization': token}
     params = {'name': fname}
```

which then fails with:
```
python3 src/github-api-releases.py ../20160709.tar.gz.asc Winetricks winetricks 20160709
Traceback (most recent call last):
  File "src/github-api-releases.py", line 96, in <module>
    ret = upload_asset(path, owner, repo, tag)
  File "src/github-api-releases.py", line 79, in upload_asset
    headers=headers, params=params)
  File "/usr/lib/python3/dist-packages/requests/api.py", line 111, in post
    return request('post', url, data=data, json=json, **kwargs)
  File "/usr/lib/python3/dist-packages/requests/api.py", line 57, in request
    return session.request(method=method, url=url, **kwargs)
  File "/usr/lib/python3/dist-packages/requests/sessions.py", line 475, in request
    resp = self.send(prep, **send_kwargs)
  File "/usr/lib/python3/dist-packages/requests/sessions.py", line 585, in send
    r = adapter.send(request, **kwargs)
  File "/usr/lib/python3/dist-packages/requests/adapters.py", line 403, in send
    timeout=timeout
  File "/usr/lib/python3/dist-packages/urllib3/connectionpool.py", line 578, in urlopen
    chunked=chunked)
  File "/usr/lib/python3/dist-packages/urllib3/connectionpool.py", line 362, in _make_request
    conn.request(method, url, **httplib_request_kw)
  File "/usr/lib/python3.5/http/client.py", line 1106, in request
    self._send_request(method, url, body, headers)
  File "/usr/lib/python3.5/http/client.py", line 1146, in _send_request
    self.putheader(hdr, value)
  File "/usr/lib/python3.5/http/client.py", line 1082, in putheader
    if _is_illegal_header_value(values[i]):
TypeError: expected string or bytes-like object
```

didn't debug further after that

josephbisch:
Sorry for the delay.

It looks like you are now using a different Python library that is also imported as magic.

There is what I'm using (https://github.com/ahupp/python-magic) and what you appear to be using (https://github.com/file/file/tree/master/python). I didn't realize there were multiple libraries that are imported as magic. I should have linked to the particular one I used.

The detect_from_filename function from the latter library doesn't return a string or bytes-like object:

```
def detect_from_filename(filename):
'''Detect mime type, encoding and file type from a filename
Returns a `FileMagic` namedtuple.
'''
```

You should be able to turn what detect_from_filename returns into a string, but I would just try the other library (python-magic).

austin987:
Hi @josephbisch

Sorry for my delay.

The problem is that the library your using does not appear to be packaged in Debian. I'd like to avoid installing something from pip/etc.

josephbisch:
Okay, but does it work now the way you have it in the winetricks repo with the try except to allow either library to be used? It looks like it should be working now with the python-magic in Debian, but I haven't tried it.

Of course, since the only file being uploaded in this case is the signature, we don't really need to bother with the python-magic stuff, since we can hardcode that particular mimetype.

austin987:
@josephbisch sorry for late follow up (again..). It is working for me now in Debian:
+ python3 src/github-api-releases.py ../../20161228.tar.gz.asc Winetricks winetricks 20161228
ERROR: Not Found
SUCCESS: 20161228.tar.gz.asc uploaded
+ exit 0

though there's that extraneous error, but that's a red herring.

Give https://github.com/Winetricks/winetricks/releases/download/20161228/20161228.tar.gz.asc a try :)

For reference, the working versions are:
austin@debian-laptop:~/src/winetricks$ dpkg -l | grep py.*magic
ii  python-magic                           1:5.29-2                              all          Recognize the type of data in a file using "magic" numbers - Python bindings
ii  python-pythonmagick                    0.9.14-3                              amd64        Object-oriented Python interface to ImageMagick
ii  python3-magic                          1:5.29-2                              all          Recognize the type of data in a file using "magic" numbers - Python 3 bindings
ii  python3-pgmagick                       0.6.2-2                               amd64        Yet Another Python wrapper for GraphicsMagick

PR #643, opened 2016-05-22
Re-add Songbird (2.2.0-2453) using FileHippo

einstein95:
All the official sites are dead and redirect to unrelated services, so this is basically for anyone that wants it.

### DIFF
--- a/src/winetricks
+++ b/src/winetricks
@@ -10979,6 +10979,27 @@ load_sketchup()
 
 #----------------------------------------------------------------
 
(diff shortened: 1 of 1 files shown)


austin987:
Is a dead Songbird really that useful? There's a free derivative that's still active-ish:
https://getnightingale.com/

einstein95:
Not really. Thanks for linking Nightingale, never heard of that before. I'll close this PR as it's useless.

PR #657, opened 2016-06-22
Add code editor hints

vanillajonathan:
To indent 4 spaces.

### DIFF
--- a/src/winetricks
+++ b/src/winetricks
@@ -1,4 +1,5 @@
 #!/bin/sh
+# vim: tabstop=8 expandtab shiftwidth=4 softtabstop=4
 
 # Name of this version of winetricks (YYYYMMDD)
 # (This doesn't change often, use the sha1sum of the file when reporting problems)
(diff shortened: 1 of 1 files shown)


src/winetricks
```diff
@@ -1,4 +1,5 @@
 #!/bin/sh
+# vim: tabstop=8 expandtab shiftwidth=4 softtabstop=4
```
austin987: Move to the end of the file please

vanillajonathan: I believe that the most common practice is to have it at the top of the file.

austin987: I've never seen one like that, in my workplace or open source projects.

Chiitoo: I'm not too familiar with 'vim', but from a quick look-see, I find more suggestions to having it in the end of the file instead of the beginning, although over at 'vim.wikia.com' [here] (http://vim.wikia.com/wiki/Modeline_magic#Examples), they say it can be up or down.

Aside from the few findings, I don't think I've ever seen it myself before, anywhere, so I can't say which really is the more common practice.  ^^

PR #638, opened 2016-05-05
Update winetricks

vanillajonathan:

PR #658, opened 2016-06-22
Update winetricks

vanillajonathan:
Proper capitalization of acronyms (such as GUI, URL, DLL and SP).
End sentences with a period.
Fix names of software titles where incorrect.
Capitalized software titles and brand names (such as Windows and Steam).
Adjust comments to not exceed 80 characters.

### DIFF
--- a/src/winetricks
+++ b/src/winetricks
@@ -1311,7 +1311,7 @@ fi
 # game_id
 #     Used for main installer name and download URL.
 # game_title
-#     Used for AutoHotKey and installation path in bat script.
+#     Used for AutoHotkey and installation path in bat script.
 # other_files
 #     Extra installer files, in one string, space-separated.
 # reader_control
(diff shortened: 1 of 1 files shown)


Chiitoo:
This is a bit late comment (would have been better for 'Update winetricks #638'), but shouldn't 'AutoHotKey' have been 'AutoHotkey'?  ^^

By the by, I for one would appreciate at least a bit more descriptive titles (“Update winetricks” doesn't tell much).

Thanks!

qwertychouskie:
"the POSIX stanard"?

vanillajonathan:
@Chiitoo, you're right it should be AutoHotkey, not AutoHotKey.

@qwertychouskie, thanks for spotting that regression where a typo was introduced.

austin987:
As others have said, the changelog entry needs to be updated. You really need to use a real git client for changes like this (and again, squash them).

Aside from that, some of these changes were rejected before (e.g., Xvid), and others are unneeded. Primarily, your use of periods is completely inconsistent (e.g., in the first few lines, you add a period to the first, but not the next 3 lines). Not every comment needs a period to end it, especially bullet points

Some others are undesirable, e.g., sha1sum is a known technical term, there's no need to obfuscate it for a more formal sentence that is a code comment (and that no longer matches the function name).

vanillajonathan:
Well you could just merge commit 367c34cab948da7a02eb6421909760160d98739f then.

austin987:
The Github gui merge doesn't allow that, as they're intermingled in one PR.

PR #667, opened 2016-09-06
winetricks: Fix a vcrun2015 override

### LINKED ISSUE #597: [Solved] Call from to unimplemented function api-ms-win-crt-math-l1-1-0.dll._except1, aborting
The game "Grim Dawn" on Steam requires vcrun2105.
Unfortunately there is a missing DLL. We got this error

Call to unimplemented function api-ms-win-crt-math-l1-1-0.dll._except1, aborting

This DLL has to be copied and set
[winetricks_20105.txt](https://github.com/Winetricks/winetricks/files/99729/winetricks_20105.txt)

 as native.
Can you apply this  patch so it can be fixed?

After that the game works properly

boltronics:
Fixes #597 which was erroneously closed.

### DIFF
--- a/src/winetricks
+++ b/src/winetricks
@@ -8456,7 +8456,7 @@ load_vcrun2015()
         w_warn "This may fail in non-XP mode, see https://bugs.winehq.org/show_bug.cgi?id=37781"
     fi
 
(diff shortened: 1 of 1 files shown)


austin987:
Nice catch, thanks.

PR #712, opened 2017-01-03
Add desktop and icon files

### LINKED ISSUE #710: Desktop and icon files missing
Winetricks is missing a desktop launcher and icon files. Because of that, Winetricks cannot be launched from the desktop menu and cannot be installed from GNOME Software (or other package manager GUI using the AppStream specification). Please, provide a valid desktop and icon files for Winetricks.

AsciiWolf:
This pull request adds the previously missing desktop and icon files.
Fixes #710.


AsciiWolf:
@austin987 Please let me know if there are any problems with this pull request.

jre-wine:
In Debian we have a nearly identical .desktop file, however it has
`Categories=Wine;`
while this PR has:
`Categories=Utility;Emulator; `
The latter are official ones, see https://specifications.freedesktop.org/menu-spec/latest/apa.html and https://specifications.freedesktop.org/menu-spec/latest/apas02.html. I'm not sure about this, but do you think "Wine" should be added to them? btw, the Wine wine.desktop specifies no Categories.



For the copyright:
I have verified that the suggested winetricks.svg is identical to the one in Debian:
$ md5sum debian/winetricks.svg winetricks.svg 
308b1036aadd9abde3f3925006027ad1  debian/winetricks.svg

Then we have in debian/copyright
```
Files: debian/winetricks.svg
Copyright:
 2010-2011 Joel Holdsworth <joel@airwebreathe.org.uk>
Comment:
 See http://bugs.debian.org/708198 for confirmation of copyright.
 .
 Date: Wed, 20 Apr 2011 18:10:48 -0700
 From: Scott Ritchie (winetricks Ubuntu maintainer) <scott@open-vote.org>
 To: Jari Aalto <jari.aalto@cante.net>
 Subject: Re: winetricks: *.svg file copyright in Ubuntu package
 .
 On 04/20/2011 12:01 AM, Jari Aalto wrote:
 .
 Oh hey!
 .
 > The Ubuntu package ships *.svg icon which would also be good for Debian.
 > Could your point em to the person who holds the copyright. Do you know
 > under which licence it is released?
 >
 .
 The icon is the same copyright as the icons in the wine/winetricks
 package itself -- Copyright Joel Holdsworth <joel@airwebreathe.org.uk>
 and licensed the same way Wine is (LGPL 2.1+)
 .
 I'll update the copyright file next upload.
 .
 Thanks,
 Scott Ritchie
License: LGPL-2.1+

```

AsciiWolf:
"Wine" is a non-standard category. I don't think it should be added.

Wine.desktop doesn't have any categories because it's not a menu launcher file.

jre-wine:
You're right, after running `desktop-file-validate winetricks.desktop` I'd suggest (if at all):
`Categories=Utility;Emulator;X-Wine;`

jre-wine:
Ok, after looking into this again I now really propose a change:
`Categories=Emulator;System;`

This way I see it in Gnome in "System Tools", instead of "Accessories". I think it matches much better in that category: "Accessories" is for things like Calculator, that you use for doing other things. "System Tools" is for changing parts of your System, which is what winetricks does.
"Emulator" seems to have no effect here, but "desktop-file-validate" suggested to use it together with System.
Both "Wine" and "X-Wine" seem to have no effect here, although I have a "Wine" category, created by installed Windows applications.

austin987:
Where does utilities end up? I think winetricks is more of a utility than a system tool. It should run as an unprivileged user, and is used within user writeable locations, not /var or the like. 

I'm not strongly opinioned, as long as there's consensus on what is chosen. I'm not going to be using the GUI to run it ;)

jre-wine:
Not a strong opinion here either. I just want to get this implemented and be done with it.

Here in Gnome if I change winetricks.desktop to ...

- "Categories:Utility" --> Menu "Accessories", not in "System Tools" or "Utilities".
- "Categories:System" --> Menu "System Tools", not in "Accessories" or "Utilities".

Other entries in menu...
- "Accessories": Bluetooth Transfer, Calculator, Desktop Search, Help, Maps, Notes, Remote Desktop Viewer, Gnome Tweak Tool. VirtualBox, Weather, several editors and pdf viewers
- "System Tools": Gnome Settings, CD/DVD Creator, Color Picker, GSmartControl, Keyboard Layout, Network Tools, Software, System Log, System Monitor, Terminal
- "Utilities": Archive Manager, Characters, Disks, Disk Usage Analyzer, Font Viewer, Help, Image Viewer, Logs, Passwords and Keys, Gnome Tweak Tool, pdf Viewer

austin987:
What is archive manager set to that gets it in utilities?

jre-wine:
/usr/share/applications/org.gnome.FileRoller.desktop:
`Categories=GTK;GNOME;Utility;Archiving;Compression;X-GNOME-Utilities;`

So here in Gnome category "Utility" is for menu "Accessories", and X-GNOME-Utilities for menu "Utilities". I'm against using desktop specific settings here (I assume we all agree on that).
My reasoning is I see "Accessories" as everyday helpers (weather, calculator, ...), while "System Tools" seems to be more for "change a setting and be done with it". I see winetricks in the latter (but no hard opinion on this). Obviously there's no clear line between those, Gnome's third category adds even more confusion.

AsciiWolf:
In my opinion, the "Utility" category is a better choice (I like it more and it's also [used](https://github.com/PlayOnLinux/POL-POM-4/blob/master/etc/PlayOnLinux.desktop) by PlayOnLinux), but as you noted, according to the [Desktop Menu Specification](https://standards.freedesktop.org/menu-spec/latest/apas02.html), the correct main category for the "Emulator" subcategory is "System", not "Utility", so the "System" category will probably be better after all.

austin987:
IMO wine is the emulator, and winetricks is a utility for it, not an
emulator itself.

jre-wine:
By now, I get why you "like" Accessories more than System Tools. And I think the actual feeling of someone who uses this is important here. Personally I only had the formal view on this. Further I agree that winetricks and playonlinux should be in the same category. So yes, maybe leave it in this menu.

I didn't see any effect of emulator anyway and also agree with Austin here. So maybe just drop "Emulator" and go with `Categories=Utility;` only.
desktop-file-validate is also happy with this.

AsciiWolf:
@jre-wine Done.

austin987:
@AsciiWolf, how would you feel about updating the comment to something like:
Work around problems in and install applications under Wine

In my experience, most users want to install native DLLs, not install games or applications.


I want to make some TravisCI changes to test make install before committing this, fyi, but should be Tuesday or Wednesday.

AsciiWolf:
@austin987 Done.

jre-wine:
@AsciiWolf Great! We'll probably go ahead with this version in Debian.

PR #718, opened 2017-01-18
Updated irfanview installer

bgstack15:
First pull request, so let me know how to do it better.
Updated the irfanview installer to use the latest irfanview version, 4.44. Updated the ahk script for its unattended install as well.

### DIFF
--- a/src/winetricks
+++ b/src/winetricks
@@ -10450,15 +10450,15 @@ load_iceweasel()
 w_metadata irfanview apps \
     title="Irfanview" \
     publisher="Irfan Skiljan" \
-    year="2014" \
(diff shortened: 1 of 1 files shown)


austin987:
@bgstack15 while it's a good idea to keep commits as small as possible, here you want one commit. Both changes are related and needed for winetricks -q irfranview to work after updating, so there's no need to separate those fixes. OTOH, adding new quiet support would be better done as its own commit.

I've pushed a squashed commit, https://github.com/Winetricks/winetricks/commit/defc0cb083a612f3081048a3840de0c634f6f501

Thanks

PR #726, opened 2017-01-29
Run Winetricks in a GUI mode by default

AsciiWolf:


jre-wine:
This is not needed for vanilla winetricks (this here), because it already defaults to use the GUI if no option is given.
Contrary in Debian and newer Ubuntu that default is changed (no option only gives a "help" message; to start the GUI "--gui" is required), and indeed we use the proposed patch in the desktop file.

So I don't understand the motivation to propose this here (maybe I missed something).

However I'd like to discuss this topic anyway: Which behavior is wanted if no option is given: start GUI or give help message?
Are there any strong opinions for this? Personally I'd just like Debian and upstream to do the same.
[Sorry if this is too close to an issue hijack, but I think that's still relevant here, and maybe even is the motivation for this PR.]

AsciiWolf:
Ah, thank you for the clarification. I think the current upstream behavior is proper.

austin987:
@jre-wine @AsciiWolf, while I think upstream winetrick's behavior is fine as it is, I don't have a problem with this patch. If someone is running from a .desktop file, it's obvious that they want the GUI and not the CLI option.

it is more explicit, doesn't hurt to change it upstream, and reduces the downstream diff. I think those are all great reasons to do it upstream.

AsciiWolf:
@austin987 I am getting an error message when launching upstream version of winetricks with the `--gui` parameter. I think it will be better to revert this commit.

winetricksgui_error

austin987:
@AsciiWolf that's a separate issue. I have a patch in progress for it. Feel free to file a bug so I don't forget :)

jre-wine:
@austin987 just for the record: I plan to completely drop that patch in Debian (will propose it soon, then wait for feedback and then do it once Debian is unfrozen).

austin987:
@jre-wine, cool. My goal was to reduce your diff going forward. If you've got more to upstream, please send PRs.

jre-wine:
@austin987, will definitely do so, I hate unnecessary diffs. And I really appreciate that you applied this patch to reduce the diff!

austin987:
@AsciiWolf, that problem should be fixed by https://github.com/Winetricks/winetricks/commit/d6cf3bd72d3bb6518763deb3b581a1e62e4a363f

AsciiWolf:
@austin987 Thanks! But since the Debian patch is planned to be dropped, I still think it would be better to revert this pull request as it is not needed anymore. :-)

austin987:
@AsciiWolf, I don't understand your reasoning. If we're running from a .desktop file, the user definitely wants a GUI. I find it preferable to be explicit if we can, and here it is very easy.

What's the harm from it?

AsciiWolf:
@austin987 Well, you're right, there is no harm from it.

PR #729, opened 2017-02-05
Clarify license

jre-wine:
This PR clarifies and sometimes changes the license of some files. So **we need consent from some authors before merging this**.

Some notes, but see also the specific commit messages.

1. LGPL 2 vs. 2.1
see commit message for src/winetricks

2. Custom license notices without "or later"
Several files used a custom license notice which lacks the "or later". **Allowing "later" requires the consent of all authors. That's Austin, Dan Kegel and Joseph Bisch.** I suggest to use the SPDX short license identifier here if the usual license notifier is not wanted. Although these short identifiers are not what GNU suggests in the original license, I think they are a fully valid replacement. I've seen some other projects, but by far not the majority of them, use them.
Alternatively I'd also be happy to provide a PR which always uses the official license notifier. I strongly suggest not to use custom notifiers. As seen here this can go wrong easily.

3. Missing copyright/license header
For src/uniquefiles.sh I added this based on the git history.
For winetricks-test I just referenced the copyright list in src/winetricks.

Austin, are you generally ok with this? If yes, do you prefer the SPDX short identifier, or the traditional identifier? Once I've the ACK from you I'd ask Joseph and Dan.


austin987:
@jre-wine I'm generally okay with it. However src/winetricks has had many more authors than just myself/Dan/Joseph.

While the text is already or later, I'm not sure it's kosher to change it without consent from all authors (may be a question for debian-legal@)

jre-wine:
@austin987 Yes, I also wondered about that. I'll ask at debian-legal, with my reasoning that this is a typo (see commit message), not a change of license.

austin987:
@jre-wine, great. Yeah, it appears to be a typo, I agree.

Please CC me on that email

austin987:
@dankegel @josephbisch

austin987:
Though most of these commits can be merged (aside from src/winetricks) if
Dan and Joseph are okay with it (for the abbreviation, the formatting is
fine as is IMO). Thanks

josephbisch:
ACK

jre-wine:
I asked at [debian-legal](https://lists.debian.org/debian-legal/2017/02/msg00005.html).

@austin987 "the formatting is fine as is IMO"  - I'm not sure what you mean with that. Should I avoid changing of the formatting (e.g. by prepending "Copyright:" and "License:") where possible? I'm fine with either way, the current layout is just a result of my workflow (copy&pasting a stub and see what changes).
Or was it a general ACK? I've got at least one minor fix not yet (force) pushed, I'll do that in a batch once I've got sufficient answers.

Thanks @josephbisch

austin987:
@jre-wine Consistency is nice, but I also wasn't bothered by the current state. I'm fine with either way.

jre-wine:
I force-updated my license branch (the old one is now called license.v1.SPDX). I dropped the SPDX things again, and instead just added the "or later" to the short custom notice.

@dankegel We need your agreement to change the license from LGPL-2.1-only to LGPL-2.1+ for Makefile and src/linkcheck.sh, and to put a LGPL-2.1+ license notice to src/uniquefiles.sh.

Once we have Dan's ok, I think we're good to know.

dankegel:
I so agree.  (The original choice was just to match Wine, but that's not a hard requirement.)

jre-wine:
Thanks @dankegel (It now matches Wine's license).
@austin987 Please merge if you are fine with it.

austin987:
Thanks @dankegel!

And thanks @jre-wine for driving this effort!

PR #730, opened 2017-02-10
Unpack "can't" to fix colorer highlighting

birdie-github:
`open FILE, $ARGV[0] || w_die "can't open ".$ARGV[0];` breaks colorer ( http://colorer.sourceforge.net/ ).

Please merge.

### DIFF
--- a/src/winetricks
+++ b/src/winetricks
@@ -1199,7 +1199,7 @@ sub disable_notifications{
 
 # Read the file
 my(%top);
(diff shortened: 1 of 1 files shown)


austin987:
I'm not opposed, but there are other uses of single quotes inside double quotes by w_die(), etc. in winetricks that you may need fixed as well.

Really though, this seems like a bug in colorer.

birdie-github:
Colorer is not actively maintained these days, so please apply this patch ;-) Thank you.

PR #731, opened 2017-02-10
Code reusal for exiting on win64

birdie-github:


austin987:
A) This fails shellcheck, which needs to be fixed.
B) I think consolidating this code is good, but w_die() can be used for things other than the package not working at all. I don't like using a generic name for a specific error. Instead, something like w_die_win64_unsupported_package (that's a bit long, just an idea).

birdie-github:
> I don't like using a generic name for a specific error.

It's actually a _generic error_ since you used the same code at least 10 times. Besides there are and there will be just two X86 architectures, so again it's more than generic. ;-)

austin987:
https://github.com/Winetricks/winetricks/commit/66bc94a419ff641237b9c285065219c42e36cf35

PR #736, opened 2017-02-17
FreeBSD work 1/?

lifanov:
These are low-hanging fruits to use native utilities on FreeBSD.

### DIFF
--- a/src/winetricks
+++ b/src/winetricks
@@ -2733,6 +2733,9 @@ winetricks_get_sha1sum_prog() {
     if [ -x "$(which sha1sum 2>/dev/null)" ]
     then
         WINETRICKS_SHA1SUM="sha1sum"
+    elif [ -x "$(which sha1 2>/dev/null)" ]
+    then
+        WINETRICKS_SHA1SUM="sha1"
     elif [ -x "$(which openssl 2>/dev/null)" ]
     then
         WINETRICKS_SHA1SUM="openssl dgst -sha1"
--- a/src/winetricks
+++ b/src/winetricks
@@ -2702,6 +2702,9 @@ winetricks_dl_url_to_stdout()
                 "$1" > /dev/null
         cat "${stdout_tmpfile}"
         rm "${stdout_tmpfile}"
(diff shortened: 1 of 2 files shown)


src/winetricks
```diff
@@ -942,6 +942,14 @@ w_download_to()
                ${_W_cookiejar:+--cookie "$_W_cookiejar"} \
                ${_W_agent:+--user-agent "$_W_agent"} \
                "$_W_url"
+        elif [ "${WINETRICKS_DOWNLOADER}" = "fetch" ]
+        then
+           # shellcheck disable=SC2086
+           $torify fetch \
+               -T "${WINETRICKS_DOWNLOADER_TIMEOUT}" \
```
austin987: Does fetch work under torify (if torify available on FreeBSD?)

This also lacks support for $WINETRICKS_DOWNLOADER_RETRIES

lifanov:
Hi! torify is available on FreeBSD and fetch does work under it. I'll make the requested changes!
The fetch command does support retrying a download, but not a configurable retry cap.

austin987:
Okay, then please put a comment saying that retries are not configurable.

austin987:
@lifanov looking better (and thanks for the Makefile fix!). ) winetricks_dl_url_to_stdout() still needs fetch support, then I think this should be good.

lifanov:
I'm working on another change, which is fallback to bsdtar for cabextract, unzip, unrar, and 7zip.
This one is really invasive. I would need to abstract calls to all four utilities and systems other than FreeBSD (macOS, some Linux distros) have bsdtar utility. Also, there is no built-in way to change filenames to lowercase, which is done by some cabextract invocations, so I would need to move files in place after extract.

Would this be acceptable?

austin987:
Let's discuss that in a new issue, please. I'll review this and merge in the next few days.

FYI, There's already something similar for using windows 7zip if those utilities aren't available (except cabextract). AFAICT it would mostly be adding the bsdtar fallback in the middle of the current paths (e.g., currently it does if unzip ; else windows 7zip. You'd want if unzip, elif bsdtar ; else windows 7zip).

austin987:
Merged, thanks @lifanov!

PR #744, opened 2017-03-17
convert to sha256

vszakats:

PR #745, opened 2017-03-17
homepage/url-in-comment updates

vszakats:
Secure certain (mostly Microsoft) homepages, follow redirects and some misc URL updates.
This will probably conflict with pending PR #744, I'm going to resolve these once the latter is merged.


vszakats:
Alternatively, `/en-us/` can be removed from all MS URLs, so that the content appears according to browser/OS language preferences.

PR #752, opened 2017-03-24
Update QQ to 8.9.1

hillwoodroc:
Update QQ.


austin987:
Thanks.

For future reference, you can use git push -f (force push) when updating a local branch on pull request to avoid a second PR.

Issue #756, opened 2017-03-29
Winetricks freezes trying to set wine-ver to win2k on dotnet45 installation.


dkanada:
I have tried reinstalling multiple times, but when I try to install dotnet45 winetricks will stop on the following command.

`Executing winetricks_early_wine regedit C:\windows\Temp\_dotnet20sp1\set-winver.reg`

It stops there on bspwm and openbox, which ruled out an issue I had a while ago that was due to bspwm not showing a floating window for one of the dotnet installations.

austin987:
I've seen this as well on Gentoo, with 4.9.7 kernel, gcc 4.9.4, but not consistently. Others I've asked have been unable.

What's your distro/kernel/compiler?

dkanada:
Distro: Arch
Kernel: 4.10.6-1
Compiler: gcc 6.3.1-2

Let me know if you need more information or want me to test something.

austin987:
Please test #757, it's not specifically for this, but it likely avoids this problem.

dkanada:
That fixed my issue, thanks for the help!

austin987:
https://github.com/Winetricks/winetricks/commit/ddf5c7c059ebc9b78031584cb82332e091c52d28 fixes this

PR #757, opened 2017-03-31
Fix multiple 64-bit dotnet issues

austin987:
This should greatly improve the .Net Experience on win64.

### DIFF
--- a/src/winetricks
+++ b/src/winetricks
@@ -6647,7 +6647,7 @@ w_metadata dotnet40 dlls \
 
 load_dotnet40()
 {
-    w_package_unsupported_win64
+    w_package_warn_win64
 
     case "$W_PLATFORM" in
         windows_cmd) ;;
@@ -6693,10 +6693,7 @@ w_metadata dotnet45 dlls \
 
 load_dotnet45()
 {
-    if [ $W_ARCH = win64 ]
-    then
-        w_warn "This package may not work on a 64-bit installation"
-    fi
+    w_package_warn_win64
 
     # The installer will fail later in dotnet35 if cabextract isn't installed, so check/fail now:
     w_try_cabextract -v >/dev/null
@@ -6707,7 +6704,9 @@ load_dotnet45()
     w_call remove_mono
 
     # See https://appdb.winehq.org/objectManager.php?sClass=version&iId=25478 for Focht's recipe
-    w_call dotnet35
+
+    # Seems unneeded in wine-2.0
+    # w_call dotnet35
     w_call dotnet40
     w_set_winver win7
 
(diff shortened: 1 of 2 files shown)


src/winetricks
```diff
@@ -6682,7 +6685,9 @@ load_dotnet45()
     w_call remove_mono
 
     # See https://appdb.winehq.org/objectManager.php?sClass=version&iId=25478 for Focht's recipe
-    w_call dotnet35
+
+    # Seems unneeded in wine-2.0
```
qwertychouskie: As far as I know, this is only fixed in staging, see https://bugs.winehq.org/show_bug.cgi?id=26757.

src/winetricks
```diff
@@ -6736,7 +6741,9 @@ load_dotnet452()
...
```
qwertychouskie: Same here.

src/winetricks
qwertychouskie: As far as I know, this is only fixed in staging, see https://bugs.winehq.org/show_bug.cgi?id=26757.

src/winetricks
```diff
@@ -6622,7 +6622,10 @@ w_metadata dotnet40 dlls \
 
 load_dotnet40()
 {
-    w_package_unsupported_win64
```
qwertychouskie: Maybe add a `w_package_warn_win64`?

src/winetricks
austin987: That bug refers to allowing .msu installers to work. That's unrelated to if .net45 itself can install.

I've never used staging on a regular basis, but I can tell you that wine-2.4-55-g5b4f8a3 + this version of winetricks allows dotnet45 to install (with a 32-bit prefix, previously also tested with win64).

src/winetricks
austin987: See above.

src/winetricks
austin987: Good idea.

austin987:
@qwertychouskie I've added w_package_broken_win64 (broken on wine) and w_package_warn_win64() (may or may not work on win64), and used that for .Net.

I've also re-enabled dotnet46* in the tests, removing dotnet35 seems to have helped there.

I'd still like to see some users test this. I've pinged all related bugs, but no one has given any results yet. For as many people that complained about this, you'd think someone would test it ;)

austin987:
This fixed #756

src/winetricks
qwertychouskie: From my understanding based on bug reports, etc. the issue is that the .NET installers, while using a Wine-compatible install method for that version, it uses MSUs to install the previous versions. In Windows, this means that installing .NET 4.5 installs .NET 4.5, 4.0, 3.5, 3.0, 2.0, and 1.1.  In Wine non-staging (or staging < 1.8), the MSU installer is just a stub that fakes a successful install, so the .NET installation seems to go fine, but is incomplete.  The installer file should have its contents extracted and the MSUs examined to make sure no files are needed from .NET 3.5 or below.

austin987: --verify was used, which means the dotnet_verifier was run. That only checks the given .Net version though, not earlier versions.

fxthomas:
Just tested this because I needed .NET to work on a new Wine prefix. So far I've had no issues!

austin987:
Committed as https://github.com/Winetricks/winetricks/commit/ddf5c7c059ebc9b78031584cb82332e091c52d28

Pirat90pl:
I feel so dumb, dont know how to run it. Teach me please!

PR #766, opened 2017-04-13
windows -> Windows

Faalagorn:
changed "windows" for uppercase "Windows" to match all other entries and allow sorting by title properly

### DIFF
--- a/src/winetricks
+++ b/src/winetricks
@@ -18125,7 +18125,7 @@ _EOF_
 
 w_metadata nt40 settings \
     title_uk="Поставити версію Windows NT 4.0" \
-    title="Set windows version to Windows NT 4.0"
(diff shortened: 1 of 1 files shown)


austin987:
Change is fine, but please use a better commit message (i.e. something like, "nt40: change 'windows' to 'Windows'")

Faalagorn:
Thanks! I figured it would be easy to figure by looking at the files comparison, but on the other hand I guess there shouldn't be need to do so. Thank you for pointing that out, I'll keep that in mind for future :)

cher-nov:
@Faalagorn, I believe, @austin987 meant that you could change commit message by `git commit --amend` and make then a force-push of `Faalagorn:patch-1`.

austin987:
Yes, I did. Thanks for clarifying @cher-nov

PR #782, opened 2017-05-04
fixes two little typos in settings.txt and all.txt

mchlnix:


austin987:
You should fix winetricks itself as well, otherwise this will be overwritten the next time those are updated.

mchlnix:
I see. The GitHub search didn't pick it up in winetricks.

PR #800, opened 2017-05-18
Add 64-bit support for ie8

qwertychouskie:
Experimental, so in as a separate verb.  Eventually will be merged with ie8.

### DIFF
--- a/src/winetricks
+++ b/src/winetricks
@@ -11241,7 +11241,7 @@ load_ie8()
 
     # Put a dummy catalog file in place
     mkdir -p "$W_SYSTEM32_DLLS"/catroot/\{f750e6c3-38ee-11d1-85e5-00c04fc295ee\}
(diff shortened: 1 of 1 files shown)


austin987:
Fails Travis

qwertychouskie:
Fixed.

On 05/18/2017 04:40 PM, Austin English wrote:

Fails Travis

austin987:
Sorry @qwertychouskie, thought I'd sent this earlier. I'm not a fan of making this a separate verb name, then deprecating it. I think you should only merge this when win64 is ready (and it should be within the same verb name, with if "$W_ARCH" as appropriate.

qwertychouskie:
Sorry for the delay, I'll probaby work on this again once IE8 works in current Wine again.

JamieMagee:
Apologies for digging up an old issue, but did you ever get a chance to look at this again @qwertychouskie? The reason I ask, is I think IE8 will help to solve [bug 46313](https://bugs.winehq.org/show_bug.cgi?id=46313).

PR #822, opened 2017-07-31
Add AppStream appdata file

AsciiWolf:
AppStream is a cross-distro XML format to provide metadata for software components and to assign unique identifiers to software. The metadata can for example be used by software centers like GNOME Software or KDE Discover to display a user-friendly application-centric way on the package archive.

This PR adds an AppStream AppData file.

### DIFF
--- a/src/winetricks.appdata.xml
+++ b/src/winetricks.appdata.xml
@@ -1,0 +1,20 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!-- Copyright 2017 Daniel Rusek <mail@asciiwolf.com> -->
+<component type="desktop">
+  <id>winetricks.desktop</id>
+  <project_license>GPL-2.1</project_license>
+  <metadata_license>CC0-1.0</metadata_license>
+  <name>Winetricks</name>
+  <summary>Work around problems and install applications under Wine</summary>
+  <description>
+    <p>
+    Winetricks is an easy way to work around problems in Wine.
+    </p>
+    <p>
+    It lets you install missing DLLs or tweak various Wine settings individually.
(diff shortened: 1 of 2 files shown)


src/winetricks.appdata.xml
```diff
@@ -0,0 +1,20 @@
...
+  <metadata_license>CC0-1.0</metadata_license>
+  <name>Winetricks</name>
+  <summary>Work around problems and install applications under Wine</summary>
+  <description>
+    <p>
+    Winetricks is an easy way to work around problems in Wine.
+    </p>
+    <p>
+    It has a menu of supported games/apps for which it can do all the workarounds automatically.
+    It also lets you install missing DLLs or tweak various Wine settings individually.
```
austin987: I'd prefer if the order of these were flipped. Winetricks is mostly used for DLLs/settings, not apps/games.

AsciiWolf:
@austin987 I cannot make a screenshot right now, but it simply makes possible to install Winetricks using AppStream-based package managers. Without the appdata file, Winetricks cannot be found or installed using, for example, GNOME Software or KDE Discover. Other modern package managers are all starting to use AppStream because it is a [standard](https://www.freedesktop.org/wiki/Distributions/AppStream/). Debian or Ubuntu has AppStream Generator that generates the appdata file from a desktop file, but most distributions don't have such thing and require the appdata file to be provided by upstream.

austin987:
K, thanks. Please see my other comment.

src/winetricks.appdata.xml
AsciiWolf: Done.

austin987:
Merged, thanks.

PR #826, opened 2017-08-17
replace "sort -V" with "sort -n -t. -k1 -k3"

k3erg:
"sort -V" is not supported on OS X and older coreutils. "sort -n -t. -k1 -k3" supports sorting of "abc x.x.x".

### DIFF
--- a/src/winetricks
+++ b/src/winetricks
@@ -2422,7 +2422,7 @@ w_compare_wine_version()
         *) w_die "Unsupported comparison. Only -ge and -le are supported" ;;
     esac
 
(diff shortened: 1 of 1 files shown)


k3erg:
Please check and add usecases, if other than "abc x.x.x". because "2.0-rc1" would be sorted after "2.0".

austin987:
winetricks-test w_compare_wine_version passes with it. While the rc stuff isn't perfect, it's not a common problem, I'd rather portability over 100% correctness.

Thanks.

Issue #846, opened 2017-10-07
URL fixes/updates


vszakats:
```patch
diff --git a/src/winetricks b/src/winetricks
index b28aee8..16aa8a7 100755
--- a/src/winetricks
+++ b/src/winetricks
@@ -5374,7 +5374,7 @@ w_metadata binkw32 dlls \
 
 load_binkw32()
 {
-    # Mirror: http://www.dlldump.com/download-dll-files_new.php/dllfiles/B/binkw32.dll/1.0q/download.html
+    # Mirror: https://www.dlldump.com/download-dll-files_new.php/dllfiles/B/binkw32.dll/1.0q/download.html
     # sha256sum of the decompressed file: 1fd7ef7873c8a3be7e2f127b306d0d24d7d88e20cf9188894eff87b5af0d495f
     #
     # Zip sha256sum:
@@ -7097,7 +7097,7 @@ load_dotnet462()
     w_package_warn_win64
 
     if w_workaround_wine_bug 42170 "Running un-official repacked .NET 4.6.2 setup until the official version is fixed."; then
-        # Un-official slim version. See http://repacks.net/forum/viewtopic.php?t=7
+        # Un-official slim version. See https://repacks.net/forum/viewtopic.php?t=7
         file_package="dotNetFx462_Full_x86_x64_Slim.exe"
         w_download https://www.dropbox.com/s/1bdocgbate8hc8z/dotNetFx462_Full_x86_x64_Slim.exe?dl=1 8f03c8a9b2f4666c3707abe2d70145f200cb2f0f3b792626e64d167d1ad0565f $file_package
         unattended_args="/ai /gm2"
@@ -7440,7 +7440,7 @@ load_gmdls()
 {
     # Originally at https://download.microsoft.com/download/whistler/Update/8.1/W982KMeXP/EN-US/DX81Redist.exe
     # Only archive.org seems to have it now
-    w_download_to directx8 http://web.archive.org/web/20070105100243if_/http://download.microsoft.com:80/download/whistler/Update/8.1/W982KMeXP/EN-US/DX81Redist.exe 5ddc1a8e204381254dc5d65f406584787155983adf245a75000dcd0d2efb04c6
+    w_download_to directx8 https://web.archive.org/web/20070105100243if_/download.microsoft.com/download/whistler/Update/8.1/W982KMeXP/EN-US/DX81Redist.exe 5ddc1a8e204381254dc5d65f406584787155983adf245a75000dcd0d2efb04c6
 
[...]

PR #853, opened 2017-10-11
Secure URLs

vszakats:


vszakats:
Ref: https://github.com/Winetricks/winetricks/issues/846

Issue #831, opened 2017-08-27
winhttp breaks after upgraded to win7sp1


oott123:
[This commit](https://github.com/Winetricks/winetricks/commit/17dd1fb82953b40c8a5167d8a68249589c1acd34) upgrade winhttp from win2ksp4 to win7sp1.

However, this version of winhttp requires `webio.dll` which is not extracted from the cab.

It says `err:module:import_dll Library webio.dll (which is needed by L"C:\\windows\\system32\\WINHTTP.dll") not found` when using winhttp and then fallback to wine dlls.
There is [another report at winehq forum](https://forum.winehq.org/viewtopic.php?f=8&t=29291) which meet the same error as me.

I copied webio.dll to wine, but it not works. Maybe we should rollback to win2ksp4 version ...

austin987:
Should be fixed by https://github.com/Winetricks/winetricks/commit/254cb28d4e9d6a084b41186f39670655691c5c41

Sorry about that.

vitor251093:
I receive this exact error message, but running Uplay with a Windows 7 prefix. I guess reverting that commit fixed the problem for Windows Vista and lower versions, but broke it for Windows 7 and superior versions :P 

Since Windows 7 is the default prefix for Wine now, maybe the most proper solution would be to add a condition which checks the Windows versions, or maybe revert the reversion :P

austin987:
@vitor251093 please file a new issue.

I think the proper solution is to include both, under two names. I.e., winhttp / winhttp_win7

vitor251093:
Uhn, maybe. But considering other winetricks, like gdiplus, wouldn't be better to give the suffix to the older version? Like winhttp_winxp? 

I will file a new issue like you instructed.

PR #855, opened 2017-10-12
use win2ksp4 winhttp again

austin987:


austin987:
This reverts commit 17dd1fb.

It breaks things, see #831

Simply adding webio isn't enough, then it fails in a misisng ntdll function.
While we should prepare for when win2ksp4 isn't available, it currently is, so revert.

austin987:
For some reason, Linux Travis CI is hanging; but it passes OSX and for me locally, so calling it their problem.

PR #863, opened 2017-11-01
Add a couple of missing double semicolons

mihaitodor:
I spotted these while digging through the script in Sublime because they confuse the syntax highlighter.


austin987:
Nice catch, thanks.

Issue #861, opened 2017-10-30
w_register_font doesn't back up reg files properly (--no-cleanup)


mihaitodor:
Currently, the code is using `$$` to get the process ID, which gets appended to the backed up `.reg` file name [here](https://github.com/Winetricks/winetricks/blob/24c573368a5dce0a81b4e0ca6c66be9ae3cbfe3d/src/winetricks#L2764) and [here](https://github.com/Winetricks/winetricks/blob/24c573368a5dce0a81b4e0ca6c66be9ae3cbfe3d/src/winetricks#L2775). This is not working well when registering multiple fonts, like, for example, with `winetricks corefonts`, because we'll get only the last two `.reg` files in `$W_TMP_EARLY`.

PS: `w_register_font_replacement` doesn't seem to backup the generated `.reg` file to `$W_TMP_EARLY`

PR #864, opened 2017-11-02
Backup font reg files properly

mihaitodor:
This is an attempt to resolve https://github.com/Winetricks/winetricks/issues/861 I don't know any highly portable solution for generating random numbers in shell scripts, so I leveraged the existing `w_get_sha256sum` verb in order to get a unique string to append to the name of each file that gets placed in `$W_TMP_EARLY`.

I didn't want to append the font name to the file name because it might contain weird characters, but I'm OK to do that instead of the truncated sha256 if you think it's safe enough.

### DIFF
--- a/src/winetricks
+++ b/src/winetricks
@@ -2741,14 +2741,25 @@ w_call()
     w_try w_do_call "$@"
 }
 
+w_backup_reg_file()
+{
+    W_reg_file=$1
+
+    w_get_sha256sum "$W_reg_file"
+
+    w_try cp "$W_reg_file" "$W_TMP_EARLY/_reg_$(echo "$_W_gotsha256sum" | cut -c1-8)"_$$.reg
+
+    unset W_reg_file _W_gotsha256sum
+}
+
 w_register_font()
 {
-    file=$1
+    W_file=$1
     shift
-    font=$1
+    W_font=$1
 
-    case "$file" in
-        *.TTF|*.ttf) font="$font (TrueType)";;
+    case "$W_file" in
+        *.TTF|*.ttf) W_font="$W_font (TrueType)";;
     esac
 
     # Kludge: use _r to avoid \r expansion in w_try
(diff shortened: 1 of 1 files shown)


src/winetricks
```diff
@@ -2741,38 +2741,51 @@ w_call()
     w_try w_do_call "$@"
 }
 
+w_backup_reg_file()
+{
+    W_reg_file=$1
+
+    w_get_sha256sum "$W_reg_file"
+
+    cp "$W_reg_file" "$W_TMP_EARLY"/_reg_"$(echo "$_W_gotsha256sum" | cut -c1-8)"_$$.reg
```
austin987: Please run under w_try (or explain why you can't).

austin987:
I think appended sha256 is fine, there's an extremely low risk of conflict with that. The alternative is to dd from /dev/urandom, but that would be more complicated for what's very very unlikely to happen.

After adding w_try, please squash all commits, and add a more descriptive commit message (you can force push to your personal github fork). Something like:

 w_register_font/w_register_font_replacement: properly backup font reg files

Thanks for working on this btw!

src/winetricks
mihaitodor: Nice catch! Fixed.

mihaitodor:
OK @austin987, I have made the changes you requested.

> Thanks for working on this btw!
You're welcome and thanks for maintaining this project :)

src/winetricks
```diff
@@ -2741,38 +2741,51 @@ w_call()
     w_try w_do_call "$@"
 }
 
+w_backup_reg_file()
+{
+    W_reg_file=$1
+
+    w_get_sha256sum "$W_reg_file"
+
+    w_try cp "$W_reg_file" "$W_TMP_EARLY"/_reg_"$(echo "$_W_gotsha256sum" | cut -c1-8)"_$$.reg
```
austin987: Sorry I missed this before, but the quoting could be simplified, remove the quote after early and _reg_.

After that, ready for merge!

mihaitodor: Yes, well spotted. I fixed it.

austin987:
Thanks!

PR #876, opened 2017-11-28
Update netfx_setupverifier_new.zip URL

naderhen:
This was breaking on CI for us, looks like the URL has been updated


austin987:
Thanks

Issue #885, opened 2017-12-06
winetricks_is_installed doesn't work if the verb doesn't create any new files


mihaitodor:
I was reviewing [`winetricks_is_installed`](https://github.com/Winetricks/winetricks/blob/389d1ed6810cab5a54be2d68c2c6b4752a04f272/src/winetricks#L3904), which looks at the contents of `installed_file1` in `w_metadata` and, if that file is present, it returns true. However, this prevents verbs which only overwrite existing files from running (try running `eufonts` after `corefonts` when #865 gets fixed).

What would be a good way to fix this? Would you consider parsing `winetricks.log` (which gets created by `winetricks_stats_log_command`) in `winetricks_is_installed` to determine if a verb has been executed before?

austin987:
@mihaitodor I think parsing winetricks.log is a good solution.

Issue #865, opened 2017-11-07
eufonts doesn't overwrite corefonts, creating duplicates


mihaitodor:
The `corefonts` package copies the following fonts under `.wine/drive_c/windows/Fonts/`:
```
Arialbd.TTF
Arialbi.TTF
Ariali.TTF
Arial.TTF
AriBlk.TTF
Comicbd.TTF
Comic.TTF
courbd.ttf
courbi.ttf
couri.ttf
cour.ttf
Georgiab.TTF
Georgiai.TTF
Georgia.TTF
Georgiaz.TTF
Impact.TTF
Timesbd.TTF
Timesbi.TTF
Timesi.TTF
Times.TTF
Trebucbd.ttf
trebucbi.ttf
trebucit.ttf
trebuc.ttf
Verdanab.TTF
Verdanai.TTF
Verdana.TTF
Verdanaz.TTF
Webdings.TTF
```

Because some of their extensions have caps (`.TTF`), the `eufonts` package (which gets installed by the `allfonts` package) does not overwrite them, since it contains the following font files:
```
ArialBd.ttf
ArialBI.ttf
ArialI.ttf
Arial.ttf
TimesBd.ttf
TimesBI.ttf
TimesI.ttf
Times.ttf
trebucbd.ttf
trebucbi.ttf
trebucit.ttf
trebuc.ttf
Verdanab.ttf
Verdanai.ttf
Verdana.ttf
Verdanaz.ttf
```

Since we'll end up with 2 copies of the same file under `.wine/drive_c/windows/Fonts/`, I'm not sure which one Wine will pick up.

austin987:
@mihaitodor I'm beginning to think that the simplest fix would be to forcibly rename all fonts to lowercase (in all verbs). Given that you work with fonts more extensively than I, do you see any issue with that approach?

mihaitodor:
I have thought about this option too and it's probably a good idea. I doubt Windows / Wine cares about font names having upper or lower case characters, but I can't guarantee. I don't mind doing it and asking for forgiveness later, if you fancy this solution :)

Still, there is still room for error if the user installs some font packages without using winetricks and then tries to install some other font packages via winetricks. To address this issue in a robust manner, my idea is to do something like this:

```shell
w_try_force_cp_files()
{
    # $1 - source files
    # $2 - destination directory

    _W_src_files="$1"
    _W_dest_dir="$2"
    shift 2

    if test -z "$_W_src_files"; then
[...]

PR #888, opened 2017-12-07
Fix font duplication

mihaitodor:
This is an attempt to partially address #865 and clean up some inconsistencies in the font-related verbs.

- Add verb `w_try_cp_font_files`, which makes sure that we don't get the same font file but with different casing under `C:\windows\Fonts`. It also forces lower case for font names to help alleviate duplication issues.
- Cache PowerPointViewer.exe in "$W_CACHE/PowerPointViewer" for consistency.
- Remove "# FIXME: why is this commented out? Should be removed or enabled." comment. This was referring to the Andalé Mono font, which is [no longer distributed with Windows](https://en.wikipedia.org/wiki/Andal%C3%A9_Mono#Distribution). If you wish, I can create a separate verb for it since it's available [here](https://mirrors.kernel.org/gentoo/distfiles/andale32.exe), but I think we don't need it at all, since it's almost identical to [Courier](https://en.wikipedia.org/wiki/Courier_(typeface)).
- Remove `chmod +w "${W_FONTSDIR_UNIX}"/tahoma*.ttf` since no app should be allowed to overwrite fonts. I'd rather have this posted as a workaround on https://appdb.winehq.org/ instead of maintaining it in Winetricks.
- Place font package URLs in quotes and leave them hardcoded in the verb body because they're much easier to grep for and replace that way.
- `w_try_cabextract` already passes `-q` to cabextract.

Note: Since eufonts [only updates trebucbd.ttf](https://github.com/Winetricks/winetricks/blob/389d1ed6810cab5a54be2d68c2c6b4752a04f272/src/winetricks#L10271), allfonts will skip installing eufonts after this PR, since trebucbd.ttf is installed in advance by corefonts. #885 needs to be addressed in order to fix this.

Note2: eufonts is an update to some of the fonts installed by corefonts, so corefonts should always be installed first. Luckily, allfonts installs them in the right order.

As a reference, here is the difference between installed font files based on this PR:
[original_font_files.txt](https://github.com/Winetricks/winetricks/files/1539318/original_font_files.txt)
[updated_font_files.txt](https://github.com/Winetricks/winetricks/files/1539320/updated_font_files.txt)

And here is the difference in registry entries:
[original_windows_nt.txt](https://github.com/Winetricks/winetricks/files/1539346/orig_windows_nt.txt)
[updated_windows_nt.txt](https://github.com/Winetricks/winetricks/files/1539350/updated_windows_nt.txt)

The observed differences in the registry are due to the fact that eufonts was not installed as explained above.

As described in CONTRIBUTING.md (#871), I ran `./tests/shell-checks` successfully and I skipped `winetricks-test`, since it didn't work as described. See [here](https://github.com/Winetricks/winetricks/pull/871#discussion_r155545701).

### DIFF
--- a/src/winetricks
+++ b/src/winetricks
@@ -514,6 +514,49 @@ w_try_unzip()
     w_try "$WINE" "$W_PROGRAMS_X86_WIN\\7-Zip\\7z.exe" x "$(w_pathconv -w "$zipfile")" -o"$(w_pathconv -w "$destdir")" "$@"
 }
 
+# Copy font files matching a glob pattern from source directory to destination directory.
+# Also remove any file in the destination directory that has the same name as
+# any of the files that we're trying to copy, but with different case letters.
+# Note: it converts font file names to lower case to avoid inconsistencies due to paths
+#       being case-insensitive under Wine.
+w_try_cp_font_files()
+{
+    # $1 - source directory
+    # $2 - destination directory
+    # $3 - optional font file glob pattern (default: "*.ttf")
+
+    _W_src_dir="$1"
+    _W_dest_dir="$2"
+    _W_pattern="$3"
+    shift 2
+
+    if test ! -d "$_W_src_dir"; then
+        w_die "bug: missing source dir"
+    fi
+
+    if test ! -d "$_W_dest_dir"; then
+        w_die "bug: missing destination dir"
+    fi
+
+    if test -z "$_W_pattern"; then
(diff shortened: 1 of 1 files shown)


mihaitodor:
Looks like something went wrong on Travis for the OSX build:
> /Users/travis/.travis/job_stages: line 57: pip: command not found

The Linux build was successful.

PR #893, opened 2017-12-15
winetricks_is_installed: return false if the verb hasn't been run before

### LINKED ISSUE #885: winetricks_is_installed doesn't work if the verb doesn't create any new files

mihaitodor:
This fixes #885.

With this fix, eufonts is no longer skipped when installing allfonts (a regression introduced by #888).

### DIFF
--- a/src/winetricks
+++ b/src/winetricks
@@ -3957,6 +3957,12 @@ winetricks_is_installed()
         return 1  # not installed
     fi
 
+    # Test if the verb has been executed before
+    if ! grep -qw "$1" "$WINEPREFIX/winetricks.log"; then
+        unset _W_file
+        return 1  # not installed
+    fi
+
     case "$W_PLATFORM" in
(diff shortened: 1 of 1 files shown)


mihaitodor:
@austin987 I rebased this PR to pass the Travis build.

PR #898, opened 2017-12-19
Add PowerPoint Viewer fonts

### LINKED ISSUE #883: Not all fonts from PowerPointViewer.exe are available as verbs
`PowerPointViewer.exe` contains additional fonts that I think winetricks should register: [Calibri](https://en.wikipedia.org/wiki/Calibri), [Candara](https://en.wikipedia.org/wiki/Candara), [Corbel](https://en.wikipedia.org/wiki/Corbel_(typeface)) and [Meiryo](https://en.wikipedia.org/wiki/Meiryo). Is there any reason not to register them?

mihaitodor:
This fixes #883.

It adds the following verbs: load_calibri, load_candara, load_corbel, load_meiryo, load_pptfonts. The hindmost one installs all the fonts that PowerPointViewer.exe contains.


src/winetricks
```diff
@@ -10288,6 +10288,25 @@ load_meiryo()
...
+    media="download"
+
+load_pptfonts()
+{
+    w_call calibri
+    w_call cambria
+    w_call candara
+    w_call constantia
+    w_call corbel
+    w_call consolas
```
austin987: consolas should be before constantia

mihaitodor: @austin987 Nice catch! Fixed.

PR #904, opened 2017-12-20
load_cambria: Register Cambria Math

### LINKED ISSUE #881: Cambria font not registered as "Cambria Math"
On Windows, `CAMBRIA.TTC` is registered as `Cambria & Cambria Math (TrueType)`. Winetricks simply [registers it](https://github.com/Winetricks/winetricks/blob/389d1ed6810cab5a54be2d68c2c6b4752a04f272/src/winetricks#L10089) as `Cambria (TrueType)`. I am not sure if this is supported under Wine, but I think we should register it in the same way. WDYT?

mihaitodor:
This fixes #881, and it's probably relevant only for apps which list fonts manually from the registry. Wine seems to return the fonts that are under `HKCU\Software\Wine\Fonts\Cache` when calling `EnumFontFamiliesEx` and it creates both `Cambria` and `Cambria Math` entries automatically, which is great because it seems to support TTC files quite well:

screen shot 2017-12-19 at 23 33 42
screen shot 2017-12-19 at 23 35 33


austin987:
Have you looked at what windows does? I.e., is it one entry or two there?

mihaitodor:
Just one entry, like I said in #881. See here:

screen shot 2017-12-20 at 00 06 18

PR #905, opened 2017-12-20
load_fakejapanese_vlgothic: Deprecate and forward to load_meiryo

### LINKED ISSUE #899: load_fakejapanese_vlgothic should be removed after #883
load_fakejapanese_vlgothic substitutes Meiryo and Meiryo UI with VL Gothic, which I think we don't need any more, since #883 will make the Meiryo fonts available. If we leave it in there, running allfonts will load this verb automatically, which is probably not what the users want.

Font substitutes are documented, albeit sketchy, [here](https://technet.microsoft.com/en-us/library/cc939627.aspx).

LE: I think Wine [will ignore](https://github.com/wine-mirror/wine/blob/7183dc7360f1f7a514f9649fced4171175f16f06/dlls/gdi32/freetype.c#L2408) the substitute if Meiryo is installed, but not sure how to test this. WDYT?

mihaitodor:
This fixes #899.

I think it's safe to skip it in `load_allfonts`. Otherwise, users will see the deprecation warning when they install allfonts.

### DIFF
--- a/src/winetricks
+++ b/src/winetricks
@@ -10563,17 +10563,14 @@ w_metadata fakejapanese_vlgothic fonts \
 
 load_fakejapanese_vlgothic()
 {
-    w_call vlgothic
+    w_warn "load_fakejapanese_vlgothic is deprecated, now only calls load_meiryo and adds an alias for the Japanese spelling of Meiryo"
 
-    # Aliases to set:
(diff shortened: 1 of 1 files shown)


mihaitodor:
Closing this since some users require this verb. Details in #899.

Issue #902, opened 2017-12-19
Use full path to fonts fix is broken


mihaitodor:
The fix added by 2c2b44ee has some issues with Cambria and it registers it like this:
```
REGEDIT4

[HKEY_LOCAL_MACHINE\Software\Microsoft\Windows\CurrentVersion\Fonts]
"Cambria Italic (TrueType)"="c:\\windows\\Fonts"
```

Here's the relevant debugging info from `winetricks -v cambria`:
```
+ w_register_font cambria.ttc Cambria
+ W_file=cambria.ttc
+ shift
+ W_font=Cambria
+ echo cambria.ttc
+ tr [:upper:] [:lower:]
+ W_font=Cambria (TrueType)
+ [ ! -f cambria.ttc ]
+ [ -f /home/mtodor/.wine/dosdevices/c:/windows/Fonts/cambria.ttc ]
+ W_file_unescaped=c:\windows\Fonts\cambria.ttc
+ echo c:\windows\Fonts\cambria.ttc
+ sed -e s!\\!\\\\!g
+ W_file=c:\\windows\\Fonts
...
```

I'm not yet sure how to fix it, but I'm also wondering if there is any value in using the absolute path to these font files. For example, my app seems to pick them up fine from `C:\windows\Fonts` if only the file name is present in the registry. Not sure about other apps, though.

mihaitodor:
Fixed by #907

PR #907, opened 2017-12-20
w_register_font: Revert 2c2b44e because it's not needed

mihaitodor:
I think it's best to revert this fix, because [according to Huw](https://www.winehq.org/pipermail/wine-devel/2017-December/120684.html):
> Fonts without a path correspond to c:\windows\fonts.  These files
> should be installed into that directory.

Also, it doesn't seem to work well on my Ubuntu installation. See #902.


austin987:
Yeah, I misunderstood things, thanks.

PR #908, opened 2017-12-23
Font replacements related fixes

kakurasan:
* allfonts: don't call verbs for font replacements (fake*) or "pptfonts" metapackage
    * This prevents `allfonts` from adding font replacements
    * `pptfonts` can be skipped because it only calls other verbs for installing fonts
* cjkfonts: call fakejapanese_vlgothic
    * Now it adds font replacements for Meiryo
* fakejapanese_vlgothic, meiryo: conflict each other
    * Wine uses Meiryo when both are installed

winetricks.log after installing patched `allfonts`:
```baekmuk
calibri
cambria
candara
consolas
constantia
corbel
corefonts
droid
eufonts
ipamona
liberation
lucida
meiryo
opensymbol
tahoma
takao
uff
unifont
vlgothic
wenquanyi
allfonts
```

winetricks.log after installing patched `cjkfonts`:
```wenquanyi
fakechinese
takao
fakejapanese
vlgothic
fakejapanese_vlgothic
baekmuk
fakekorean
unifont
cjkfonts
```

### DIFF
--- a/src/winetricks
+++ b/src/winetricks
@@ -10282,6 +10282,7 @@ w_metadata meiryo fonts \
     publisher="Microsoft" \
     year="2009" \
     media="download" \
+    conflicts="fakejapanese_vlgothic" \
     file1="PowerPointViewer.exe" \
     installed_file1="$W_FONTSDIR_WIN/meiryo.ttc"
 
@@ -10566,6 +10567,7 @@ load_fakejapanese_ipamona()
 w_metadata fakejapanese_vlgothic fonts \
     title="Creates aliases for Japanese Meiryo fonts using VLGothic fonts" \
     publisher="Project Vine / Daisuke Suzuki" \
+    conflicts="meiryo" \
     year="2014"
 
 load_fakejapanese_vlgothic()
--- a/src/winetricks
+++ b/src/winetricks
@@ -10132,6 +10132,7 @@ load_cjkfonts()
 {
     w_call fakechinese
     w_call fakejapanese
+    w_call fakejapanese_vlgothic
     w_call fakekorean
     w_call unifont
 }
--- a/src/winetricks
+++ b/src/winetricks
@@ -10908,7 +10908,12 @@ load_allfonts()
     do
         cmd=$(basename "$file" .vars)
         case $cmd in
-            allfonts|cjkfonts) ;;
+            # "fake*" verbs need to be skipped because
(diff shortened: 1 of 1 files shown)


mihaitodor:
Looks good to me! This is probably good enough to cover what I was asking for in #899.

mihaitodor:
@austin987 If you get a chance, it would be great if you have a look at this PR. It makes `allfonts` more consistent.

src/winetricks
```diff
@@ -10908,7 +10911,7 @@ load_allfonts()
     do
         cmd=$(basename "$file" .vars)
         case $cmd in
-            allfonts|cjkfonts) ;;
+            allfonts|cjkfonts|fake*|pptfonts) ;;
```
austin987: As this list is growing, please add a comment explaining why they're being skipped.

Other than that, LGTM

wdlkmpx:
i remember having an issue with 'allfonts' 2 weeks ago. 

this is probably the fix the world is waiting for. there is a conflict in 'allfonts' which prevents the installation from finishing. the world is waiting..

mihaitodor:
@wdlkmpx I haven't seen any conflict that prevents `allfonts` from installing. This fix simply makes `allfonts` install fonts only. Users can then install any font substitutes on top of it, if they need to.

Without this PR, `allfonts` installs, besides all fonts, all the available font substitutes, which overlap with each other (like, for example, `load_fakejapanese` and `load_fakejapanese_ipamona`), causing a bit of a mess. Hopefully @kakurasan will add the requested changes to this PR soon, so it can be merged.

Please open a bug if you get a reproducible crash while installing fonts.

kakurasan:
> Hopefully @kakurasan will add the requested changes to this PR soon, so it can be merged.

I already added a comment that explains why they're being skipped, and pushed it last week. Now I look forward to merging (or any additional feedback).

mihaitodor:
Oops, my bad. I didn't notice that. @austin987 would you mind having a look at the updated PR?

austin987:
Merged, thanks 

In the future, please ping me when you've force pushed updates, GitHub doesn't send notifications for that, unfortunately.

kakurasan:
> In the future, please ping me when you've force pushed updates, GitHub doesn't send notifications for that, unfortunately.

OK, I'll add a comment after updating, in the future.

PR #913, opened 2017-12-29
Fix #912

### LINKED ISSUE #912: Gksu is deprecated, should use pkexec instead
See https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=885548 for background.

Pull requests appreciated, otherwise I'll look at the end of January when I get back from holiday.

kakurasan:
Please note that I can't test this because I don't have any discs that can be used with Winetricks.


austin987:
Please see discussion on #915

PR #915, opened 2017-12-31
Simple update of gksu to pkexec

ShannonVanWagner:
Seems usage of gksu is very specific so drop-in replacement with pkexec seems like it works fine.

### DIFF
--- a/src/winetricks
+++ b/src/winetricks
@@ -3132,8 +3132,8 @@ winetricks_detect_sudo()
         elif test -x "$(which kdesudo 2>/dev/null)"; then
             WINETRICKS_SUDO=kdesudo
         # fall back to the su versions if sudo isn't available (Fedora, etc.):
-        elif test -x "$(which gksu 2>/dev/null)"; then
-            WINETRICKS_SUDO=gksu
+        elif test -x "$(which pkexec 2>/dev/null)"; then
+            WINETRICKS_SUDO=pkexec
         elif test -x "$(which kdesu 2>/dev/null)"; then
             WINETRICKS_SUDO=kdesu
         fi


ShannonVanWagner:
Looks like someone already beat me to this.

ShannonVanWagner:
Wait a sec... are we updating gksudo or just gksu? I was thinking the latter..

kakurasan:
* The `pkexec` command is part of `policykit-1` package on Debian-based distros.

```
Recommends: zenity | kdebase-bin,
        xdg-utils,
        sudo | policykit-1,
        wine
```

* The `gksudo` command is also part of `gksu` package. We should not use this as well.

```
$ dpkg-query --listfiles gksu | grep bin/                                      
/usr/bin/gksu                                                                  
/usr/bin/gksudo
```

austin987:
@ShannonVanWagner/@kakurasan/@jre-wine this is the patchset I'm considering pushing:
https://github.com/austin987/Winetricks/tree/sudo

Please review.

@kakurasan regarding your kdesu* changes, did you actually test that on kde? I don't run kde so I can't test, so I don't want to commit if it's untested.

I'm attaching a .verb file I used for testing this, which downloads and mounts a ubuntu ISO.

austin987:
Upload failed, here it is inline:

```
w_metadata ubuntu apps \
    title="Ubuntu test" \
    publisher="canonical" \
    year="2016" \
    media="cd" \
    file1="ubuntu-16.04.3-desktop-i386.iso" \
    installed_exe1="C:/md5sum.txt"

load_ubuntu()
{
    # I had issues caching discs, but that's a separate bug. Since we can just download ubuntu, do so
    # Note: downloading to $(volname).iso
    w_download "http://releases.ubuntu.com/16.04/ubuntu-16.04.3-desktop-i386.iso" e28a99e1754a5b04741cd09ff557df2173a0177df79782aa0eb952d3d9803a2b "Ubuntu 16.04.3 LTS i386.iso"
    w_mount "Ubuntu 16.04.3 LTS i386"
    w_info "W_ISO_MOUNT_ROOT is $W_ISO_MOUNT_ROOT"
    w_try cp "$W_ISO_MOUNT_ROOT/md5sum.txt" "$W_DRIVE_C/md5sum.txt"
}
```

kakurasan:
> regarding your kdesu* changes, did you actually test that on kde? I don't run kde so I can't test, so I don't want to commit if it's untested.

I thought I checked command line arguments handling of kdesu/kdesudo on Plasma 5, but I noticed that kdesu can't execute quoted command line string. Sorry.

By the way, I think we can also remove kdesu/kdesudo support because:

* pkexec is desktop-independent and works on KDE Plasma
* I guess pkexec is now available on most Linux distros
  (in Ubuntu, `policykit-1` is available as of 14.04 LTS)
* Winetricks warns when Wine version is older than 2.0,
  that is released in Jan 2017. It seems Winetricks
  expects users to use recent distros.

jre-wine:
Yes, it works. But for running @austin987 's example I had to give my password 4 or 5 times for each separate command. I guess we'd need some kind of special polkit configuration for handling winetricks to avoid that. Unfortunately I don't know much about that, so can't propose something.

In any way, since this is probably not a heavily used feature, I see this as a sufficient solution for Debian with the current gksu removal bug.

Besides that:
Can you mention pkexec in src/winetricks line38 (btw, not sure if the described use case is the only one, maybe make it a bit more general)
`# - sudo is used to mount .iso images if the user cached them with -k option.`

I also don't use KDE, but after a quick web search I guess it's still the recommended frontend for KDE. I think as winetricks upstream you should keep support for all frontends, kdesu(do) but also the deprecated gksu(do), in order to keep a maximum of compatibility both for very old systems, but eventually also for non-Linux systems or whatever reasons there exist to not use policykit (otoh the command-line sudo is there as last resort in any way). Personally I love getting rid of old stuff.

I also wondered if there should be a deprecation warning, but I think it's not necessary for winetricks to give one, since this is a central piece of the desktop environments, and thus its deprecation should be handled by the DE anyway.

d/control: I think we should separate the dependency on sudo from the dependencies for graphical tools. AFAICS src/winetricks defaults to sudo, and only chooses a graphical sudo if it's running in GUI mode itself. In any case policykit-1 should have the highest priority between the graphical frontends (you had that correct).  In the debian.org packages there will be of course no dependency on gksu.
```
Recommends:
 sudo,
 policykit-1 | kdesudo | gksu,
```
Note: In Debian I found no /usr/bin/kdesu (only a libexec in package kde-cli-tools), but /usr/bin/kdesudo does exist (in the package kdesudo) - so I also added that new dependency here.

jre-wine:
```
 policykit-1 | kdesudo,
 sudo,

```
... so that it stays sorted alphabetically.

austin987:
Thanks for the feedback, I'll push an update tonight or tomorrow morning.

austin987:
Thanks for your feedback @kakurasan / @jre-wine.

I think jre has good points, and gksu*/kdesu* should stay (at least for a while).

@jre-wine, some comments in reply:
* this is rarely used, currently only by some games winetricks supports, to mount their ISOs.
* gksudo also requires multiple passwords. I think we may be able to avoid it by using an 'action definition file', though I'm not sure it's worth the effort given how rarely this is used.

Honestly, I'd prefer removing all of the iso handling stuff, but I don't want to remove code as long as it's working.

austin987:
Branch updated. I think I address all of your concerns @jre-wine, but please let me know if there's anything I can improve further.

kakurasan:
@austin987 There are still some problems (unreachable code, and broken kdesudo/kdesu support), but I found that `unar` command (package `unar` on Debian-based distros, also available on Mac) can unpack ISO disc images. I hope this helps.

```
$ unar -o /path/to/output_dir /path/to/image.iso
```

austin987:
Could you explain on that? What makes you think kde is broken? I thought you hadn't tested it..

jre-wine:
@austin987 actually I don't see any changes in branch sudo, except that you dropped "misc: quote commands for kdesu*, not gksudo". Am I looking at the wrong place, or did you push the wrong changeset?

kakurasan:
> Could you explain on that? What makes you think kde is broken? I thought you hadn't tested it..

The problem is the command line arguments handling.

When I forced to use `kdesudo` (e.g. temporarily adding `WINETRICKS_SUDO=kdesudo` to the end of `winetricks_detect_sudo()`), I got:

```
Executing w_do_call ubuntu                                                     
Executing load_ubuntu                                                          
Running kdesudo umount /mnt/winetricks                                         
kdesudo: Unknown option 'l'.                                                   
kdesudo: Use --help to get a list of available command line options.
Executing kdesudo rm -rf /mnt/winetricks                                       
rm: cannot remove '/mnt/winetricks': Is a directory                            
Note: command kdesudo rm -rf /mnt/winetricks returned status 1. Aborting.
```
```
$ kdesudo --version                                                            
Qt: 4.8.7                                                                      
KDE Development Platform: 4.14.34
KdeSudo: 3.4.2.3
```

And `kdesu`:

```
Executing w_do_call ubuntu
Executing load_ubuntu
Running kdesu umount /mnt/winetricks
Unknown option 'l'.
Executing kdesu rm -rf /mnt/winetricks
org.kde.kdesu: Config file not found:  ""                                      
                             writable                                          
  -u <user>                  Specifies the target uid
  ...
  --desktopfile <file name>  The base file name of the desktop entry for this
                             application.

Arguments:
  command                    Specifies the command to run as separate arguments
Note: command kdesu rm -rf /mnt/winetricks returned status 1. Aborting.        
```
```
$ /usr/lib/x86_64-linux-gnu/libexec/kf5/kdesu --version                        
kdesu 5.10.5
```

Note that `kdesu` is located in `/usr/lib/<ARCH_NAME>/libexec/kf5/` on Debian-based distros (`/usr/bin` on Mageia) and needs to use `-c "<command line string>"` option. For example:

```
$ /usr/lib/x86_64-linux-gnu/libexec/kf5/kdesu -c "mkdir -p /a/b/c"
```

And I considered doing like this (if we continue to use gksu*/kdesu*):

```sh
if test -x "$(which pkexec 2>/dev/null)"; then
    WINETRICKS_SUDO=pkexec
elif test -x "$(which gksudo 2>/dev/null)"; then
    WINETRICKS_SUDO=gksudo
elif test -x "$(which gksu 2>/dev/null)"; then
    WINETRICKS_SUDO=gksu
elif test -x "$(which kdesudo 2>/dev/null)"; then
    WINETRICKS_SUDO=kdesudo
elif test -x "$(which kdesu 2>/dev/null)"; then
    WINETRICKS_SUDO=kdesu
fi

# ...

case "$WINETRICKS_SUDO" in
    gksu*|kdesudo)
      # $WINETRICKS_SUDO "cmdline"
      ;;
    kdesu)
      # $WINETRICKS_SUDO -c "cmdline"
      ;;
    *)
      # $WINETRICKS_SUDO cmd ...
      ;;
esac
```

I meant I only tested the command line arguments handling of GUI sudo programs and your `ubuntu` verb, and didn't/can't test the real verbs in `winetricks` script because (again,) I don't have any discs that can be used with Winetricks.

Anyway, why not use `unar`?

```sh
# - sudo is used to mount .iso images if the user cached them with -k option.
# FIXME: find a way to mount or copy from image without sudo
```

kakurasan:
I forgot `fuseiso`. This can mount .iso images as normal user. It's very old but still available on at least Debian/Ubuntu and Mageia.
I guess `unar` is only useful for fallback in most cases because of disk usage in temporary directory and waiting time for unpacking.

kakurasan:
In addition, `archivemount` can mount .iso images as normal user. This is available on at least Debian/Ubuntu and Fedora.

```
(mount)
$ archivemount -o readonly /path/to/image.iso /path/to/mount_point
(unmount)
$ fusermount -u /path/to/mount_point
```

austin987:
I'll check the branch tomorrow, busy week.

@kakurasan my main reason for not moving to new tools is that I'm not sure it will work in replacement. Some games will install via files copied to disc, but others won't because of DRM. Granted, not all will install from ISO, but I'd bet just mounting without a symlinked device would break something.

kakurasan:
@austin987 For now, I consider that we should add `pkexec` as default option and fix the remaining command line related issues. I updated #913 and would like you to review it. All "GUI sudo" programs (`pkexec`, `gksudo`, `kdesudo`, `gksu`, and `kdesu`) are already tested with [ubuntu.verb](https://github.com/Winetricks/winetricks/pull/915#issuecomment-362885013).

Also I already wrote an implementation that uses FUSE-based tools, but I also am not sure it works for all existing verbs. So I decided to add (not replace) the mount/unmount implementation in order to migrate from "`sudo` + `mount`" to FUSE-based tools in a phased manner.

austin987:
@kakurasan, thanks for the fix (and KDE testing!). I've merged #912.

I'm open to further improvements. You may want to look at #937, which involves mounting a .img file.

PR #926, opened 2018-01-27
Update Polish translation

m4sk1n:


austin987:
Thanks @m4sk1n; not sure what's up with Travis failing it, apparently they're having network issues.

PR #929, opened 2018-01-31
travis.yml: switch to Ubuntu Artful (17.10)

kakurasan:
Ubuntu Zesty (17.04) reached End of Life on January 13, 2018.
We need to switch to Artful (17.10).

See also: https://wiki.ubuntu.com/Releases

Note: Xenial (16.04 LTS, End of Life = April 2021) is available on Travis CI (experimental), but shellcheck is too old (0.3.7).


austin987:
Thanks.

PR #930, opened 2018-02-01
wmp9: update overrides

obsequiousnewt:
Wine now has l3codeca.acm, wmplayer.exe, and a version resource for wmvcore,
so we need to remove and override these as well.

wmp9 also tries and fails to install msdmo, which I've left alone for now.


austin987:
Thanks for the pr.

PR #940, opened 2018-02-19
Add MultisampleTextures option.

glintwine:


austin987:
@glintwine sorry, I meant disabled/enabled (with a d). Was my mistake; so I fixed it and pushed it:
https://github.com/Winetricks/winetricks/commit/185acea55d3b69361ea7eee8240e59dbeec9af28

thanks for the patch / sorry for the delay.

PR #945, opened 2018-03-01
Small fixes

kakurasan:
* cod1: fix typos (w_try_cd -> cd)
* w_download_to: exit from loop when download succeeded
    * Without this, `winetricks_dl_warning()` is called even if a file is successfully downloaded. This is meaningless.

### DIFF
--- a/src/winetricks
+++ b/src/winetricks
@@ -1178,6 +1178,9 @@ w_download_to()
 
             w_try_cd "$_W_dl_olddir"
             unset _W_dl_olddir
+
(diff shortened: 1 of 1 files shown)


src/winetricks
```diff
@@ -14094,7 +14097,7 @@ load_cod1()
             WinWait, Call of Duty, Start
             ControlClick Button1
         }
-        WinWait, Insert CD, Please insert the Call of Duty w_try_cd 2
```
qwertychouskie: LOL, find and replace for the win :D

PR #956, opened 2018-03-24
fix upstream issue Winetricks/winetricks/issues/955

apemberton:
Fixes winetricks/issues/955 - allows windowscodecs verb to finish without error


austin987:
Thanks; but I beat you to it. Include it with your bug next time ;)

PR #957, opened 2018-03-29
xact: Don't install 64-bit xact libraries

ColdPie1:
These are known to be broken in Wine, see Wine Bug 41618 comments 5 and
6. This also adds overrides for the 32-bit DLLs, since Wine now prefers
built-in.


austin987:
Please add a comment about why 64 isn't there (with link to the wine bug) so no one tries to reimplement it.

PR #963, opened 2018-04-03
travis.yml: install bashate with .deb files for Ubuntu 18.04 LTS

kakurasan:
The new script `travis-install-bashate.sh` installs bashate
by unpacking .deb files built for Ubuntu Bionic (18.04 LTS).
This also reduces job time.

Ubuntu Bionic will be supported until April 2023.

Also, ShellCheck is already available on Ubuntu image and
we don't need to install it.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,23 +16,43 @@ dist: trusty
 addons:
   apt:
     packages:
+    # for bashate
+    - python-tz
     # for checkbashisms
     - devscripts
 
 # Note: if statements cannot go across multiple lines. Combine into one, or move to a shell script!
 
-# On Ubuntu, shellcheck is not in trusty, but is in trusty-backports, but those aren't available in TravisCI..)
-# Using artful to get 0.4.6, which has support for disabling a check globally within a file
+# Ubuntu notes:
+#    * ShellCheck is already installed in TravisCI
+#    * bashate is not available on Trusty and we need to install it manually
+# macOS note:
+#    * Homebrew doesn't provide bashate, they don't package things
+#      that are on pypi
+#
+# We use a script that installs bashate by unpacking .deb files built for
+# Ubuntu Bionic (18.04) because:
+#   1. python-bashate 0.3.1-2 (for Artful) can be installed but
+#      Artful will not be supported after July 2018
+#   2. python-bashate 0.5.1-1 (for Bionic) can't be installed on Trusty
+#      because it contains "control.tar.xz"
+#      > dpkg-deb: error: archive '/var/cache/apt/archives/python-bashate_0.5.1-1_all.deb' has premature member 'control.tar.xz' before 'control.tar.gz', giving up
+#      > dpkg: error processing archive /var/cache/apt/archives/python-bashate_0.5.1-1_all.deb (--unpack):
+#      > subprocess dpkg-deb --control returned error exit status 2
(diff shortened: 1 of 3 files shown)


misc/travis-install-bashate.sh
```diff
@@ -0,0 +1,65 @@
...
+#     apt:
+#       packages:
+#       - python-tz
+#       - ...
+#   ...
+#   before_install:
+#       - time sh ./misc/travis-install-bashate.sh
+#       - ...
+
+set -x
```
austin987: You probably want (at least) set -e and set -u as well.

.travis.yml
```diff
@@ -16,23 +16,17 @@ dist: trusty
...
 # Note: if statements cannot go across multiple lines. Combine into one, or move to a shell script!
 
-# On Ubuntu, shellcheck is not in trusty, but is in trusty-backports, but those aren't available in TravisCI..)
-# Using artful to get 0.4.6, which has support for disabling a check globally within a file
+# On Ubuntu, shellcheck is already available in TravisCI
 before_install:
-    - if [[ "$TRAVIS_OS_NAME" == "linux" ]] ; then echo "deb http://archive.ubuntu.com/ubuntu/ artful universe" | sudo tee -a /etc/apt/sources.list ; fi
-    - if [[ "$TRAVIS_OS_NAME" == "linux" ]] ; then time sudo apt-get -q update ; fi
-    - if [[ "$TRAVIS_OS_NAME" == "linux" ]] ; then time sudo apt-get install python-bashate shellcheck ; fi
-
```
austin987: Please add a comment about why you're installing via this script.

austin987:
What's the time difference? It seems like this will have higher overhead, since we're hardcoding version names. Maybe it should install from repos as a fallback.

misc/travis-install-bashate.sh
kakurasan: `set -x` was replaced with `set -eux`, and `if ... fi` for `curl` was removed.

.travis.yml
kakurasan: Done.

kakurasan:
Logs in TravisCI say that `apt-get`, `easy_install`, and `pip` take longer time than this new script.

* Job 525.1 (Ubuntu):
    * `sudo apt-get -q update`: 7.15s
    * `sudo apt-get install`: 8.02s
    * bashate + shellcheck installation: 15.17s

* Job 525.2 (macOS):
    * `sudo easy_install pip`: 4.79s
    * `sudo pip install bashate`: 3.10s
    * bashate installation: 7.89s

* Job 528.1 (Ubuntu):
    * New script: 0.86s
* Job 528.2 (macOS):
    * New script: 2.14s

The total job time is not always reduced by this script.

I added a commit to skip Homebrew's `update` and this considerably reduces job time on macOS.

Also, I added a script that installs bashate via `pip`. It will be used when `travis-install-bashate-deb.sh` fails. The fallback script is also tested in TravisCI (passed).

austin987:
@kakurasan sorry, just getting back around to this. I put in a quick fix for OSX that now conflicts with your changes. Could you rebae please? Sorry about that.

kakurasan:
Done. `misc/travis-install-bashate-pip.sh` was changed to use the updated code (`curl` and `sudo python`).

PR #973, opened 2018-04-24
Add winetricks.bash-completion BASH completion script

bobwya:
```
winetricks list-all
winetricks --help
```

Requires BASH version 4.x - due to use of readarray.
Breaks MacOSX support?


bobwya:
Remove to refactor with Posix (**checkbashisms**) compliance...

PR #979, opened 2018-05-02
Add winetricks.bash-completion BASH Completion Script

bobwya:
```
winetricks list-all
winetricks --help
```
Regular expressions and verbs, that are explicitly referenced
in the script, are broken out to global (read-only) variables
at the head of the script (to ease future updates).
The script is designed to be Posix compliant. Passes all
shellcheck and checkbashism's current tests.

Add DISPLAY="" to winetricks_early_wine() function
The DISPLAY env variable was incorrectly commented as
not useable - in this context.
However this wine wrapper function is only used for early
wine cli / non-gui command calls. The env variable won't
persist to affect subsquent calls, to wine, from winetricks
verbs.

Makefile updated.
Installs the winetricks.bashcomp BASH completion script
in the standard Linux distribution directory
for completion helper scripts.

### DIFF
--- a/src/winetricks
+++ b/src/winetricks
@@ -3129,16 +3129,23 @@ winetricks_print_version()
 # Handy place to put small workarounds
 winetricks_early_wine()
 {
-    # The sed works around https://bugs.winehq.org/show_bug.cgi?id=25838
-    # which unfortunately got released in wine-1.3.12
-    # We would like to use DISPLAY= to prevent virtual desktops from
-    # popping up, but that causes AutoHotKey's tray icon to not show up.
-    # We used to use WINEDLLOVERRIDES=mshtml= here to suppress the Gecko
-    # autoinstall, but that yielded wineprefixes that *never* autoinstalled
-    # Gecko (winezeug bug 223).
-    # The tr removes carriage returns so expanded variables don't have crud on the end
-    # The grep works around using new wineprefixes with old wine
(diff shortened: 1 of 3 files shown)


src/winetricks
```diff
@@ -3131,14 +3131,15 @@ winetricks_early_wine()
...
+    # We can use DISPLAY= here (e.g. to prevent virtual desktops from
+    # popping up and WINEPREFIX update notifications) as this function
+    # is only called for command line wine instructions.
     # We used to use WINEDLLOVERRIDES=mshtml= here to suppress the Gecko
     # autoinstall, but that yielded wineprefixes that *never* autoinstalled
     # Gecko (winezeug bug 223).
     # The tr removes carriage returns so expanded variables don't have crud on the end
     # The grep works around using new wineprefixes with old wine
-    WINEDEBUG=-all "$WINE" "$@" 2> "$W_TMP_EARLY"/early_wine.err.txt | ( sed 's/.*1h.=//' | tr -d '\r' | grep -v -e "Module not found" -e "Could not load wine-gecko" || true)
+    DISPLAY="" WINEDEBUG=-all "$WINE" "$@" 2> "$W_TMP_EARLY"/early_wine.err.txt | ( sed 's/.*1h.=//' | tr -d '\r' | grep -v -e "Module not found" -e "Could not load wine-gecko" || true)
```
qwertychouskie: This might cause the initialization to forever pause at the "Install Gecko?" screen, as the yes/no button can't be clicked.

bobwya: Could point - I'm spoilt by Gentoo in that regard :-)
I'll repush with a hybrid solution - partially based on the **winetricks** sandbox verb (which I had originally planned to use).

bobwya:
Re-submitting with a fix for the DISPLAY="" Wineprefix update issue...

PR #980, opened 2018-05-03
Add winetricks.bash-completion BASH Completion Script

bobwya:
        ```
        winetricks list-all
        winetricks --help
        ```
Regular expressions and verbs, that are explicitly referenced
in the script, are broken out to global (read-only) variables
at the head of the script (to ease future updates).
The script is designed to be Posix compliant. Passes all
shellcheck and checkbashism, current tests.

Makefile updated to install the winetricks.bash-completion script
Uses the (currently) standard Linux distribution directory
for BASH completion helper scripts.

winetricks: disable wine gui for winetricks_early_wine() function
(... except when initially creating a Wineprefix!)

When working with an existing Wineprefix...
Temporarily disabled Wineprefix updates.
Use DISPLAY= (e.g. to prevent virtual desktops from
popping up and WINEPREFIX update notifications).

This provides support for the winetricks.bash-completion helper
script - by avoiding annoying popups when a user is simply trying to
do a winetricks cli completion.

Remove obsolete sed usage (only required for Wine versions 1.3.11-1.3.15).

### DIFF
--- a/src/winetricks
+++ b/src/winetricks
@@ -15760,7 +15760,7 @@ load_fifa11_demo()
 w_metadata hon games \
     title="Heroes of Newerth" \
     publisher="S2 Games" \
-    year="2013" \
+    year="2018" \
     media="download" \
     file1="HoNClient.exe" \
     installed_exe1="$W_PROGRAMS_X86_WIN/Heroes of Newerth/hon.exe"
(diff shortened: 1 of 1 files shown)


bobwya:
Austin,
Any chance of this getting merged?
I've cleaned up the Travis CI issues (as best I can)...
Thanks
Robert

austin987:
@bobwya my goal was to merge it this weekend, then I got distracted with fixing tests and some minor refactoring.

My new goal is to do so Tuesday/Wednesday.

PR #994, opened 2018-05-15
URL updates for pptfonts and gmdls

kakurasan:
* `pptfonts` (`calibri`, `cambria`, `candara`, `consolas`, `constantia`, `corbel`, and `meiryo`): Use alternative URL because PowerPoint Viewer is no longer available from Microsoft
* `gmdls`: Use DirectX April 2006 instead of DirectX 8 because it seems that DirectX 8 redist is available only in archive.org
    * The file (`gm.dls`) is the same as DirectX 8 (sha256:`3229b09b9d7d9f3f4793b0d9b34fe6abc75cfa4a2503c0c90f43ff651ba7f2c0`)

### DIFF
--- a/src/winetricks
+++ b/src/winetricks
@@ -10321,7 +10321,8 @@ w_metadata calibri fonts \
 load_calibri()
 {
     # https://www.microsoft.com/en-us/download/details.aspx?id=13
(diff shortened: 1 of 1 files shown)


austin987:
Thanks for the quick patch @kakurasan!

PR #1016, opened 2018-06-06
Bash completion

bobwya:
I'm trying to get more organised now... I've put these 3 BASH completion commits in separate **bash-completion** git branch... Still don't have a clue how to use git... But I'm getting there!

Also fixed an issue with array bounds checking (after noticing that that the **winxp** verb had disappeared from the completions list!)

### DIFF
--- a/src/winetricks.bash-completion
+++ b/src/winetricks.bash-completion
@@ -1,0 +1,819 @@
+#!/bin/bash
+#
+# BASH completion for the Winetricks package manager
+#
+#--------------------------------------------------------------------
+#
+# Copyright:
+#   Copyright (C) 2018 Rob Walker <bob.mt.wya!gmail.com>
+#
+# License:
+#   This program is free software; you can redistribute it and/or
+#   modify it under the terms of the GNU Lesser General Public
+#   License as published by the Free Software Foundation; either
+#   version 2.1 of the License, or (at your option) any later
+#   version.
+#
+#   This program is distributed in the hope that it will be useful,
+#   but WITHOUT ANY WARRANTY; without even the implied warranty of
+#   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#   GNU Lesser General Public License for more details.
+#
+#   You should have received a copy of the GNU Lesser General Public
+#   License along with this program.  If not, see
+#   <https://www.gnu.org/licenses/>.
+
+
+##### Define Global constants / variables #####
+WINETRICKS_PATH="$(command -v winetricks)"
+COUNTRY_CODES="AF AX AL DZ AS AD AO AI AQ AG AR AM AW AU AT AZ BS BH BD BB BY BE BZ BJ BM BT BA BW BV BR IO BN BG BF BI \
+KH CM CA CV KY CF TD CL CN CX CC CO KM CG CK CR CI HR CU CW CY CZ DK DJ DM DO EC EG SV GQ ER EE ET FK FO FJ FI FR \
+GF PF TF GA GM GE DE GH GI GR GL GD GP GU GT GG GN GW GY HT HM VA HN HK HU IS IN ID IQ IE IM IL IT JM JP JE JO KZ KE KI KW KG \
+LA LV LB LS LR LY LI LT LU MO MG MW MY MV ML MT MH MQ MR MU YT MX MC MN ME MS MA MZ MM NA NR NP NL NC NZ NI NE NG NU NF MP NO \
+OM PK PW PA PG PY PE PH PN PL PT PR QA RE RO RU RW BL KN LC MF PM VC WS SM ST SA SN RS SC SL SG SX SK SI SB SO ZA GS SS ES LK \
+SD SR SJ SZ SE CH SY TJ TH TL TG TK TO TT TN TR TM TC TV UG UA AE GB US UM UY UZ VU VN WF EH YE ZM ZW"
+COUNTRY_CODE_URL="https://pkgstore.datahub.io/core/country-list/data_csv/data/d7c9d7cfb42cb69f4422dec222dbbaa8/data_csv.csv"
+INVERTIBLE_OPTS="isolate opt"
+TERMINATING_OPTS="help update version"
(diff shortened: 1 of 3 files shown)


bobwya:
Austin,

Don't bother pulling or working on this Pull request _yet_, till I've updated push a significant re-factoring ...

I've got a working prototype which doesn't call **winetricks** at all. I went back to one of my original plans, which was to simply parse the **winetricks** script file itself... This will also avoid the _awkward_ _pause_ when you first try to tab complete **winetricks** in a shell session (which is really annoying!!)

This change actually required very little modification of the existing BASH completion script... Thankfully!

I've tried to tidy up the script quite a bit as well and pepper it with helpful comments...

Obviously I've got to run some more tests/checks to test the functionality is 100% there. It's easy to miss something, since **winetricks** has so many verbs!

One thing I did notice is that:

    winetricks *=builtin

etc. is impossible to parse... That will simply complete to the expression, i.e. with a **dll** name of * !! The user will then have to edit in the **dll** name manually... That can't be worked around easily, especially as the argument could be virtually any value...

**winetricks** could probably do with a test for trying to override bogus **dll** names like * ... Maybe an NTFS invalid character check would be a sufficient first step?

Rob

bobwya:
@austin987 

I've updated the bash completion script, so it now scrapes the raw **winetricks** script, without calling it.

Generally I've made some further improvements... Including hopefully making it more readable (with lots of comments)!

I'm pretty pleased with the performance now... There are no discernible startup delays or completion delays. Well, that I can detect anyway. Generally I've run out of obvious improvements to make to the script, hence why I'm throwing it over the wall!

On my Gentoo install BASH completion is ignored for BASH shells <**4.1** ... I've tried to keep the script BASH **3.x** compliant, but I've probably missed something, somewhere... :-)

austin987:
@bobwya, sounds good. I won't be able to test/review this until after wineconf, fyi. On my after I'm back todo list though ;)

bobwya:
@austin987 

I've rebased this again... But how are we doing, at getting it merged? ;-)

src/winetricks.bash-completion
```diff
@@ -0,0 +1,805 @@
...
+
+##### Define Global constants / variables #####
+WINETRICKS_PATH="$(command -v winetricks)"
+# shellcheck disable=SC2140
+COUNTRY_CODES="AF AX AL DZ AS AD AO AI AQ AG AR AM AW AU AT AZ BS BH BD BB BY BE BZ BJ BM BT BA BW BV BR IO BN BG BF BI"\
+" KH CM CA CV KY CF TD CL CN CX CC CO KM CG CK CR CI HR CU CW CY CZ DK DJ DM DO EC EG SV GQ ER EE ET FK FO FJ FI FR"\
+" GF PF TF GA GM GE DE GH GI GR GL GD GP GU GT GG GN GW GY HT HM VA HN HK HU IS IN ID IQ IE IM IL IT JM JP JE JO KZ KE KI KW KG"\
+" LA LV LB LS LR LY LI LT LU MO MG MW MY MV ML MT MH MQ MR MU YT MX MC MN ME MS MA MZ MM NA NR NP NL NC NZ NI NE NG NU NF MP NO"\
+" OM PK PW PA PG PY PE PH PN PL PT PR QA RE RO RU RW BL KN LC MF PM VC WS SM ST SA SN RS SC SL SG SX SK SI SB SO ZA GS SS ES LK"\
+" SD SR SJ SZ SE CH SY TJ TH TL TG TK TO TT TN TR TM TC TV UG UA AE GB US UM UY UZ VU VN WF EH YE ZM ZW"
```
austin987: Instead of disabling the check, this should be:
COUNTRY_CODE="A B C D \
E F G H \
I J K"

bobwya:
Hmmm, just let me know what issues are cropping up on the Arm system... I'll take a look at that.

austin987:
`winetricks TAB TAB` just shows files in the current directory, not any winetricks commands. I wasn't sure what would be useful for debugging that, maybe `set -x` in the completion file?

bobwya:
"Phase one" testing would be to apply this patch (or add the lines manually):
[winetricks-20180805_bash_completion_debug.patch.txt](https://github.com/Winetricks/winetricks/files/2259997/winetricks-20180805_bash_completion_debug.patch.txt)
to the **winetricks** bash-completion completion script.

Then just re-source the script in a shell session, with bash-completions already enabled.
This will produce a log file from just sourcing the completion script, without attempting any completions.
If you post this first I'll take a look.

austin987:
Sure, here you go.
[debug_output.txt](https://github.com/Winetricks/winetricks/files/2260059/debug_output.txt)

bobwya:
@austin987 

I think I've seen this problem before with Debian / Ubuntu (for my Wine build script)...

What's the output from:

```
awk -W version
```

on both Debian systems?

Ta
Bob

austin987:
good:
austin@debian-laptop:~$ awk -W version
GNU Awk 4.1.4, API: 1.1 (GNU MPFR 4.0.1, GNU MP 6.1.2)
Copyright (C) 1989, 1991-2016 Free Software Foundation.

bad:
mawk 1.3.3 Nov 1996, Copyright (C) Michael D. Brennan

compiled limits:
max NF             32767
sprintf buffer      1020

austin987:
If it's easy, that would be good. Along with a comment in the completion file (include a link to this PR) so we have something to point people to if they complain about it not working ;)

bobwya:
@austin987 

I've added an exhaustive search for all functioning **awk** variants (including the **busybox** builtin). If none is found then just exit gracefully. So I think your Arm system problem _should_ be solved. :-)

Also fixed that **shellcheck** multiline string issue.

Bob

austin987:
The debian stretch (arm) box still shows no completions, but I don't think it's worth blocking over. I'm not super worried about it, but if you want I can file an issue to track it.

Also, thanks for your patience on this. I appreciate you seeing it through, I'm sure a lot of people will like having working bash completion (if they're not on stretch ;)).

Fixed by https://github.com/Winetricks/winetricks/commit/81d3cf3bec93560fb29f110490a85db1020f4bd4

PR #1019, opened 2018-06-07
Fix handling the case where wine binaries have a suffix

bobwya:
E.g. Gentoo multi-slotted Wine variants could be of the form:

**wine-staging-9999**
**wine64-staging-9999**


austin987:
Thanks @bobwya 

This doesn't work if WINE isn't a full path (but is a binary in $PATH). E.g.,:
WINE=wine-vanilla-3.0
though WINE=$(which wine-vanilla-3.0) works now

bobwya:
Now correctly handles being passed just a filename path (in env var **WINE**). Whoops my bad!

bobwya:
@austin987 
Now really, really correctly handles being passed just a filename path (in env var WINE). Whoops my double bad! :-)

austin987:
Thanks for fixing that @bobwya

PR #1024, opened 2018-06-09
Add verb isolate_home

bobwya:
This verb just removes Wineprefix symlinks to the Wine user's
$HOME directory. Complementary to the sandbox verb.

### DIFF
--- a/src/winetricks
+++ b/src/winetricks
@@ -19375,6 +19375,34 @@ load_hosts()
 
 #----------------------------------------------------------------
 
+w_metadata isolate_home settings \
+    title_uk="Видалити посилання на вино преміум на \$HOME" \
+    title="Remove wineprefix links to \$HOME"
+
+load_isolate_home()
+{
+    w_skip_windows isolate_home && return
+
+    _olddir="$(pwd)"
+    w_try_cd "$WINEPREFIX/drive_c/users/$USER"
+    for x in *
+    do
+        if test -h "$x" && test -d "$x"; then
+            rm -f "$x"
+            mkdir -p "$x"
+        fi
+    done
+    w_try_cd "$_olddir"
+    unset _olddir
+
+    # Workaround for:
+    # https://bugs.winehq.org/show_bug.cgi?id=22450 (sandbox verb)
+    # https://bugs.winehq.org/show_bug.cgi?id=22974 (isolate_home, sandbox verbs)
+    echo disable > "$WINEPREFIX/.update-timestamp"
+}
+
+#----------------------------------------------------------------
+
 w_metadata native_mdac settings \
     title_uk="Перевизначити odbc32, odbccp32 та oledb32" \
     title="Override odbc32, odbccp32 and oledb32"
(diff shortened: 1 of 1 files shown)


src/winetricks
```diff
@@ -19261,25 +19292,7 @@ load_sandbox()
...
-    unset _olddir
-
-    # Disable unixfs
-    # Unfortunately, when you run with a different version of Wine, Wine will recreate this key.
-    # See https://bugs.winehq.org/show_bug.cgi?id=22450
-    "$WINE" regedit /d 'HKEY_LOCAL_MACHINE\\Software\\Microsoft\Windows\CurrentVersion\Explorer\Desktop\Namespace\{9D20AAE8-0625-44B0-9CA7-71889C2254D9}'
-
-    # Disable recreation of the above key - or any updating of the registry - when running with a newer version of Wine.
-    echo disable > "$WINEPREFIX/.update-timestamp"
+    load_isolate_home
```
austin987: This should be 'w_call isolate_home'

bobwya:
I've rebased my **sandbox_home** branch... It looks like this Pull request is hanging off your Master, in the Winetricks forks graph... So we're good now... Right??!! :-)

austin987:
Reading a bit closer, I'm not sure why you are moving unixfs disabling. Isolate home implies this only isolates home, not disabling Unix file system access.

bobwya:
Right, I've made that fix... Hows it looking now?

src/winetricks
```diff
@@ -19354,6 +19354,37 @@ load_hosts()
...
+            rm -f "$x"
+            mkdir -p "$x"
+        fi
+    done
+    w_try_cd "$_olddir"
+    unset _olddir
+
+    # Disable automatic Wineprefix updates.
+    # Disable recreation of the $HOME directory symlinks.
+    # In addition to suppport the sandbox verb...
```
austin987: Support*

src/winetricks
```diff
@@ -19354,6 +19354,37 @@ load_hosts()
...
+    # Disables recreation of the {9D20AAE8-0625-44B0-9CA7-71889C2254D9}
```
austin987: No longer valid.

src/winetricks
```diff
@@ -19354,6 +19354,37 @@ load_hosts()
...
+    # GUID key (aka unixfs disable) or any automatic updating of the
```
austin987: Same.

src/winetricks
```diff
@@ -19354,6 +19354,37 @@ load_hosts()
...
+    # Wine registry.
```
austin987: And this.

bobwya:
@austin987 

I was simply trying to comment the full side effects of the:
`echo disable > "$WINEPREFIX/.update-timestamp"`
call... The temptation might be to remove this line... If say "someone" was to be working on the Wineprefix update and re-creation of the $HOME symlinks issue (Upstream in Wine)... :-)

Maybe it's easier to keep the  **isolate_home** and **sandbox** verbs completely separate (i.e. not share any code)??!!
To avoid issues like this!

austin987:
@bobwya sharing the code is fine (and smart). The problem is that unixfs is only disabled in the sandbox() verb, but the comment in isolate_home() implies that it also happens there. Lines 19377-19378 are fine (and true), but 19379-19382 need to be removed because they're not true.

bobwya:
@austin987 

I've reworded the timestamp file comment in the **isolate_home** verb following your suggestion. I still think it's important to mention that this file modification is also required to disable dynamic registry re-creation, during a WIneprefix update. If https://bugs.winehq.org/show_bug.cgi?id=22974 is fixed Upstream (by say me for example)... Then the **sandbox** verb will still need Wineprefix updates disabled, to workaround the independent _unixfs_ Registry key recreation issue.

austin987:
@bobwya, I guess I was unclear. I agree with you that the registry update disabling should be documented ;). I just don't want isolate_home to say it disables unixfs when it doesn't.

bobwya:
@austin987 
That comment about disabling **unixfs** remained in the original **sandbox** verb, I haven't moved that. There is no reference to **unixfs** in the **isolate_home** verb... So I'm confused as to what you mean?

Disabling of the current Wineprefix timestamp is _very clearly_ commented now (in the **isolate_home** verb):
```
    # Workaround for:
    # https://bugs.winehq.org/show_bug.cgi?id=22450 (sandbox verb)
    # https://bugs.winehq.org/show_bug.cgi?id=22974 (isolate_home, sandbox verbs)
    echo disable > "$WINEPREFIX/.update-timestamp"
```
Seriously, I don't think I can make it any clearer than that! Less is more, etc.
You're really out to "draw blood" over this one? ;-)

austin987:
@bobwya, not at all! I was just replying to your previous comment. I'll clarify over a pint next week ;).

PR #1028, opened 2018-06-10
DXVK fixes

kakurasan:
This PR fixes the following:

* `w_override_dlls native d3d11.dll dxgi.dll` doesn't work because it contains `.dll`
* `_W_dxvk_dir` is not unset


austin987:
Nice catch, thanks

PR #1036, opened 2018-06-17
dx8vb: new verb

### LINKED ISSUE #1035: Request: dx8vb verb
Can we get a verb for dx8vb? That's an old, rarely used dll from XP times, to allow VB programs to use DirectX8.

The dll can be found (for example) in the [DirectX 8.1 Runtime](https://www.microsoft.com/de-de/download/details.aspx?id=10968)
I just used 7z to unpack the exe and got the dll. It needs to go into the 32bit system folder (system32/syswow64) and currently doesn't even need a dll override or regsvr32 - Though an override probably would be useful.

kakurasan:
Fixes #1035

@DarkShadow44 Could you test `https://raw.githubusercontent.com/kakurasan/winetricks/128bb4e63dd4152bd171bbe02d13c8ef142e3c39/src/winetricks`? Does the verb `dx8vb` work as you expect? Unfortunately I don't know any application that uses it.

### DIFF
--- a/src/winetricks
+++ b/src/winetricks
@@ -7770,6 +7770,26 @@ load_dotnet_verifier()
 
 #----------------------------------------------------------------
 
+w_metadata dx8vb dlls \
+    title="MS dx8vb.dll from DirectX 8.1 runtime" \
+    publisher="Microsoft" \
+    year="2001" \
+    media="download" \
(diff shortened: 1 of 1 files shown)


DarkShadow44:
Thank you, works like a charm!

PR #1039, opened 2018-06-22
Update dxvk verb to version 0.60 (released 22 June 2018)

bobwya:
Add warning about wine version (>=3.10) requirement.

### DIFF
--- a/src/winetricks
+++ b/src/winetricks
@@ -6605,23 +6605,28 @@ load_dxsdk_jun2010()
 
 #----------------------------------------------------------------
 
-w_metadata dxvk dlls \
-    title="Vulkan-based D3D11 implementation for Linux / Wine" \
-    publisher="Philip Rebohle" \
-    year="2018" \
-    media="download" \
-    file1="dxvk-0.54.tar.gz" \
-    installed_file1="$W_SYSTEM32_DLLS_WIN/d3d11.dll" \
-    installed_file2="$W_SYSTEM32_DLLS_WIN/dxgi.dll"
-
-load_dxvk()
+# $1 - dxvk archive name (required)
+# $2 - minimum Wine version (optional)
+# $3 - minimum Vulkan API version (optional, requires $2 be set)
+helper_dxvk()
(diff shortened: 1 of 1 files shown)


bobwya:
@austin987 

I did think about that... The graphics driver requirements are pretty extreme (for non-**Gentoo** / **Arch** users, hah, hah). ... Perhaps a generic **winetricks** warning to ensure that users check their graphics driver version is supported, with a link to [doitsujin / dxvk Wiki: Driver support](https://github.com/doitsujin/dxvk/wiki/Driver-support), would be sufficient?

austin987:
@bobwya, that sounds like a better plan to me!

austin987:
@bobwya I have no strong feelings about it. If it's supposed to be backwards compatible, we shouldn't need to have more than one, but it's not a perfect world..

Given that, I would:
* rename current verb to dxvk54, add warning about driver (optionally in a helper function, but I don't really care either way)
* add new dxvk60 based on what you have now
* add dxvk verb that just does w_call dxvk60

as new versions are added, we will add them and then change the dxvk to call the newest version. Sound good?

bobwya:
Not sure if I've followed the right layout for adding a helper function??!!
The **dxvk** verb seems to be quite extensible now...

austin987:
@bobwya, yep, it's mostly good (just some minor cleanups at this point).
A) I would ask that you add comments documenting the args.

E.g., at top, add:
# $1 - dxvk archive (required)
# $2 - minimum Wine version (optional)
# $3 - Vulkan API version (optional)
...

B) It would also be nice to assign them to variables instead of depending on arg position later, in case the args change later. So,
dxvk_archive="$1"
min_wine="$2"
vulkan_api="$3"

and replace $1/$2/$3 in the helper with those variables. Also, don't feel tied to those names, if you have better ideas ;).

austin987:
Also, I'm glad to see you're using force pushes now :)

bobwya:
Learning from the "master"... :-)

PR #1041, opened 2018-06-24
flash: update to 29.0.0.171

### LINKED ISSUE #1040: 'flash' winetrick not working properly; one of the URLs is now broken
More specifically, that one:
https://download.macromedia.com/pub/flashplayer/updaters/14/flashplayer_14_sa.exe

Accessing it now leads to a 'Not Found' page instead of the file. I didn't had the file with me, but I've looked for it in the internet until I found it in a torrent. It matches the hash in the winetricks file, so I guess it can be considered legitimate. 

I've hosted it in the Porting Kit website. This is the URL in case you want to use it, or at least if you want to download it to host somewhere else:
http://portingkit.com/wineskin/winetricks/flash/flashplayer_14_sa.exe

kakurasan:
Fixes #1040

### DIFF
--- a/src/winetricks
+++ b/src/winetricks
@@ -7939,20 +7939,26 @@ load_esent()
 # FIXME: update winetricks_is_installed to look at installed_file2..n
 # https://github.com/Winetricks/winetricks/issues/988
 w_metadata flash dlls \
-    title="Flash Player 14" \
+    title="Flash Player 29" \
     publisher="Adobe" \
-    year="2014" \
+    year="2018" \
     media="download" \
-    file1="install_flash_player.exe" \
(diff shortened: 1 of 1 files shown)


kakurasan:
ActiveX plugin, NPAPI plugin, and standalone player are already tested.
`w_set_winver winxp` is needed because NPAPI plugin (Firefox) doesn't work when OS version is Vista or newer.

@vitor251093 I would be happy if you would test `https://raw.githubusercontent.com/kakurasan/winetricks/d81d51e3aa242dfa8facdc0a1c0bd1cd158a3b23/src/winetricks`.

PR #1044, opened 2018-07-02
dotnet: Remove obsolete workarounds for dotnet < 40

zfigura:
Thanks to extensive help from Hans Leidekker.

These don't require previous versions of .NET to install or successfully
verify, and bug 25060 has been fixed.

Tested all both with and without -q, no failures.

### DIFF
--- a/src/winetricks
+++ b/src/winetricks
@@ -7459,8 +7459,7 @@ load_dotnet20()
     w_call remove_mono
     w_call fontfix
 
-    # Recipe from https://bugs.winehq.org/show_bug.cgi?id=10467#c57
-    # and https://bugs.winehq.org/show_bug.cgi?id=30845#c10
+    # Needed for https://bugs.winehq.org/show_bug.cgi?id=12401
     w_set_winver win2k
 
     w_try_cd "$W_CACHE"/"$W_PACKAGE"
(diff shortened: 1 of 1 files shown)


qwertychouskie:
Which Wine versions was this tested with?  AFAIK devel/stable still doesn't have a `.msu` installation implementation, which AFAIR was needed for the dotnet installers unless the old versions were called first.

zfigura:
3.11. Perhaps the new dism stub is enough?

zfigura:
I think MSU support is only necessary if in Vista mode (which is required for 4.5+). Otherwise the MSI packages work fine.

chambm:
Just FYI, from a compatibility perspective, 35sp1 should cover all previous versions, but 40 and up will not.

I can conceive that some app could test which .NET version < 4.0 was in use and change its code path accordingly. Then, if the app had a bug in one of the codepaths such that only 2.0 would work and 3.5 would be bugged, that could be a reason to keep the older verbs around. But it's a stretch.

zfigura:
Patch rebased, and also updated to remove workarounds for some other early versions.

austin987:
For reference: I tested this with wine-3.0 and wine-3.19 (32-bit prefixes only) after cleaning up dotnet in tests/winetricks-test; the tests pass all 4 cases on my laptop.

Fixed by https://github.com/Winetricks/winetricks/commit/431bb18719c4acd134983a198fae6c7e98c3b281

PR #1059, opened 2018-07-21
dxvk63: new verb

bobwya:
The DXVK Vulkan header version was raised to 1.1.80. However Philip says he isn't using any newer extensions - so that is not a strict requirement.

### DIFF
--- a/src/winetricks
+++ b/src/winetricks
@@ -6710,6 +6710,22 @@ load_dxvk62()
     helper_dxvk "$file1" "3.10" "1.0.76"
 }
 
+w_metadata dxvk63 dlls \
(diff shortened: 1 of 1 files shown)


austin987:
Man, he really like re-releasing stuff, doesn't he ;).

@bobwya could you add a comment about the header being 1.1.80 but it being unused at this time please? (Just above the helper_dxvk() call)

bobwya:
Well technically he likes releases more than re-releases...
But perhaps that's just being too pedantic... :-)

I've add in the brief exchange, on the **dxvk-dev Discord** channel, RE the Vulkan API versioning...

PR #1060, opened 2018-07-25
foobar2000: update to v1.4

kakurasan:
foobar2000 v1.4 final is now available.


kakurasan:
@austin987 I noticed that the commit (https://github.com/Winetricks/winetricks/commit/5ed413d3de65c589fbe38276f600c85028e9ab6c) disappeared. It seems that the commits after 58750aa were changed. What happened?

Issue #970, opened 2018-04-19
Winetricks 64-bit issue with windowscodecs


apemberton:
Ran into an interesting problem. Haven't quite root-caused it but thought to open an issue early in case it's a known issue.

If I run a simple winetricks command (any command, eg: `winetricks winxp`) before attempting to install WindowsCodecs (wic_x64_enu.exe) on a 64-bit prefix / 64-bit wine binary, the installer consistently fails.

If I use `winecfg` to set windows to winxp before installing the same `wic_x64_enu.exe`, things work fine. 

It's as though winetricks is causing some corruption or not properly setting things in a 64 bit prefix.

austin987:
What winetricks version?

Please include the output.

apemberton:
```
$ winetricks --version
20180217 - sha256sum: 59c0c463152738d673bbbb044f8c379cf878d4c6575ad7888d7cf063a9a16d3f
```

slyscorpion:
Try the following command 

`sudo winetricks -self--update`

and then see if you still have the same issue once you have the latest version of winetricks.

austin987:
@qwertychouskie https://github.com/Winetricks/winetricks/issues/795

austin987:
I'm not able to reproduce this:
Using winetricks 20180217-next - sha256sum: 8c32bb4e5c61ba435ad632cfa6a825662fddd7f69d02ed0a52c8f0497b38fd24 with wine-3.0 and WINEARCH=win64|

both winetricks winxp and winetricks win10 work for me (verified in winecfg)

apemberton:
@austin987 same - the os is indeed changed, but as stated in description, the issue is actually that downstream windowscodecs verb fails for unknown reason @ this point.

austin987:
@apemberton from your initial description it seemed that setting the windows version was the problem.

In any case, you may be hitting the wmphoto problem, can't say without seeing the output. If so, https://github.com/Winetricks/winetricks/commit/f4d7b0ac35ae6bd1e23f2d904b31c9739ce3fc88 would fix that.

apemberton:
Ha. No. Not that one. (That's the one I sent you a duplicate PR, remember?
:)

I'll investigate this one more soon when I can. My bigger fish to fry is
the dotnet461 problem with 64 bit.

zfigura:
[...]

PR #1068, opened 2018-08-08
winxp: set proper version numbers for 32/64-bit

austin987:
@qwertychouskie @apemberton @zfigura @bobwya 

This is a follow up to #970. With this, winecfg sees version as winxp in a 64-bit prefix (and windowscodecs installs as well, with winxp instead of win2k3).

I'd appreciate a second set of eyes and wider testing, if you've got something that was broken before.

### DIFF
--- a/src/winetricks
+++ b/src/winetricks
@@ -2445,10 +2445,22 @@ _EOF_
             csdversion_hex=dword:00000400
             ;;
         winxp)
-            csdversion="Service Pack 3"
-            currentbuildnumber="2600"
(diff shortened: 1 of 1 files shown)


zfigura:
I can't say there was anything else I was looking at that was broken here, but this does look like a better solution to #970 than what I proposed.

austin987:
Merged as https://github.com/Winetricks/winetricks/commit/37af7264c88fd126d84288fafb49bb7263523840

PR #1075, opened 2018-08-12
run tests/linkcheck.sh from any dir

### LINKED ISSUE #891: winetricks-test assumes that winetricks is located in the same folder
See [here](https://github.com/Winetricks/winetricks/blob/389d1ed6810cab5a54be2d68c2c6b4752a04f272/tests/winetricks-test#L73).

Since they are located in different folders, users won't be able to simply run `winetricks-test` without some manual steps, like copying winetricks around.

austin987:
@mihaitodor 

This (should) fix #891, for both tests/winetricks-test and src/linkcheck.sh

MIhai, could you confirm this does what you expect?

### DIFF
--- a/src/linkcheck.sh
+++ b/src/linkcheck.sh
@@ -1,7 +1,8 @@
 #!/bin/sh
 # Link checker for winetricks.
 #
-# Copyright (C) 2011,2012,2013 Dan Kegel.
+# Copyright (C) 2011-2013 Dan Kegel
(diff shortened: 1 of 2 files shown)


austin987:
https://github.com/Winetricks/winetricks/commit/cf53698820c44de50c91ab446cac4f7e8ecabb89
https://github.com/Winetricks/winetricks/commit/62426e87c60c30a97c072dc0d1e97d48c0fbe509

PR #1078, opened 2018-08-14
Add dotnet472 verb

chambm:
@austin987 Force push apparently closed my other PR.

For dotnet472, the repacked installer works but not the official one. With Wine 3.13 the repacked installer is no longer needed for dotnet462 but I left it in anyway. I didn't see a reason to add 4.7 and 4.7.1 but correct me if I'm wrong.

I also fixed the installed file check for dotnet461 because it is incredibly fragile to rely on something staying in TEMP. But maybe that presents a backward compatibility problem? Any way to check for one of a set of files to maintain backward compatibility?

I added /sfxlang:1027 to force English during install, gleaned from another issue.

### DIFF
--- a/src/winetricks
+++ b/src/winetricks
@@ -884,6 +884,7 @@ w_dotnet_verify()
         dotnet46) version="4.6" ;;
         dotnet461) version="4.6.1" ;;
         dotnet462) version="4.6.2" ;;
+        dotnet472) version="4.7.2" ;;
         *) echo error ; exit 1 ;;
     esac
             w_call dotnet_verifier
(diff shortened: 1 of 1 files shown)


rik-shaw:
@chambm thanks a LOT for this work on dotnet472.  I am trying to test it out but am having some trouble with the application I am trying crashing after running for some seconds, with the beginning of the stack trace referencing this issue:

```
System.AccessViolationException: Attempted to read or write protected memory. This is often an indication that other memory is corrupt.
```

I will open a new bug report but it is hard to know if I should submit under winetricks dotnet472 or rather whether this type of bug is for wine proper.

rik-shaw:
@chambm thanks a LOT for this work on dotnet472.  I am trying to test it out but am having some trouble with the application I am trying crashing after running for some seconds, with the beginning of the stack trace referencing this issue:

```
System.AccessViolationException: Attempted to read or write protected memory. This is often an indication that other memory is corrupt.
```

I will open a new bug report but it is hard to know if I should submit under winetricks dotnet472 or rather whether this type of bug is for wine proper.

chambm:
What application are you testing? Maybe try a simple console application?

rik-shaw:
@chambm thanks for the reply.  I have a pretty high testing bar as I am working on behalf of many colleagues to see if Logos 7 (https://www.logos.com/install) can work with WINE.  The download there is a bootstrap installer, so I had to install in Windows (Virtualbox) in order to obtain the full installer.  But clearly there is a challenge with those "protected memory" errors.  It does seem that the .NET 4.7.2 install did complete.

Issue #988, opened 2018-05-08
# FIXME: update winetricks_is_installed to look at installed_file2..n


austin987:
Mentioned above flash_metadata.

We should loop over a reasonable number of installed files instead of only checking 1. I'd say 25 is probably a safe limit, since a file existence check is pretty cheap and I doubt we'll ever need more than 5/10.

h1z1:
Can  you elaborate a bit on what you mean here?  Batch downloading or?

austin987:
@h1z1 no. winetricks_is_installed() checks if the file passed to it is present in the WINEPREFIX. This is used to check if a verb has already been installed and to check for success after installing.

This issue is about expanding that function to allow checking for more than one file in the prefix (as flash tries to do, but it's currently a no-op).

PR #1081, opened 2018-08-17
dxvk helper/verbs: add support for d3d10

bobwya:
add verb: dxvk70

### DIFF
--- a/src/winetricks
+++ b/src/winetricks
@@ -6795,6 +6795,25 @@ load_dxvk65()
     helper_dxvk "$file1" "d3d10_disabled" "3.10" "1.0.76"
 }
 
+w_metadata dxvk70 dlls \
+    title="Vulkan-based D3D10/D3D11 implementation for Linux / Wine (0.70)" \
+    publisher="Philip Rebohle" \
+    year="2018" \
+    media="download" \
+    file1="dxvk-0.70.tar.gz" \
+    installed_file1="$W_SYSTEM32_DLLS_WIN/d3d10.dll" \
+    installed_file2="$W_SYSTEM32_DLLS_WIN/d3d10_1.dll" \
+    installed_file3="$W_SYSTEM32_DLLS_WIN/d3d10core.dll" \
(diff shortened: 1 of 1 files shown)


src/winetricks
```diff
@@ -6792,6 +6792,25 @@ load_dxvk65()
...
+    title="Vulkan-based D3D10/D3D11 implementation for Linux / Wine (0.70)" \
+    publisher="Philip Rebohle" \
+    year="2018" \
+    media="download" \
+    file1="dxvk-0.70.tar.gz" \
+    installed_file1="$W_SYSTEM32_DLLS_WIN/d3d10.dll" \
+    installed_file1="$W_SYSTEM32_DLLS_WIN/d3d10_1.dll" \
+    installed_file1="$W_SYSTEM32_DLLS_WIN/d3d10core.dll" \
+    installed_file1="$W_SYSTEM32_DLLS_WIN/d3d11.dll" \
+    installed_file2="$W_SYSTEM32_DLLS_WIN/dxgi.dll"
```
austin987: Redefining the same variable 4 times doesn't make much sense.

Note that currently only installed_file1 is checked. See https://github.com/Winetricks/winetricks/issues/988

src/winetricks
```diff
@@ -6632,13 +6632,18 @@ load_dxsdk_jun2010()
 #----------------------------------------------------------------
 
 # $1 - dxvk archive name (required)
-# $2 - minimum Wine version (optional)
-# $3 - minimum Vulkan API version (optional, requires $2 be set)
+# $2 - dxvk d3d10 support 0=no / 1=yes (required)
```
austin987: Since this value is only used by the helper (and not, e.g., set in the registry), I'd prefer something more descriptive. E.g., d3d10_enabled / d3d10_disabled (doesn't have to be exactly that, if you've got better ideas).

austin987:
Thanks.

Exactly, I want your verb to be ready for when it does work. There's a solid chance that bored/Ballmer Peak Austin will fix this at some point in the near future ;).

PR #1085, opened 2018-08-19
Native font names replacements related improvements

### LINKED ISSUE #781: Chinese cannot be displayed correctly in the input field when using QQ
The Chinese characters are displyed as blocks when I am using QQ and inputing Chinese. They cannot be displayed correctly. Howerer, when I send the messages(blocks) I've just input, they changed into normal Chinese characters in the chat interface.

Is this because of the absence of some kind of Chinese fonts? How could I solve this problem?

kakurasan:
Fixes #781 and potential problems in `fakejapanese*` verbs.

* ~~Set proper locale when running `regedit` (`w_register_font_replacement`)~~
* ~~Change existing variable names in `fakejapanese*`~~
* ~~Unset local variables in `fakejapanese*`~~
* Support native (non-English) font names in `w_register_font_replacement()` and pass native font names in `fakejapanese*`
* Add native font name replacements in `fakejapanese`
* Add Chinese and Korean native font name replacements

### DIFF
--- a/src/winetricks
+++ b/src/winetricks
@@ -11746,6 +11746,16 @@ load_fakechinese()
     w_register_font_replacement "NSimSun" "WenQuanYi Micro Hei"
     w_register_font_replacement "SimKai" "WenQuanYi Micro Hei"
     w_register_font_replacement "SimSun" "WenQuanYi Micro Hei"
+    w_register_font_replacement "宋体" "WenQuanYi Micro Hei"
+    w_register_font_replacement "新宋体" "WenQuanYi Micro Hei"
+    w_register_font_replacement "黑体" "WenQuanYi Micro Hei"
+    w_register_font_replacement "微软雅黑" "WenQuanYi Micro Hei"
+    w_register_font_replacement "仿宋" "WenQuanYi Micro Hei"
+    w_register_font_replacement "楷体" "WenQuanYi Micro Hei"
+    w_register_font_replacement "細明體" "WenQuanYi Micro Hei"
+    w_register_font_replacement "新細明體" "WenQuanYi Micro Hei"
+    w_register_font_replacement "微軟正黑體" "WenQuanYi Micro Hei"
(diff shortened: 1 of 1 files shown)


bsjeon:
[add_register_font_replacement_unicode.txt](https://github.com/Winetricks/winetricks/files/2702242/add_register_font_replacement_unicode.txt)

The Unicode registry file makes the task simpler.

kakurasan:
Good idea, but unfortunately it doesn't work with `dash` (`bash` works). Could you fix that (and supersede this PR) ?

bsjeon:
Replace "\xff\xfe" with "\377\376".

I hope this will be your work.

kakurasan:
I modified `w_register_font_replacement()` instead of adding `w_register_font_replacement_unicode()` because it's quite similar to `w_register_font_replacement()`.

Now the implementation is locale-independent and `w_register_font_replacement()` supports native (non-English) font names.

src/winetricks
```diff
@@ -2951,13 +2951,16 @@ _EOF_
...
 }
 
+# Note: we use UTF-16 (little endian) in .reg file for native (non-English) font names.
 w_register_font_replacement()
 {
     _W_alias=$1
     shift
     _W_font=$1
+    # UTF-16 BOM (U+FEFF, "0xEF 0xBB 0xBF" in UTF-8)
+    printf "\357\273\277" | iconv -f utf8 -t utf16le > "$W_TMP"/_register-font-replacements.reg
```
austin987: It looks like BSD doesn't support utf8/utf16le; it uses UTF-8/UTF-16LE (not sure what mac does, but that should be verified as well).

kakurasan: Fixed. This works on macOS:
iconv_test_macos

austin987:
I merged 3/4 commits, fakechinese now has a conflict that can be addressed as its own PR.

Sorry this took so long, I wasn't clear on how to test it before.

PR #1110, opened 2018-10-08
Improve graphical wineprefix management and bugfixes

mmmaisel:
This pull request allows to graphically manage wineprefixes and graphically install/run custom applications which should be very helpful for new users who don't like to use the terminal.

This is realized by:
* Adding "Create new wineprefix" menu item in prefix-menu which shows a new dialog for naming the wineprefix and select its architecture (32|64 bit).
* Adding "run explorer" menu item to main-menu so that you can install/run any application graphically by double-click within the selected wineprefix.
* Adding the new verb "arch=32" which creates a new 32bit wineprefix on 64bit systems.

Bugfixes:
* Wineprefixes from "~/.local/share/wineprefixes" are missing in prefix-menu when using kdialog.
* "Show/Hide broken apps" and "Enable/diable silent install" is missing in prefix-menu when using kdialog.
* Fixed a typo in "de" translation.
* Fixed missing wineprefix path in "da" translation

### DIFF
--- a/src/winetricks
+++ b/src/winetricks
@@ -3328,6 +3328,7 @@ winetricks_prefixmenu()
              _W_msg_games='Установить игру'
              _W_msg_benchmarks='Установить приложение для оценки производительности'
              _W_msg_default="Выберите путь для wine по умолчанию"
+             _W_msg_mkprefix="Создать новый путь wine"
              _W_msg_unattended0="Отключить автоматическую установку"
              _W_msg_unattended1="Включить автоматическую установку"
              _W_msg_showbroken0="Спрятать нерабочие программы (например, использующие DRM)"
@@ -3340,6 +3341,7 @@ winetricks_prefixmenu()
              _W_msg_games='Встановити гру'
              _W_msg_benchmarks='Встановити benchmark'
              _W_msg_default="Вибрати wineprefix за замовчуванням"
+             _W_msg_mkprefix="створити новий wineprefix"
              _W_msg_unattended0="Вимкнути автоматичне встановлення"
              _W_msg_unattended1="Увімкнути автоматичне встановлення"
              _W_msg_showbroken0="Сховати нестабільні додатки (наприклад з проблемами з DRM)"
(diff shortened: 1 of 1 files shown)


bobwya:
I would recommend using the `trans` command line tool, for filling in your translation placeholders, e.g.:

```
$ trans -s en -t ru 'Create new wineprefix'

Create new wineprefix

Создать новый файл wineprefix
(Sozdat' novyy fayl wineprefix)

Translations of Create new wineprefix
[ English -> Русский ]

Create new wineprefix
    Создать новый файл wineprefix, Создать новую wineprefix
```

It usually spits out something pretty close, to a good translation, in my experience!

mmmaisel:
I replaced the translation placeholders with Google translate results.
Hopefully they are understandable and, especially in case of Chinese, not completely wrong.

austin987:
Mostly looks good, thanks. Just concerned about the 64bit prefix thing.

mmmaisel:
I'm not sure if I understood your concerns correctly.
As far as I know, wine fails if WINEARCH is set to anything else than "win32" or "win64" and 32bit wine will create 32bit wineprefixes anyway. Pure 64bit wine fails if WINEARCH is to "win32" and pure 32bit wine fails if WINEARCH is "win64". wow64 wine creates wineprefixes with the same bit count as your system by default.

Do you mean running winetricks with "env WINEARCH=win32" set on 64 bit systems which may create 32bit wineprefixes independent of the user choise?
Usually, this will fail because the default wineprefix ~/.wine is already created as 64bit wineprefix.

If you set the default wineprefix to a 32bit one and then run "env WINEARCH=win32 winetricks",
creating a new 64 bit wineprefix worked without problems in during my tests because WINEARCH is unset at the end of winetricks_set_wineprefix() which is called from winetricks_init().

Nevertheless I added "export WINEARCH=win64" to winetricks_set_winearch() so that it will fail on 32 bit systems and any other inconsistencies.

mmmaisel:
@austin987 Is there any news on this?

austin987:
@mmmaisel sorry, I've been sick past few weeks. Looks okay at a glance, I just want to test it myself a bit. Hopefully this weekend

mmmaisel:
OK, I squashed the commits and added support for `arch=win32|win64`, as well as an error message for invalid values.

The `arch` verb does only work for non-standard wineprefixes because the default wineprefix is already created within `winetricks_init` if it does not exist, before any verb is processed (and thus it uses the existing environment). Fixing this requires a major rewrite of winetricks' codebase. I added a hint in the documentation for this behaviour.

I order to test the `arch` verb, run: `WINEARCH=win64 ./winetricks arch=32 prefix=test32`.
This should create a 32bit wineprefix at `~/.local/share/wineprefixes/test32`, overriding the environment settings.

austin987:
Thanks for fixing.

You're right about the CLI of course; I hadn't looked at winetricks_init() in a while, thanks for clarifying that.

PR #1113, opened 2018-10-09
secur32: remove verb

gverm:
Overriding secur32 leads to an immediate stack overflow, see [this bug report](https://bugs.winehq.org/show_bug.cgi?id=45344).

Quote from that bug report:
> Consider secur32.dll of the same level as kernel32, user32 and ntdll which can't be replaced.

IMO this verb either needs to be removed or a warning needs to be added to make people aware of this fact. (even though users are supposed to keep the pieces when a native dll breaks something)

If you are more in favor of a warning, I can do that too.


austin987:
Yeah, make it a warning. I'm not opposed to requiring --force, but that's up to you.

qwertychouskie:
Well, if it is guaranteed to cause a crash, then why have it at all?

gverm:
@austin987 I have changed it into a warning now.

gverm:
@austin987 Anything blocking this one?

PR #1130, opened 2018-11-09
wininet: update to newer version and add 64bit compatibility

gverm:
Might help to workaround [Bug 45847](https://bugs.winehq.org/show_bug.cgi?id=45847). According to Focht, it needs at least wininet.dll from Windows 7.
I had to add an extra override for iertutil.dll because testing QQIntl revealed that this version of wininet.dll requires it.
It would be nice if someone could test QQIntl completely. I couldn't go further than the login screen, because registration requires a phone number.
I also tried ie7 and ie8, which also use native wininet, but those are currently blocked by #911.

### DIFF
--- a/src/winetricks
+++ b/src/winetricks
@@ -10854,15 +10854,44 @@ load_winhttp()
 w_metadata wininet dlls \
     title="MS Windows Internet API" \
     publisher="Microsoft" \
+    year="2011" \
+    media="download" \
+    file1="../win7sp1/windows6.1-KB976932-X86.exe" \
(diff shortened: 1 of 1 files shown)


austin987:
I think we may want to keep both, and rename one with _winver suffix, so we don't break anything that doesn't like the win7 version. See, e.g., gdiplus. I'm with making this the default going forward, and renaming the old one to wininet_win2k().

gverm:
@austin987 Thanks for the suggestion. I've made the changes.

PR #1137, opened 2018-11-18
ie8: Override ieproxy to native and copy it to system32.

zfigura:
Fixes https://bugs.winehq.org/show_bug.cgi?id=44005.

### DIFF
--- a/src/winetricks
+++ b/src/winetricks
@@ -12824,7 +12824,7 @@ load_ie8()
     w_call msls31
 
     # Change the override to the native so we are sure we use and register them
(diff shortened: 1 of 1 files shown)


austin987:
Nice, thanks.

I tried the same fix for ie7. There, IE opens and stays open, but fails to load its page set to first time settings. You can then load winehq.org. On second run, though, the same problem returns. I suspect it needs more native dlls.

gverm:
@austin987 Not sure if it's related, but have you checked that my recent wininet upgrade has any influence?

PR #1140, opened 2018-11-21
README: improve Installation and Updating guides

bobwya:
Responding to user feedback... See: [Installation using wine gets stuck at .NET Framework 4.6.2](https://forum.winehq.org/viewtopic.php?f=8&t=31452).
Linux user's don't like to copy-paste random shell scripts from the Internet! An Ubuntu user will not think to remove a previously version of winetricks - so the new / manually installed and up-to-date winetricks version will be ignored.

Updates to README:

* Breakdown the winetricks update script in excruciating detail. With helpful links to the TLDP website.
* Split off an alternative update script, to show how to set it up, without using sudo. Keeps the non-sudo users happy.
* Don't use /usr/local/bin as an installation location. Even if this is in the user's PATH, it will get lower priority than /usr/bin. So a previously installed winetricks version (e.g. package manager version from 2 years ago) will take precedence.
* Reduce the complexity / usage of root privileges, for all scripts. Only escalate to root as required.
* Recommend to uninstall any distribution (package manager) version of winetricks first (with a stern warning!) Otherwise a manually installed version of winetricks will conflict with the packaged version. Single out Debian / Ubuntu - as the primary culprits, for offering totally obsolete packaged versions of winetricks - only offering de-installation instructions for apt.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,34 +28,108 @@ The ```winetricks``` package should be used if it is available and up to date. T
 
 Note: packaged Debian / Ubuntu winetricks versions are typically outdated, so a manual installation is recommended.
 
-If the package is unavailable, outdated, or the latest version is desired, a manual installation of winetricks can be done. E.g.:
+If the package is unavailable, outdated, or the latest version is desired, a manual installation of winetricks can be done.
+It is _highly_ recommended to uninstall any previously installed version of winetricks first.
+
+**_If you don't uninstall a previously installed, packaged version of winetricks... Well then you get to pick up the pieces!_**
+
+E.g. for Debian / Ubuntu:
+```
+sudo apt-get purge winetricks
+```
+
+Then, for Ubuntu, use a shell script to download the current winetricks script(s).
+E.g.:
 
 ```
-sudo -- sh -c '
-cat > /usr/local/bin/update_winetricks.sh <<_EOF_SCRIPT
+# Create and switch to a temporary directory writeable by current user. See:
+#   https://www.tldp.org/LDP/abs/html/subshells.html
+cd "$(mktemp -d)"
+
+# Use a BASH "here document" to create an updater shell script file.
+# See:
+#   https://www.tldp.org/LDP/abs/html/here-docs.html
+# >  outputs stdout to a file, overwriting any pre-existing file
+# << takes input, directly from the script itself, till the second '_EOF_SCRIPT' marker, as stdin
+# the cat command hooks these 2 streams up (stdin and stdout)
+###### create update_winetricks START ########
+cat > update_winetricks <<_EOF_SCRIPT
 #!/bin/sh
 
-cd "${HOME}" # change to a path writeable by current user
-wget https://raw.githubusercontent.com/Winetricks/winetricks/master/src/winetricks
-chmod +x winetricks
-sudo mv winetricks /usr/local/bin
-wget https://raw.githubusercontent.com/Winetricks/winetricks/master/src/winetricks.bash-completion
-sudo mv winetricks.bash-completion /usr/share/bash-completion/completions/winetricks
-_EOF_SCRIPT
-chmod +x /usr/local/bin/update_winetricks.sh
-'
+# Create and switch to a temporary directory writeable by current user. See:
+#   https://www.tldp.org/LDP/abs/html/subshells.html
+cd "$(mktemp -d)"
 
+# Download the latest winetricks script (master="latest version") from Github.
+wget https://raw.githubusercontent.com/Winetricks/winetricks/master/src/winetricks
+
+# Mark the winetricks script (we've just downloaded) as executable. See:
+#   https://www.tldp.org/LDP/GNU-Linux-Tools-Summary/html/x9543.htm
+chmod +x winetricks
+
+# Move the winetricks script to a location which will be in the standard user PATH. See:
+#   https://www.tldp.org/LDP/abs/html/internalvariables.html
+sudo mv winetricks /usr/bin
+
+# Download the latest winetricks BASH completion script (master="latest version") from Github.
+wget https://raw.githubusercontent.com/Winetricks/winetricks/master/src/winetricks.bash-completion
+
+# Move the winetricks BASH completion script to a standard location for BASH completion modules. See:
+#   https://www.tldp.org/LDP/abs/html/tabexpansion.html
+sudo mv winetricks.bash-completion /usr/share/bash-completion/completions/winetricks
+_EOF_SCRIPT 
+###### create update_winetricks FINISH ########
+
+# Mark the update_winetricks script (we've just written out) as executable. See:
+#   https://www.tldp.org/LDP/GNU-Linux-Tools-Summary/html/x9543.htm
+chmod +x update_winetricks
+
+# We must escalate privileges to root, as regular Linux users do not have write access to '/usr/bin'.
+sudo mv update_winetricks /usr/bin/
+```
+
+See the manpages for the individual functions, if you are not clear how they are being used, e.g.
+```
+man mktemp
+man mv
+man wget
+man sudo
+...
+```
+
+An alternative updater script implementation, using **su** in place of **sudo**, is also possible:
+
+```
+cd "$(mktemp -d)"
+cat > update_winetricks <<_EOF_SCRIPT
+#!/bin/sh
+
+cd "$(mktemp -d)"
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -28,34 +28,105 @@ The ```winetricks``` package should be used if it is available and up to date. T
...
-cat > /usr/local/bin/update_winetricks.sh <<_EOF_SCRIPT
+
+Then, for Ubuntu, use a simple shell script to download the current winetricks scripts.
+This script uses a single parameter (to set a working directory).
+E.g.:
+
+```
+# Change to a directory writeable by current user (their HOME directory). See:
+#   https://www.tldp.org/LDP/abs/html/internalvariables.html
+cd "${HOME}"
```
austin987: If you want to cd to $HOME, you can just use 'cd'

README.md
```diff
@@ -28,34 +28,105 @@ The ```winetricks``` package should be used if it is available and up to date. T
...
+###### create update_winetricks.sh START ########
+cat > update_winetricks.sh <<_EOF_SCRIPT
 #!/bin/sh
 
-cd "${HOME}" # change to a path writeable by current user
+# Change to a directory writeable by current user. See:
+#   https://www.tldp.org/LDP/abs/html/internalvariables.html
+# Use a script parameter $1 or fallback to the user's HOME directory.
+# This works around the lack of env variable support in systemd units (see below).
+cd "\${1:-\${HOME}}"
```
austin987: Why not just use a tempdir? E.g., `cd $(mktemp -d)`. The user won't need to interact with anything in there anyway, since it's an automated script.

README.md
```diff
@@ -28,34 +28,105 @@ The ```winetricks``` package should be used if it is available and up to date. T
...
-sudo mv winetricks /usr/local/bin
+
+# Move the winetricks script to a location which will be in the standard user PATH. See:
+#   https://www.tldp.org/LDP/abs/html/internalvariables.html
+sudo mv winetricks /usr/bin
+
+# Download the latest winetricks BASH completion script (master="latest version") from Github.
 wget https://raw.githubusercontent.com/Winetricks/winetricks/master/src/winetricks.bash-completion
+
+# Move the winetricks BASH completion script to a location standard location for BASH completion modules. See:
```
austin987: 'location standard location'?

README.md
```diff
@@ -28,34 +28,105 @@ The ```winetricks``` package should be used if it is available and up to date. T
...
+#   https://www.tldp.org/LDP/abs/html/tabexpansion.html
 sudo mv winetricks.bash-completion /usr/share/bash-completion/completions/winetricks
-_EOF_SCRIPT
-chmod +x /usr/local/bin/update_winetricks.sh
+_EOF_SCRIPT 
+###### create update_winetricks.sh FINISH ########
+
+# Mark the update_winetricks script (we've just written out) as execuable. See:
```
austin987: Executable*

README.md
```diff
@@ -28,34 +28,105 @@ The ```winetricks``` package should be used if it is available and up to date. T
...
+#   https://www.tldp.org/LDP/GNU-Linux-Tools-Summary/html/x9543.htm
+chmod +x update_winetricks.sh
+
+# We must escalate privileges, to root, as regular Ubuntu users do not normally have write access to '/usr/bin'.
```
austin987: You can drop Ubuntu, no distro that I know of gives non-root access to /usr

README.md
```diff
@@ -28,34 +28,105 @@ The ```winetricks``` package should be used if it is available and up to date. T
...
+cd "${HOME}"
+
+# Use a BASH "here document" to create an updater shell script file.
+# See:
+#   https://www.tldp.org/LDP/abs/html/here-docs.html
+# >  outputs stdout to a file, overwriting any pre-existing file
+# << takes input, directly from the script itself, till the second '_EOF_SCRIPT' marker, as stdin
+# the cat command hooks these 2 streams up (stdin and stdout)
+###### create update_winetricks.sh START ########
+cat > update_winetricks.sh <<_EOF_SCRIPT
```
austin987: The defacto standard is not to have .sh / .bash extensions on things in $PATH

README.md
```diff
@@ -28,34 +28,105 @@ The ```winetricks``` package should be used if it is available and up to date. T
...
+```
 
+See the manpages for the individual functions, if you are not clear how they are being used, e.g.
+```
+man wget
+man sudo
+...
+```
+
+For Linux distributions that do not have **sudo** available by default, then use this shell script variant to download the current winetricks scripts:
```
austin987: What distros are those? Or are you just being safe?

README.md
```diff
@@ -86,7 +162,7 @@ sudo systemctl start winetricks_update.timer
 ```
 Or using the more tradional Unix crontab...
 ```
-sudo cp "/usr/local/bin/update_winetricks.sh" "/etc/cron.weekly/"
+sudo ln "/usr/bin/update_winetricks.sh" "/etc/cron.weekly/update_winetricks.sh"
 ```
```
austin987: Suggestion: considering how much simpler this method is, I would list it first.

austin987:
What distro are you seeing that puts /usr/local/bin *after* /usr/bin? I've never seen that. The whole point of /usr/local is to allow an administrator to override the distro version, which is useless if it's second place.

README.md
bobwya: I was trying to make the cd destination directory explicit - for "new users".

README.md
bobwya: That was a joke about one of the recent winetricks pull requests... :-)
I'll re-phrase that.

README.md
bobwya: Could call... I'll put a little warning for **systemd** lovers there...

bobwya:
@austin987 
Righteo v2 for review, with the suggested clean-ups.

bobwya:
@austin987
Righteo v3 for review, with the suggested clean-ups.
Removed **systemd** service unit reference to **%h** (which had sneakly stayed in place).

README.md
```diff
@@ -28,34 +28,110 @@ The ```winetricks``` package should be used if it is available and up to date. T
...
+_EOF_SCRIPT 
+###### create update_winetricks FINISH ########
+
+# Mark the update_winetricks script (we've just written out) as executable. See:
+#   https://www.tldp.org/LDP/GNU-Linux-Tools-Summary/html/x9543.htm
+chmod +x update_winetricks
+
+# We must escalate privileges to root, as regular Linux users do not have write access to '/usr/bin'.
+sudo mv update_winetricks /usr/bin/
 '
```
austin987: You should remove the apostrophe here, it's dangling on its own since you removed the old sh -c bits.

README.md
```diff
@@ -28,34 +28,110 @@ The ```winetricks``` package should be used if it is available and up to date. T
...
+```
+
+Then, for Ubuntu, use a shell script to download the current winetricks script(s).
+E.g.:
 
 ```
-sudo -- sh -c '
-cat > /usr/local/bin/update_winetricks.sh <<_EOF_SCRIPT
+# Create and switch to a temporary directory writeable by current user. See:
+#   https://www.tldp.org/LDP/abs/html/subshells.html
```
austin987: Is there a reason you don't mention man mktemp here (but do on the second invocation)?

README.md
```diff
@@ -28,34 +28,110 @@ The ```winetricks``` package should be used if it is available and up to date. T
...
+man mv
+man wget
+man sudo
+...
+```
+
+An alternative updater script implementation, using **su** in place of **sudo**, is also possible:
+
+```
+cd
```
austin987: While I'm fine with doing this in a user's $HOME, I think you meant to may have meant to use a tempdir (for consistency), since this is the only place using an empty cd.

README.md
```diff
@@ -65,29 +141,31 @@ Persistent=true
...
 
 [Service]
-ExecStart=/usr/local/bin/update_winetricks.sh
+ExecStart=/usr/bin/update_winetricks
 Type=oneshot
-_EOF_SERVICE
+_EOF_SERVICE_UNIT
+
+sudo mv winetricks_update.timer winetricks_update.service /etc/systemd/system/
 '
```
austin987: Again, stray apostrophe.

README.md
bobwya: Nice catch.

README.md
bobwya: I've made this consistent now.

README.md
bobwya: Probably!

PR #1157, opened 2018-12-19
use $WINETRICKS_GUI instead of direct zenity/kdialog call

vitlav:
Use $WINETRICKS_GUI prevents direct requires to zenity/kdialog binaries (it is need for correct rpm packing)

### DIFF
--- a/src/winetricks
+++ b/src/winetricks
@@ -3530,16 +3530,16 @@ winetricks_mkprefixmenu()
 
     case $WINETRICKS_GUI in
         zenity)
-            zenity --forms --text="" --title "$_W_msg_title" \
+            $WINETRICKS_GUI --forms --text="" --title "$_W_msg_title" \
                 --add-combo="$_W_msg_arch" --combo-values=32\|64 \
                 --add-entry="$_W_msg_name" \
                 | sed -e 's/^\s*|/64|/' -e 's/^/arch=/' -e 's/|/ prefix=/'
             ;;
         kdialog)
-            kdialog --title="$_W_msg_title" \
+            $WINETRICKS_GUI --title="$_W_msg_title" \
                 --radiolist="$_W_msg_arch" 32 32bit off 64 64bit on \
                 | sed -e 's/^$/64/' -e 's/^/arch=/'
-            kdialog --title="$_W_msg_title" --inputbox="$_W_msg_name" \
+            $WINETRICKS_GUI --title="$_W_msg_title" --inputbox="$_W_msg_name" \
                 | sed -e 's/^/prefix=/'
             ;;
     esac


austin987:
The change is fine (so I'll merge, but I'm curious about two things):
A) What exactly does this break for RPM? Could you point me to the bug/workaround currently used?
B) There are other uses of zenity/kdialog directly, that this patch doesn't address, e.g.,
https://github.com/Winetricks/winetricks/blob/master/src/winetricks#L3433
https://github.com/Winetricks/winetricks/blob/master/src/winetricks#L3671

do those not cause problems?

vitlav:
> A) What exactly does this break for RPM? Could you point me to the bug/workaround currently used?
RPM's autoreq adds Requires both on kdialog and zenity to package if they are called directly.
Instead, we need kdialog and zenity as optional dependencies.

> do those not cause problems?
There are no direct calls:
        kdialog)
            $WINETRICKS_GUI

austin987:
IMO, if autoreq checks for direct calls, it should see that the only direct calls are in winetricks_detect_gui():
https://github.com/Winetricks/winetricks/blob/master/src/winetricks#L3226

That said, if it has problems with zenity/kdialog being called when guarded by $WINETRICKS_GUI, then these are also problems:
https://github.com/Winetricks/winetricks/blob/master/src/winetricks#L3433
https://github.com/Winetricks/winetricks/blob/master/src/winetricks#L3671
https://github.com/Winetricks/winetricks/blob/master/src/winetricks#L3756
https://github.com/Winetricks/winetricks/blob/master/src/winetricks#L3884
https://github.com/Winetricks/winetricks/blob/master/src/winetricks#L3937
^ (note: these are behind a printf, which may be why they're not noticed).

These aren't behind printf:
vhttps://github.com/Winetricks/winetricks/blob/master/src/winetricks#L644
https://github.com/Winetricks/winetricks/blob/master/src/winetricks#L3532

IMO, autoreq isn't detecting dependencies properly. That said, I'm fine with changing winetricks to suit it, but let's not pretend it's as simple as this PR makes it seem.

vitlav:
> IMO, if autoreq checks for direct calls, it should see that the only direct calls are in winetricks_detect_gui():
> https://github.com/Winetricks/winetricks/blob/master/src/winetricks#L3226
> 
> That said, if it has problems with zenity/kdialog being called when guarded by $WINETRICKS_GUI, then these are also problems:
> https://github.com/Winetricks/winetricks/blob/master/src/winetricks#L3433

```

    case $WINETRICKS_GUI in
        zenity)
            printf %s "zenity \

```
There is no direct command call at all.
zenity) — just a case entry
printf — just print string "zenity"

> vhttps://github.com/Winetricks/winetricks/blob/master/src/winetricks#L644
> https://github.com/Winetricks/winetricks/blob/master/src/winetricks#L3532

`*zenity) W_KEY=$(zenity --entry --text "$_W_keymsg") ;;
`
it is possible, autoreq does not check subshell

> IMO, autoreq isn't detecting dependencies properly. That said, I'm fine with changing winetricks to suit it, but let's not pretend it's as simple as this PR makes it seem.
My rpm is happy with this PR :)

PR #1163, opened 2018-12-29
new verb: xact_x64

kotarou3:
Currently xact by itself only installs the 32-bit DLLs due to a *Microsoft* bug: https://bugs.winehq.org/show_bug.cgi?id=41618#c5
But some applications (e.g., Final Fantasy XIV in DX11 mode) require the 64-bit versions, and don't hit the aformentioned bug.

So provide a way to install the 64-bit xact DLLs for those use cases.

### DIFF
--- a/src/winetricks
+++ b/src/winetricks
@@ -301,6 +301,16 @@ w_killall()
     kill -s KILL $(pgrep $1)
 }
 
+# Some packages don't support win32, die with an appropriate message
+# Returns 64 (for tests/winetricks-test)
+w_package_unsupported_win32()
+{
+    if [ "$W_ARCH" = "win32" ] ; then
+        w_warn "This package ($W_PACKAGE) does not work on a 32-bit installation. You must use a prefix made with WINEARCH=win64."
+        exit 64
+    fi
+}
+
 # Warn user if package is broken. Optionally provide a link to the bug report.
 w_package_broken_win64()
 {
@@ -11026,7 +11036,7 @@ load_wsh57()
 #----------------------------------------------------------------
 
 w_metadata xact dlls \
-    title="MS XACT Engine" \
+    title="MS XACT Engine (32-bit only)" \
     publisher="Microsoft" \
     year="2010" \
     media="download" \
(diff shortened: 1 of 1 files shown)


austin987:
@aeikum any thoughts?

aeikum:
It's not as simple as that. Any game can hit this bug, it depends on how the game and Wine happen to allocate memory. If the xaudio2 objects end up using the top couple bits of the address, then the 64-bit xaudio2 DLLs will crash.

I guess I'm not opposed if there are sufficient warnings to the user that they're known to be problematic in many cases. Also, the FAudio work should be upstreamed soon, which will hopefully make native xaudio2 far less useful.

src/winetricks
```diff
@@ -11069,6 +11069,51 @@ load_xact()
 
 #----------------------------------------------------------------
 
+if test "$W_ARCH" = "win64"; then
```
austin987: This is unnecessary (and will probably cause problems in GUI mode)

src/winetricks
```diff
@@ -11069,6 +11069,51 @@ load_xact()
...
+        title="MS XACT Engine (64-bit only)" \
+        publisher="Microsoft" \
+        year="2010" \
+        media="download" \
+        file1="../directx9/directx_Jun2010_redist.exe" \
+        installed_file1="$W_SYSTEM64_DLLS_WIN64/xactengine2_0.dll"
+
+    load_xact_x64()
+    {
+        w_package_warn_win64 # https://bugs.winehq.org/show_bug.cgi?id=41618#c5
```
austin987: This should instead be:
```
w_package_unsupported_win64 # to block 32-bit prefixes
    if w_workaround_wine_bug 41618; then
        w_warn "While this helps some games, it completely breaks others. You've been warned."
    fi
```

`w_package_warn_win64` is intended for verbs that work on win32/win64, but have issues on win64.

src/winetricks
kotarou3: This was so the `$W_SYSTEM64_DLLS_WIN64` below would expand correctly. But removed

src/winetricks
kotarou3: Done

austin987:
Merged (after I manually sorted a couple things).

https://github.com/Winetricks/winetricks/commit/cd3a2821ba8e850bbc4217153c142f60fbc251d6

Issue #1162, opened 2018-12-29
New verb for FAudio (with ffmpeg support)


tgurr:
https://github.com/Kron4ek/FAudio-Builds is offering pre-built FAudio builds with FFmpeg support for WMA playback compiled in. Would be nice to have this or any alternative as a verb to make use of the FAudio improvements required for some games to work without major sound issues.

austin987:
@aeikum any thoughts?

aeikum:
I think it's using the cpp/ stuff from FAudio instead of the COM support in Wine itself, which is more correct. So this may introduce crashes. But if users find it useful, I think it's fine to do something like this.

FWIW I'm planning to merge the FAudio work upstream ASAP after the 4.0 release. This may still be useful for users on older Wine releases.

Gcenx:
It would be useful for same problematic games

austin987:
Made a PR, please test it if you can.

austin987:
Fixed by 6843ab3

lemonzest79:
Just noticed this is pulling a version from 17th of Jan 2019, the latest is 19.03 released 1st March 2019, noticed this when I did a fresh install of a new prefix earlier, could this be fixed please?

austin987:
Please open an issue for that, thanks.

PR #1168, opened 2019-01-26
faudio: new verb

austin987:
For #1162.

Needs testing, but I don't have things set up to game, so hoped y'all could ;)

@tgurr @aeikum @Gcenx (and fyi @Kron4ek)

### DIFF
--- a/src/winetricks
+++ b/src/winetricks
@@ -8468,6 +8468,48 @@ load_esent()
 
 #----------------------------------------------------------------
 
+w_metadata faudio dlls \
+    title="FAudio (xaudio reimplemntation, with xna support) builds for win32" \
+    publisher="Kron4ek" \
+    year="2019" \
+    media="download" \
+    file1="faudio-git-a756af4.tar.xz" \
+    installed_file1="$W_SYSTEM32_DLLS_WIN/FAudio.dll"
+
+load_faudio()
+{
+    # FIXME: they have a LATEST file that may be useful in the future
+    # Not sure if we want to have multiple versions like dxvk
+    # Especially when some are prefixed with 'git-'!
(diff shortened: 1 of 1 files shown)


Kron4ek:
Tested, it works properly.

I also added STABLE file into repo, you can use it in the verb. Now there are two:

LATEST - contain latest available build version (including git builds)
STABLE - contain latest avaiable stable build version (19.01 for now)

Also to note, it's ok that libgcc_s_sjlj-1.dll is missing in 64-bit build, it required only for 32-bit builds.

aeikum:
Looks fine to me, too.

bobwya:
The verb works OK for me too...

@aeikum 

Somewhat OT... I was kind of hoping that using **faudio** would help with the scratchy audio in Fallout 4... However it looks like I still need to use PULSE_LATENCY_MSEC though, for that issue... Yeah??

aeikum:
@bobwya That's a complicated question and definitely OT. Are you using the winepulse patches from wine-staging? They hopefully address that kind of problem (i.e. you should *not* have to use that environment variable if you are using those patches). If they don't fix it for you, let's move this discussion to wine bug 45623.

bobwya:
@aeikum Yeah sorry, a "slight sideways" issue hijack! I've tried both Proton 3.16-5 and Wine Staging 4.0 (with this winetricks verb). I do want to see if a high CPU load is trigging this issue... I'll report back on wine bug 45623.

I guess I thought **faudio** would be the "magic bullet" for all audio ills... :-)

Gcenx:
It’s also working fine for me also

austin987:
Fixed by https://github.com/Winetricks/winetricks/commit/6843ab37c458953bb55371381b518fa326144dde

PR #1171, opened 2019-01-26
steam verb: drop Windows XP workaround, for 39403

bobwya:
* Steam client does not support Windows XP, as of January 2019.


bobwya:
@austin987 

With the existing **Windows XP** **winetricks** workaround in place for the Steam Client, installed in a clean **WINEPREFIX**, will install. But the actual downloaded Steam Client will not start up. So it would probably be best to remove this workaround ASAP (/ merge this pull request).

PR #1174, opened 2019-02-03
add pacman

morteryler:
need editing

### DIFF
--- a/src/winetricks
+++ b/src/winetricks
@@ -423,7 +423,7 @@ w_try_cabextract()
 {
     # Not always installed, but shouldn't be fatal unless it's being used
     if test ! -x "$(command -v cabextract 2>/dev/null)"; then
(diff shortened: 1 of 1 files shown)


src/winetricks
```diff
@@ -423,7 +423,7 @@ w_try_cabextract()
 {
     # Not always installed, but shouldn't be fatal unless it's being used
     if test ! -x "$(command -v cabextract 2>/dev/null)"; then
-        w_die "Cannot find cabextract.  Please install it (e.g. 'sudo apt-get install cabextract' or 'sudo yum install cabextract')."
+        w_die "Cannot find cabextract.  Please install it (e.g. 'sudo apt-get install cabextract' or 'sudo yum install cabextract' or 'sudo pacman -S cabextract')."
```
austin987: There shouldn't be two 'or's. Instead:
`w_die "Cannot find cabextract.  Please install it (e.g. 'sudo apt-get install cabextract', 'sudo yum install cabextract' or 'sudo pacman -S cabextract')."`

Also note that there are several other places where packages are recommended, you should update all of them.

PR #1181, opened 2019-02-22
Gallium Nine bits

dhewg:
This splits the existing verb galliumnine02 into a helper and the verb using it, then adds a verb for the new v0.3, which is also using the helper (like dxvk).

Additionally, there's also a version independent verb added, which uses the github api to fetch the latest release. We get a json formatted response, which I currently 'parse' using `grep` and `cut`.

Thanks!


austin987:
Thank you!

PR #1184, opened 2019-03-02
Improve Chinese translation

njlyf2011:


austin987:
Manually rebased since I just made a conflicting commit.

https://github.com/Winetricks/winetricks/commit/70197073b5b10ab906536f7407f10c974e322b6f

Please only submit one commit in the future if it's for the same thing.

PR #1188, opened 2019-03-14
new verb: dxvk101

bobwya:


bobwya:
@austin987 

Yeah... probably due to extreme sleep deprivation :-)
I've pushed a fix.

Issue #1192, opened 2019-03-17
Provide an easy way to install winetricks?


WPFilmmaker:
Hi, thank you for your work on winetricks, it is really helpful.

I would like to request providing an easy way to install winetricks like flatpak which works on every distro.

Right now you either get mostly out of date versions from your distro's repo or you have to follow a lengthy and not noob-proof guide...

austin987:
Looks like a significant amount of work on my end. I may at some point when I have not time.

In the meantime, a single wget call can download it..

bobwya:
> I would like to request providing an easy way to install winetricks like flatpak which works on every distro.

That's a crazy idea... You'd end up with high maintenance and packaging overheads, for what amounts to nothing more than installing than 2 shell scripts!

Out of idle curiosity I took a look at the current (as of today) Linux distribution packaging of **winetricks**:

- **Arch**: 20190310
- **Debian**: _20170101_
- **Fedora**: 20181203 (testing: 20190310)
- **Gentoo**: 20190310
- **Opensuse**: 20190310
- **Ubuntu**: _20180217_ (bionic) / _20180815_ (cosmic) / _20181203_ (disco)

So it's _abundantly_ _clear_ where the stagnation, in packaging up-to-date **winetricks**, is a significant issue... The vast majority of larger Linux distributions clearly manage to package 2 simple shell scripts, in a timely fashion...

WPFilmmaker:
@austin I am no dev, so I am just throwing my two cents from a user pov.

What I find a bit paradoxical is that a useful software like winetricks which  improve WINE's user experience but then to install a recent version you still need to use bash script which according to your page requires minimum 8 steps... 

@bobwya I apologize for not mentioning that i am on a LTS, so my request was more about that.

jre-wine:
Debian testing:  20181203
I take the blame on me for not coordinating with Austin to make a release before Debian testing went into freeze (final day for upload was 20190301).

@bobwya Do other distributions with a stable release cycle update winetricks in existing releases? Serious question, but I doubt so. Of course for distros with rolling releases this is not an issue. I ask you to stop painting an incomplete picture and offering your conclusions from that.

I'm interested in constructive suggestions how to handle winetricks as packager better.  E.g. since 20181203 you can use "winetricks --selfupdate" again in the Debian/Ubuntu packages (I just patched in a warning, so that users are aware that Debian has no control over that version). However, for privacy reasons the automatic version check is disabled.

[...]

PR #1194, opened 2019-03-17
README.md: Add Quick install section

jre-wine:
For https://github.com/Winetricks/winetricks/issues/1192.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,19 +26,32 @@ The ```winetricks``` package should be used if it is available and up to date. T
 * Slackbuilds (Slackware): http://slackbuilds.org/repository/14.2/system/winetricks/?search=winetricks
 * Ubuntu: https://packages.ubuntu.com/search?keywords=winetricks
 
-Note: packaged Debian / Ubuntu winetricks versions are typically outdated, so a manual installation is recommended.
-
 If the package is unavailable, outdated, or the latest version is desired, a manual installation of winetricks can be done.
+
+# Quick install
+
+To quickly install winetricks open a terminal and type
+~~~~~
+wget  https://raw.githubusercontent.com/Winetricks/winetricks/master/src/winetricks
+~~~~~
+
+You can then run it from a terminal with
+~~~~~
+sh winetricks
+~~~~~
+
+# Full installation
+
 It is _highly_ recommended to uninstall any previously installed version of winetricks first.
 
-**_If you don't uninstall a previously installed, packaged version of winetricks... Well then you get to pick up the pieces!_**
+**_If you don't uninstall a previously installed version of winetricks, that installation might be used instead of your manual installation._**
 
 E.g. for Debian / Ubuntu:
 ```
-sudo apt-get purge winetricks
+sudo apt purge winetricks
 ```
 
-Then, for Ubuntu, use a shell script to download the current winetricks script(s).
+Then (tested on Ubuntu), use a shell script to download the current winetricks script(s).
 E.g.:
 
 ```


austin987:
Lgtm, but maybe should include instructions to install it in /usr/local/bin? That way it's in $PATH and doesn't require invoking with sh /full/path

qwertychouskie:
Honestly the quick install script should be just:
```bash
sudo curl -o /usr/local/bin/winetricks https://raw.githubusercontent.com/Winetricks/winetricks/master/src/winetricks
sudo chmod +x /usr/local/bin/winetricks
```
And the full install:
```bash
sudo curl --create-dirs -o /usr/local/bin/winetricks https://raw.githubusercontent.com/Winetricks/winetricks/master/src/winetricks
sudo chmod +x /usr/local/bin/winetricks
sudo curl --create-dirs -o /usr/local/share/man/man1/winetricks.1 https://raw.githubusercontent.com/Winetricks/winetricks/master/src/winetricks.1
sudo curl --create-dirs -o /usr/local/share/applications/winetricks.desktop https://raw.githubusercontent.com/Winetricks/winetricks/master/src/winetricks.desktop
sudo curl --create-dirs -o /usr/local/share/metainfo/winetricks.appdata.xml https://raw.githubusercontent.com/Winetricks/winetricks/master/src/winetricks.appdata.xml
sudo curl --create-dirs -o /usr/local/share/icons/hicolor/scalable/apps/winetricks.svg https://raw.githubusercontent.com/Winetricks/winetricks/master/src/winetricks.svg
sudo curl --create-dirs -o /usr/local/share/bash-completion/completions/winetricks https://raw.githubusercontent.com/Winetricks/winetricks/master/src/winetricks.bash-completion
```
Much simpler and more complete than the old instructions.  However, curl seems to be creating the directories with `drwxr-x---` as the permissions (no read for standard users), maybe one of you knows a way around this that won't complicate things a bunch.

Actually, even better, add a `--self-install`  option to Winetricks, then the entire install would be:
```
wget -qO - https://github.com/Winetricks/winetricks/raw/master/src/winetricks | pkexec sh /dev/stdin --self-install
```
The `--self-install` function would `mkdir -p` the necessary directories, then download the files to their locations as above.  `self-install` would default to `/usr/local` as the prefix, while e.g. `self-install=/usr` would use `/usr` as the prefix.
```
qwerty@qwerty-Inspiron-3520:~$ wget -qO - https://github.com/Winetricks/winetricks/raw/master/src/winetricks | pkexec sh /dev/stdin --self-install
unknown option --self-install
qwerty@qwerty-Inspiron-3520:~$ 
```
Let's make that option known! :)

README.md
```diff
@@ -26,19 +26,32 @@ The ```winetricks``` package should be used if it is available and up to date. T
...
+# Quick install
+
+To quickly install winetricks open a terminal and type
+~~~~~
+wget  https://raw.githubusercontent.com/Winetricks/winetricks/master/src/winetricks
+~~~~~
+
+You can then run it from a terminal with
+~~~~~
+sh winetricks
```
melroy89, line 40: I will not advise it like this. I would first use chmod and just execute script as intended..

```suggestion
# Make executable
chmod +x ./winetricks

# Run
./winetricks
```

jre-wine: See new comment below. But thanks @melroy89

README.md
```diff
@@ -26,22 +26,32 @@ The ```winetricks``` package should be used if it is available and up to date. T
...
 
 If the package is unavailable, outdated, or the latest version is desired, a manual installation of winetricks can be done.
-It is _highly_ recommended to uninstall any previously installed version of winetricks first.
 
-**_If you don't uninstall a previously installed, packaged version of winetricks... Well then you get to pick up the pieces!_**
+### Quick install for one time usage
+
+To quickly grab current winetricks open a terminal and type
+~~~~~
+wget  https://raw.githubusercontent.com/Winetricks/winetricks/master/src/winetricks
```
melroy89, line 38: Ps. I never seen `~~~~` but I think ``` is normally used for a code block.

Like so:

```sh
echo "Hello world";
````

Or python highlighting:

```py
print("hello world");
```

jre-wine: Thanks, I fixed the formatting.

There's an "Outdated" label in your comment. Not sure what that implies.

melroy89: That means, if you are unfamiliar with GitHub PRs, you committed a change this this file. Making the current code/markdown I see above outdated. Hence the reason GitHub marks it now as outdated. 

Ps. Which doesn't mean the code is actually fixed correctly, again only saying there is a newer version.

jre-wine: Ah, thanks. So I marked this comment as resolved.

jre-wine:
Updated and added "-O winetricks" to overwrite a previously downloaded winetricks. Otherwise the new one would get a suffix and the user would execute the old one.

I'm open to both changing this to /usr/local/bin/ and/or chmod +x. Above is just my reasoning for doing it otherwise, but each solution qualifies as a "Quick install", so I have no hard opinion on this.

PR #1196, opened 2019-03-24
Update vcrun2013 to 12.0.40660.0

Faalagorn:
Minor version bump, from 12.0.30501.0 to 12.0.40660.0. Apparently fixes "some floating-point math library functions", details at the link in commit.

Link to latest version originally at https://support.microsoft.com/en-gb/help/2977003/the-latest-supported-visual-c-downloads

### DIFF
--- a/src/winetricks
+++ b/src/winetricks
@@ -10784,9 +10784,10 @@ w_metadata vcrun2013 dlls \
 
 load_vcrun2013()
 {
-    # https://www.microsoft.com/en-us/download/details.aspx?id=40784
+    # https://support.microsoft.com/en-gb/help/3179560/update-for-visual-c-2013-and-visual-c-redistributable-package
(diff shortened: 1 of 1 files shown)


Faalagorn:
Should also theoretically fix the bug from [KB 3138367](https://support.microsoft.com/en-us/help/3138367/update-for-visual-c-2013-and-visual-c-redistributable-package) recommended from [the previous download](https://www.microsoft.com/en-us/download/details.aspx?id=40784), as these ones are version `12.0.40649.5`, but I don't know how much effect does it have in practice:

> When you use Microsoft Visual C++ 2013, you may encounter the following issues:
>- Heap corruption occurs because of a synchronization bug in SetLocale, and this triggers an access violation.
>- When you install an updated redistributable package, binaries for non-target architectures are removed. For example, after you install an update for an x86-based application, the x64 Visual C++ 2013 runtime libraries are missing. This fix makes sure that both versions of the Visual C++ redistributable are visible when you add or remove programs after installation of the update.

PR #1208, opened 2019-04-10
Update 7zip Download Link

### LINKED ISSUE #1142: 7-Zip URL obsolete, needs to be updated
When winetricks b7e9e33738835c59e5cfb44e01261327d234e64e is run (tested using a clean installation of Linux Mint 19 XFCE 32-bit) and 7-Zip is not installed, winetricks tries to download `7z1602.exe` from SourceForge and fails, and then from archive.org and fails.

Valid links:
https://www.7-zip.org/a/7z1805.exe
https://www.7-zip.org/a/7z1604.exe
https://www.7-zip.org/a/7z920.exe

ukos-git:
Sourceforge does not resolve the provided link anymore. Instead, A web
page is downloaded that does not match the sha256sum.

The newly provided link makes installation work again.

Closes #1142

### DIFF
--- a/src/winetricks
+++ b/src/winetricks
@@ -12469,7 +12469,7 @@ w_metadata 7zip apps \
 
 load_7zip()
 {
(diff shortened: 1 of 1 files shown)


austin987:
Thanks!

PR #1211, opened 2019-04-14
Several improvements for DirectShow

zfigura:
I regularly cross-test using native DLLs, so verbs for qcap and directshow are handy. I expect they may end up being handy for other users as well.

The third patch is not limited to DirectShow, but was motivated by it.


gverm:
@zfigura In the load_directx9() verb, qcap is installed manually. I think it would be a good idea to replace that with your qcap verb.

zfigura:
Indeed, I've added an extra patch for this. Thanks.

src/winetricks
```diff
@@ -6623,6 +6624,27 @@ load_directmusic()
...
+load_directshow()
+{
+    w_warn "You can specify individual DirectShow verbs instead. e.g. 'winetricks quartz'"
+
+    w_call quartz
+    w_call qasf
+    w_call qcap
+    w_call qdvd
+    w_call qedit
+    w_call amstream
```
austin987: On mobile, so can't test, but do they need to be in this order?

zfigura: No, the order shouldn't matter as far as I'm aware.

austin987: Could you sort them please? (Sorry, that should've been in my first comment)

zfigura: Yep, done.

PR #1213, opened 2019-04-16
README: improve Installation and Updating guides

bobwya:
Responding to further user feedback... :-)

Updates to README:

 * supply a simplified updater script (no comments / streamlined) that
   is easy to setup, for new users
 * credit to @kenorb for suggesting the use of curl (replacing use of: wget)
 * curl / install commands+options are available cross-platform on Linux / FreeBSD
 * note: `winetricks --self-update` is **not** the same - this will only
   update to a newer release and will not update the bash.completion script

Fixes: https://github.com/Winetricks/winetricks/issues/1192

Signed-off-by: Rob Walker <bob.mt.wya@gmail.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,7 +13,7 @@ https://raw.githubusercontent.com/Winetricks/winetricks/master/src/winetricks
 Tagged releases are accessible here:
 https://github.com/Winetricks/winetricks/releases
 
-# Installing
+# Installing (Package)
 The ```winetricks``` package should be used if it is available and up to date. The package is available in most mainstream (Unix-like) Operating Systems:
 
 * Arch: https://www.archlinux.org/packages/community/any/winetricks/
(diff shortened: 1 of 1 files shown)


qwertychouskie:
Honestly `--self-update` should update the whole thing, including the completions file.  Also the `--self-install` proposed at https://github.com/Winetricks/winetricks/pull/1194#issuecomment-474615556 would make this a lot simpler.

austin987:
I agree with @qwertychouskie, I can't think of any reason why bash-completion shouldn't be updated, particularly since it would require no user effort and just a few lines of code.

@bobwya do you have strong reason for NOT updating bash-completion?

bobwya:
@austin987 well my proposed changes do update the winetricks **bash-completion** script as well. 

I've nothing against imrpoving the **--self-update** option or adding a **--self-install**. The instructions in the README are always going to have to include a manual download of winetricks (due to folks running obsolete versions) - so they aren't going to get much simpler... At least with the proposed changes - really just an iterative improvement - they will be simpler to follow.

qwertychouskie:
@bobwya The idea with `--self-install` is that users just paste this command:
```
wget -qO - https://github.com/Winetricks/winetricks/raw/master/src/winetricks | pkexec sh /dev/stdin --self-install
```
That way there is no dependency on any installed version of Winetricks, or lack thereof.

bobwya:
@qwertychouskie 

Technically you've not invalidated my point (since your **are** manually downloading winetricks). But yes that is a nice one liner!

PR #1216, opened 2019-04-18
Small change that fixes the removal of Wine Mono

sebi2306:
This is one of the ways to fix Winetricks#1215. It should suffice as "WIne Mono Windows Support" also gets removed when removing "Wine Mono Runtime".

### DIFF
--- a/src/winetricks
+++ b/src/winetricks
@@ -20669,7 +20669,7 @@ w_metadata remove_mono settings \
 
 load_remove_mono()
 {
-    mono_uuid="$("${WINE_ARCH}" uninstaller --list | grep 'Wine Mono' | cut -f1 -d\|)"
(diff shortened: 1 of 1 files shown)


austin987:
That would break every previous version of wine-mono which doesn't have that installer.

sebi2306:
I'm thinking of just using the first GUID then. Should we also account for the off chance that someone has two versions of Mono installed?

austin987:
I committed a fix for it, thank you for the report:
https://github.com/Winetricks/winetricks/commit/7b4e62d481f631452b0f19f0209e48541ebc1f4b

It shouldn't be possible to install two versions of mono, but as you said, uninstalling the runtime will remove the windows support installer to. So, I kept the old search, but removed windows support; so it will work on older wine and wine-4.6 (already tested).

sebi2306:
Glad I could be of help :)

Issue #1215, opened 2019-04-18
Winetricks fails to remove Mono


sebi2306:
Winetricks fails to remove Mono because the `--remove` parameter gets passed 2 GUIDs. To fix this either the `grep` should be changed or it should be attempted to remove all GUIDs found one by one.

![Winetricks Oopsie1](https://user-images.githubusercontent.com/6190216/56379848-62096c80-6219-11e9-95b4-5925411f8d19.png)
![Winetricks Oopsie5](https://user-images.githubusercontent.com/6190216/56379849-62096c80-6219-11e9-94e8-fde8347a89ad.png)

sebi2306:
Please ignore my derp

austin987:
Fixed by https://github.com/Winetricks/winetricks/commit/7b4e62d481f631452b0f19f0209e48541ebc1f4b

sebi2306:
Yeah, not quite fixed. The problem is that the `--remove` parameter cannot get both GUIDs that get found (one from Mono Runtime and the other one from Windows Support) so it returns 
> uninstaller: The application with GUID '{7290EA95-9F17-58E8-8436-4A532DA0F54E}
{05706910-3B4C-5B10-9F47-759A1CDDECDE}' was not found

austin987:
Are you running from git? For me with wine-4.6-30-gf9301c2b66, I get:

```
+ load_remove_mono ''
++ /home/austin/wine-git/wine uninstaller --list
++ grep 'Wine Mono'
++ grep -v 'Wine Mono Windows Support'
++ cut -f1 '-d|'
+ mono_uuid='{7290EA95-9F17-58E8-8436-4A532DA0F54E}'
+ test '{7290EA95-9F17-58E8-8436-4A532DA0F54E}'
+ /home/austin/wine-git/wine uninstaller --remove '{7290EA95-9F17-58E8-8436-4A532DA0F54E}'
```

and with older wine (wine-3.16):
```
+ load_remove_mono ''
++ /opt/oldwine/wine-3.16/bin/wine uninstaller --list
++ grep 'Wine Mono'
++ grep -v 'Wine Mono Windows Support'
++ cut -f1 '-d|'
+ mono_uuid='{EA50D369-6B67-43BC-8153-C3553B40ABEB}'
+ test '{EA50D369-6B67-43BC-8153-C3553B40ABEB}'
[...]

PR #1217, opened 2019-04-18
remove_mono: remove all matching GUIDs

sebi2306:
An actual fix for #1215


austin987:
Squashed and merged as https://github.com/Winetricks/winetricks/commit/b18fae574f80968580084444c95e208ae3cbed3f; thanks!

PR #1224, opened 2019-05-05
dxvk verb update

bobwya:
@austin987 

I've updated all the dxvk verbs as discussed on IRC. I've also switched over to the Github API (release.json) for pulling the latest release version.

### DIFF
--- a/src/winetricks
+++ b/src/winetricks
@@ -7321,6 +7321,25 @@ load_dxvk103()
     helper_dxvk_d9vk "$file1" "4.5" "1.1.101" "dxgi,d3d10,d3d11"
 }
 
+w_metadata dxvk111 dlls \
+    title="Vulkan-based D3D10/D3D11 implementation for Linux / Wine (1.1.1)" \
+    publisher="Philip Rebohle" \
+    year="2017" \
+    media="download" \
+    file1="dxvk-1.1.1.tar.gz" \
+    installed_file1="$W_SYSTEM32_DLLS_WIN/d3d10.dll" \
+    installed_file2="$W_SYSTEM32_DLLS_WIN/d3d10_1.dll" \
+    installed_file3="$W_SYSTEM32_DLLS_WIN/d3d10core.dll" \
+    installed_file4="$W_SYSTEM32_DLLS_WIN/d3d11.dll" \
+    installed_file5="$W_SYSTEM32_DLLS_WIN/dxgi.dll"
+
+load_dxvk111()
+{
+    # https://github.com/doitsujin/dxvk
+    w_download "https://github.com/doitsujin/dxvk/releases/download/v1.1.1/dxvk-1.1.1.tar.gz" 346c523953f72ac5885071c4384039faf01f6f43a88d5b0c12d94bfaa9598c1d
+    helper_dxvk_d9vk "$file1" "4.5" "1.1.104" "dxgi,d3d10,d3d11"
+}
+
 
 #----------------------------------------------------------------
 
--- a/src/winetricks
+++ b/src/winetricks
@@ -6823,173 +6823,201 @@ load_dxsdk_jun2010()
 #----------------------------------------------------------------
 
 # $1 - dxvk archive name (required)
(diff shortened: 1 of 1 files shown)


src/winetricks
```diff
@@ -7310,12 +7325,11 @@ w_metadata dxvk dlls \
 load_dxvk()
 {
     # https://github.com/doitsujin/dxvk
-    # There's no stable exe URL, but they do provide a RELEASE file that lets us build one:
-    w_download_to "${W_TMP_EARLY}" "https://raw.githubusercontent.com/doitsujin/dxvk/latest-release/RELEASE"
-    dxvk_version="$(cat "${W_TMP_EARLY}/RELEASE")"
-    w_linkcheck_ignore=1 w_download "https://github.com/doitsujin/dxvk/releases/download/v${dxvk_version}/dxvk-${dxvk_version}.tar.gz"
-    helper_dxvk "dxvk-${dxvk_version}.tar.gz" "d3d10_enabled" "3.19" "1.1.88"
-    unset dxvk_version
+	w_download_to "${W_TMP_EARLY}" "https://api.github.com/repos/doitsujin/dxvk/releases/latest" "" "release.json"
```
austin987: Stray tab.

src/winetricks
```diff
@@ -7310,12 +7325,11 @@ w_metadata dxvk dlls \
...
+    _W_dxvk_version="$(awk -F':' '{ gsub("[\" ,]",""); if ($1 == "tag_name") {sub("^v","",$2); print $2; exit} }' "${W_TMP_EARLY}/release.json")"
```
austin987: Not necessarily a blocker, but this is broken on aria2c, see https://github.com/aria2/aria2/issues/1389.

I was thinking we could make a helper for github API release checks, then have two codepaths, one for curl/wget, one for aria2c (not sure about fetch).

src/winetricks
```diff
@@ -6825,13 +6825,11 @@ load_dxsdk_jun2010()
 # $1 - dxvk archive name (required)
 # $2 - dxvk d3d10_enabled / d3d10_disabled (required)
 # $3 - minimum Wine version (optional)
-# $4 - minimum Vulkan API version (optional, requires $3 be set)
```
austin987: I'm still not clear on why this needs to go. It doesn't show for everyone, only if it fails the check, which to me is a good thing. Is the check broken in some way? Can we fix it instead?

bobwya: @austin987 

Since I can't get a clear answer from Phillip around the Vulkan version requirements... He talks in terms of required extensions. I gather this is mainly an issue around graphics drivers - that may implement extensions, required for DXVK, but do so on top of older versions of Vulkan (that wouldn't normally support these additional extensions)...

Nobody is going to have a good experience with a recent version of DXVK and Wine 3.x (whatever)... Vulkan 1.1.88 is very out-of-date vs. the very recent driver requirements DXVK imposes on Mesa and the Nvidia driver.

I could suggest Wine and Vulkan versions, based closely matching the Vulkan headers that Wine pulls in and that DXVK pulls in. This would be a "recommended target" (in that DXVK may not use all the features of the Vulkan header version being pulls in) - but at least it would be a meaningful guideline. Rather than the current warning that is displayed and has little basis in reality... 

Thoughts?!
Bob

bobwya:
@austin987 I've reworked and tidied up all the **dxvk** verbs and added support for: [d9vk](https://github.com/Joshua-Ashton/d9vk) - which is heading rapidly towards a useable state (with some quick testing)... **d9vk** supports D3D9 games, by translating D3D9 calls to D3D11, using **dxvk**+Vulkan.

austin987:
Changing verb names breaks computability. You'll want to put a shim, look towards the end of the file. Adding a warning in the shim would probably be good too.

bobwya:
@austin987 I've added the **dxvk0*** compatibility shims, as you requested!

src/winetricks
austin987: Makes sense, thanks.

Issue #1207, opened 2019-04-10
galliumnine - Download broken while using aria2c


varris1:
winetricks version: 20190310-next - sha256sum: 908b1820d942393c188e4a714478b43f2d61f8ff8f914dcc0ee582a575241e56
[winetricks output](https://github.com/Winetricks/winetricks/files/3063504/output.log)

While winetricks is using aria2c, downloading gallium nine is broken because aria2c tries to download g9 standalone from: `https://github.com/iXit/wine-nine-standalone/releases/download/https://api.github.com/repos/iXit/wine-nine-standalone/releases/16375260/gallium-nine-standalone-https://api.github.com/repos/iXit/wine-nine-standalone/releases/16375260.tar.gz
`

Using wget seems to work fine.

austin987:
Looks like an aria2c bug; it corrupts the release.json when it downloads. I tested downloading it manually with aria2c/curl/wget; curl/wget match, aria2c doesn't:

```
austin@laptop:/tmp/tmp.3EPAPPKzKn$ sha256sum *
393191249b9eb6fc60d63cc3f2a78b013b3c2ec23167ea7c0ffa902cbc2482b3  aria2c
11dd932af3e03e36745e99bc6639c625e7e83a8ab68a2bcb1e2346ce0efe1c95  curl
11dd932af3e03e36745e99bc6639c625e7e83a8ab68a2bcb1e2346ce0efe1c95  wget
```

I'd suggest filing a bug with them, as there's not much I can do (other than maybe warn that it's broken). Their tracker is at https://github.com/aria2/aria2/issues

austin987:
Actually, it doesn't corrupt it, it compresses the json, whereas wget/curl leave it as is. The current code assumes it isn't compressed.

Ideally aria2c wouldn't do that, but the code could probably be refactored to handle that.

austin987:
Fixed by https://github.com/Winetricks/winetricks/commit/ef270d9769d3dbe130c102e81e091dd64bccce4e

PR #1228, opened 2019-05-10
Use a helper to get github releases from github API

austin987:
For #1207 et al

@bobwya I was thinking something like this to handle the aria2c issue. I'm not set on it, obviously my sed/tr stuff could be improved, but I think this approach is simpler than trying to have one regex to rule them all. It also avoids harcoding downloader names, so should be future-safe (and good for when other downloaders break it ;)).

This also lets us it for other verbs, i.e., dxvk/etc.

What do you think? @bobwya @dhewg

### DIFF
--- a/src/winetricks
+++ b/src/winetricks
@@ -718,6 +718,26 @@ w_expand_env()
     winetricks_early_wine cmd.exe /c echo "%$1%"
 }
 
+# Get the latest tagged release from github.com API
+w_get_github_latest_release()
+{
+    # $1 - github organization
+    # $2 - github repository
+    # FIXME: can we get releases that aren't on master branch?
+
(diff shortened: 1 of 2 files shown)


bobwya:
@austin987 Go for it!

It's like my open pull-request, to improve the README. They are both just incremental improvements to the status quo. Better than the alternative... Which is waiting (potentially) indefinitely for someone else to develop the "perfect answer./solution"!

austin987:
That's more my fault for not looking. Maybe soonish, got a busy weekend ahead of me ;)

bobwya:
> That's more my fault for not looking. Maybe soonish, got a busy weekend ahead of me ;)

I was only making a general observation, really!

"Keyboard warriors" can be vocally critical of my proposed changes - but not actually offer a constructive alternative... C'est la vie I guess,,,

dhewg:
Just stumbled upon this. A little late to the party, but nice work and thx! ;)

PR #1238, opened 2019-05-19
new verb: dxvk121

bobwya:


bobwya:
@austin987  Joshua kindly re-uploaded the **d9vk 0.11** release - with a "fixed" tarball URI/format - at my request and corrected version strings. I didn't generate a **winetricks** verb for this **d9vk** release version, when it was originally released... Basically because it was a bit of a mess... It should be fine now, though!

PR #1240, opened 2019-05-24
icodecs: Make unattended installation faster and more reliable

bergmn:
Use the built in InstallShield silent mode of the installers instead of trying to script UI interactions with autohotkey.

This makes the installation faster and more reliable when doing unattended installations.

### DIFF
--- a/src/winetricks
+++ b/src/winetricks
@@ -9362,26 +9362,11 @@ load_icodecs()
     # 2014/04/11: http://www.cucusoft.com/codecdownload/codinstl.exe (linked from http://www.cucusoft.com/codec.asp)
     w_download "http://www.cucusoft.com/codecdownload/codinstl.exe" 0979d43568111cadf0b3bf43cd8d746ac3de505759c14f381592b4f8439f6c95
 
-    w_try_cd "$W_CACHE/$W_PACKAGE"
-
-    w_ahk_do "
-        SetTitleMatchMode, 2
-        run codinstl.exe
-        winwait, Welcome
-        if ( w_opt_unattended > 0 ) {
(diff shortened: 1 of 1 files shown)


austin987:
With wine-4.9, this has run for me for 9h without completing:
```
austin   21828 79.4  0.1 2679508 25912 ?       Ssl  07:45 542:19 C:\users\austin\Temp\IXP000.TMP\dasetup.exe /Q /N
austin   24736  0.0  0.1 2664728 17576 pts/30  Sl+  07:58   0:00 C:\windows\system32\winevdm.exe --app-name C:\windows\temp\_icodecs\codinstl\SETUP.EXE C:\windows\temp\_icodecs\codinstl\setup.exe /s
```

austin987:
Note: state for dasetup.exe switches between running and sleeping, not just sleeping

bergmn:
Oh, I see. I just noticed that wine 4.9 was released right after I created the pull request.
I will download the latest development version and have a look.

Where do dasetup.exe come into the picture? Is it from mdac28?

austin987:
Good point; there was a different issue with mdac in 4.8 that was fixed in 4.9, so I assumed this was a problem with the change you made. Testing with wine-4.0, it works.

I'll need to look closer, but it's likely a wine issue and not a problem with your PR.

bergmn:
Ok. I compiled wine-4.9 and it still seems to work for me with the game I where troubleshooting earlier. But in my case it's limited to just a couple of verbs.

I'm gonna run the full test suite to make sure.

austin987:
Yeah, looks like it was a temporary wine regression; merging. Thanks!

PR #1244, opened 2019-05-31
docs: add xz-utils dependency

scintill:
for tar -J. Perhaps Fedora needs a similar change, but I haven't tried.


austin987:
Thanks; I edited it a bit (and used the opportunity to cleanup the dependency info as well); thanks for the report.

https://github.com/Winetricks/winetricks/commit/3d410436bc7e6cd110b5b39df53aa46115dfd349

PR #1248, opened 2019-06-04
dinput8, dsound: Don't try to register native DLLs.

zfigura:
I went and checked each individual verb that https://github.com/Winetricks/winetricks/commit/caa28d1af97ddd3d7b3e76ca7f51fb85fe1ce7eb affected, and these two indeed fail, so this patch reverts the change for those two.


src/winetricks
```diff
@@ -8807,8 +8809,9 @@ load_dsound()
     w_try_cabextract -d "$W_TMP" -L -F dxnt.cab "$W_CACHE"/directx9/$DIRECTX_NAME
     w_try_cabextract -d "$W_SYSTEM32_DLLS" -L -F 'dsound.dll' "$W_TMP/dxnt.cab"
 
-    w_override_dlls native dsound
+    # Don't try to register native dinput8; it doesn't export DllRegisterServer().
```
austin987: copy/paste typo

zfigura: Thanks, fixed.

PR #1254, opened 2019-06-12
python27: new verb

aplaice:
I haven't updated `files/verbs/all.txt` and the like, since AFAICT they're automatically generated with `src/release.sh`.

I'm not sure whether keeping pywin32 installed together with base python is necessary, but I'm keeping it for compatibility with python26.

pywin32 is now on github, so fetching the latest version from there. See the notice at sourceforge:
https://sourceforge.net/projects/pywin32/

Tested both with and without W_OPT_UNATTENDED=1.

### DIFF
--- a/src/winetricks
+++ b/src/winetricks
@@ -10421,6 +10421,41 @@ load_python26()
 
 #----------------------------------------------------------------
 
+w_metadata python27 dlls \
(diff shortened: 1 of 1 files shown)


aplaice:
Thanks for merging!

PR #1262, opened 2019-06-16
Update debian packaging and ubuntu instructions

jre-wine:
I just wanted to remove the deprecated ppa ubuntu-wine, but then started updating debian/. While doing that I tried to minimize the diff to Debian's version.
Please have a look at the single commit messages for some notes.


austin987:
Looks good, thanks.

Issue #1270, opened 2019-07-06
d9vk_master prefers cached downloads over newer online copies


CartoonFan:
Hello! I just wanted to let you know that winetricks is preferring cached downloads over newer online versions, effectively freezing the software at a particular version/release/commit. This mainly seems to affect verbs that get the "latest" or "master" version of dlls or utilities without changing the filename, such as d9vk_master. Thanks!

austin987:
As far as I can tell, only d9vk_master is affected, as faudio/dxvk save the file with a version number, whereas d9vk_master always saves to `d9vk_master.zip`.

CartoonFan:
I agree. I didn't check the others until after I posted the bug.

PR #1279, opened 2019-07-15
 d9vk_master verb: pre-delete cache archive

bobwya:
* Gitlab doesn't provide a way to determine an artifact
   commit version. So do delete/update the currently cached archive
   - but only when the verb is forced. Obviously a more elegant
   solution is required!
 * fixes: https://github.com/Winetricks/winetricks/issues/1270
 * update Vulkan requirements

### DIFF
--- a/src/winetricks
+++ b/src/winetricks
@@ -6840,6 +6840,26 @@ load_d9vk013()
     helper_dxvk_d9vk "$file1" "4.5" "1.1.104" "dxgi,d3d9,d3d10,d3d11"
 }
 
+w_metadata d9vk013f dlls \
+    title="Vulkan-based D3D9/D3D10/D3D11 implementation for Linux / Wine (0.13f)" \
+    publisher="Joshua Ashton" \
+    year="2019" \
+    media="download" \
+    file1="d9vk-0.13f.tar.gz" \
+    installed_file1="$W_SYSTEM32_DLLS_WIN/d3d9.dll" \
+    installed_file1="$W_SYSTEM32_DLLS_WIN/d3d10.dll" \
+    installed_file2="$W_SYSTEM32_DLLS_WIN/d3d10_1.dll" \
+    installed_file3="$W_SYSTEM32_DLLS_WIN/d3d10core.dll" \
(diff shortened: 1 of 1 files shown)


bobwya:
@austin987 It's not a super elegant solution to delete an existing cached **d9vk master** archive - when forced updating this verb. But I expect it will be "good enough" for the use case of most **D9VK** users. :-)

austin987:
Hi @bobwya,

You've got an extraneous commit. Otherwise, the cache change is fine (and along the same lines of what I was considering). I'm not as convinced about the vulkan version change, particularly depending on an unreleased wine version that may or may not have what is needed..

bobwya:
@austin987 

_"extraneous commit"_ - _que_? There is an actual **D9VK 0.13f** (frog release) - if that's what you are referring to?!! :-)

How about if leave the Wine version at >=**4.5** and just update the Vulkan version? Obviously the Wine version can be updated, at a later date.

I think the Vulkan version should track the headers that **DXVK**, and by association **D9VK**, are pulling in. Especially since **DXVK** **1.3** will use the new **VK_EXT_shader_demote_to_helper_invocation** Vulkan extension, if it is available.

bobwya:
@austin987 I've cleaned the **d9vk** branch up re: your comments.

src/winetricks
```diff
@@ -6856,6 +6856,10 @@ w_metadata d9vk_master dlls \
 load_d9vk_master()
 {
     # https://git.froggi.es/joshua/d9vk
+    # FIXME: Delete cached file, when verb is forced. Gitlab artifacts do not supply any version/commit hash information.
+    if test "$WINETRICKS_FORCE" == 1 && test -f "${W_CACHE}/${W_PACKAGE}/${file1}"; then
+        rm "${W_CACHE}/${W_PACKAGE}/${file1}"
```
austin987: Please wrap this in w_try (and/or if we really don't care, use -f).

bobwya:
@austin987 Updated, as requested...

austin987:
Thanks

PR #1295, opened 2019-08-14
New verb for MS Money 2005 UK

esanfo99:
Mostly from workarounds in comments on Wine bugs in MS Money 2005 UK.

### DIFF
--- a/src/winetricks
+++ b/src/winetricks
@@ -14024,6 +14024,124 @@ load_mpc()
 
 #----------------------------------------------------------------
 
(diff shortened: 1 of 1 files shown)


austin987:
Thanks for contributing. I'd rather not add new applications to winetricks unless there's a strong demand. In particular, I don't think a 15 year old program with a several year old version of wine qualifies.

I'd encourage you to maintain it as a .verb file, which you can load with winetricks to take advantage of the various winetricks public functions (e.g., w_download, w_try, etc.)

PR #1297, opened 2019-08-18
dotnet30sp1: use native prntvpt

austin987:
@zfigura what do you think of this as a workaround for https://bugs.winehq.org/show_bug.cgi?id=47436 ?

### DIFF
--- a/src/winetricks
+++ b/src/winetricks
@@ -8199,6 +8199,10 @@ load_dotnet30sp1()
     w_call dotnet20sp1
     w_wineserver -w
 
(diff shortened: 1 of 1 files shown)


zfigura:
As far as .NET goes, yeah, that seems reasonable. In that case it's possible we can also just get away with not installing XPSEP.

I've mostly been worried about the fact that XPSEP doesn't install correctly. There must be a regression there, but I haven't yet been able to track it down.

austin987:
https://github.com/Winetricks/winetricks/commit/00db4e9c1bc1a2c4ac85e5c227643f2b0f336406

Issue #1291, opened 2019-08-08
Cannot install dotnet anymore


Evernow:
Hello there!

Unfortunately I am unable to install dotnet anymore, I used to be able to but I can no more, this is what I get when I try, it also happens with winetricks unfortunately.

https://gist.github.com/Evernow/9a76e74288a4bb71f87507eec3d45b00

Any help is greatly appreciated

edintomato:
+1 to this, although I might have found a quirk. Specifying dotnet40 by itself seems to work, but trying to install dotnet472 gives me the same errors as Evernow. What's really strange is that dotnet40 complains that Mono is not installed, but still installs itself anyways.
I'd attach a log, but my terminal only scrolls back so far. Here's what I have left: https://gist.github.com/edintomato/c768f0707b72c5bb6b1b8ecaa30ca04a

bobwya:
@edintomato @Evernow 

Please can you post the output from:

    WINEDEBUG=-all wine uninstaller --list

Ta

edintomato:
@bobwya [Here's the output from that command.](https://gist.github.com/edintomato/b0a03a740785672dd1f580d8a7db2757)

austin987:
That output is after installing dotnet, I believe @bobwya wanted the output from before .net was installed.

In any case, could you try a clean prefix and latest winetricks, https://github.com/Winetricks/winetricks/commit/710c154225358e10964968ddbae7dd9679f8cb2f, will hopefully help.

Issue #1290, opened 2019-08-08
[Fedora 30] Can't Install dotnet 46


ArmagdoGaming:
Using Fedora 30.
I tried installing Vortex via Protontricks GUI. It started by installing dotnet 4.0 and it says that a version of dotnet 4.0 or higher is installed, 4.5 gives an Invalid Function error. It tries to install 4.6 and it just fails. No error messages. Nothing. I tried 

> winetricks dotnet46 --force

 but got no results. Just an error.
Mono isn't installed.
![Screenshot from 2019-08-08 03-00-14](https://user-images.githubusercontent.com/52860974/62667931-1c201200-b98a-11e9-832d-8e5b94b45a45.png)
![Screenshot from 2019-08-08 03-05-39](https://user-images.githubusercontent.com/52860974/62667932-1c201200-b98a-11e9-97ed-c5da45b9631f.png)
![Screenshot from 2019-08-08 03-06-00](https://user-images.githubusercontent.com/52860974/62667933-1cb8a880-b98a-11e9-8501-38eb39fd8a48.png)
![Screenshot from 2019-08-08 03-06-08](https://user-images.githubusercontent.com/52860974/62667934-1cb8a880-b98a-11e9-910a-1ad17aa98918.png)

bobwya:
@ArmagdoGaming 

What's the output from:

    WINEDEBUG=-all wine uninstaller --list

ArmagdoGaming:
> @ArmagdoGaming
> 
> What's the output from:
> 
> ```
> WINEDEBUG=-all wine uninstaller --list
> ```

`{BB9BDB97-C247-4F20-B710-8B6765F67105}|||Wine Gecko (32-bit)
{3AD83002-EEB7-4EB9-BDCB-A533EE281826}|||Wine Gecko (64-bit)
{A183C2D2-A80E-5701-9A35-36B46C1BEA4D}|||Wine Mono Windows Support
`

bobwya:
@ArmagdoGaming 

As I suspected... Awesome thanks!

austin987:
This should be fixed by https://github.com/Winetricks/winetricks/commit/710c154225358e10964968ddbae7dd9679f8cb2f

PR #1298, opened 2019-08-18
remove_mono: fix verb for Wine 4.6(+)

bobwya:
* fixes: https://github.com/Winetricks/winetricks/issues/1290
 * fixes: https://github.com/Winetricks/winetricks/issues/1291

### DIFF
--- a/src/winetricks
+++ b/src/winetricks
@@ -21092,8 +21092,11 @@ w_metadata remove_mono settings \
 
 load_remove_mono()
 {
-    # wine-4.6 comes with two installers, removing "Wine Mono Runtime" will also remove "Wine Mono Windows Support"
-    mono_uuid="$("${WINE_ARCH}" uninstaller --list | grep 'Wine Mono' | grep -v 'Wine Mono Windows Support' | cut -f1 -d\|)"
+    # wine-4.6(+) comes with two alternative installers:
+    #  "Wine Mono" is the older version (msi installer)
+    #  "Wine Mono Windows Support" is the newer alternative (tarball)
+    # We need to test and remove both variants
+    mono_uuid="$("${WINE_ARCH}" uninstaller --list | awk -F'|' '{ if ($NF ~ "^Wine Mono(| Windows Support)") print $1 }' )"
     if test "$mono_uuid"; then
         for uuid in $mono_uuid;
 	    do "${WINE_ARCH}" uninstaller --remove "$uuid"


bobwya:
@austin987 

The new packaging installs the new variant:
 * **Wine Mono Windows Support**

The old packaging installs the old variant:
 * **Wine Mono**

There isn't 2 packages install for a single **wine-mono** version. It's an either/ or situation.

Bob

austin987:
It would be a lot simpler to just remove the `grep -v` bit; i.e.,:

```
diff --git a/src/winetricks b/src/winetricks
index 4caa83e..ee4418b 100755
--- a/src/winetricks
+++ b/src/winetricks
@@ -21095,7 +21095,7 @@ w_metadata remove_mono settings \
 load_remove_mono()
 {
     # wine-4.6 comes with two installers, removing "Wine Mono Runtime" will also remove "Wine Mono Windows Support"
-    mono_uuid="$("${WINE_ARCH}" uninstaller --list | grep 'Wine Mono' | grep -v 'Wine Mono Windows Support' | cut -f1 -d\|)"
+    mono_uuid="$("${WINE_ARCH}" uninstaller --list | grep 'Wine Mono' | cut -f1 -d\|)"
     if test "$mono_uuid"; then
         for uuid in $mono_uuid;
            do "${WINE_ARCH}" uninstaller --remove "$uuid"
```

austin987:
Though it's curious why this doesn't work as is, it does for me. Mono installs two MSIs:
```
austin@laptop ~ $ wine uninstaller --list
{BB9BDB97-C247-4F20-B710-8B6765F67105}|||Wine Gecko (32-bit)
{95E1A5A7-9C47-5B8F-94C7-03C26A544072}|||Wine Mono Runtime
{C1E530E6-B87C-5982-8173-5417B92E5F56}|||Wine Mono Windows Support

austin@laptop ~ $ wine --version
wine-4.14
```

The current script ignores the windows support uuid, so only Wine Mono Runtime is uninstalled. Interestingly, uninstaller prints a warning about the second uuid:
`uninstaller: The application with GUID '{C1E530E6-B87C-5982-8173-5417B92E5F56}' was not found`

In any case, at the end, no mono is installed:
```
austin@laptop ~ $ wine uninstaller --list
{BB9BDB97-C247-4F20-B710-8B6765F67105}|||Wine Gecko (32-bit)
```

bobwya:
Hmm, interesting. When I unpack the newer **wine-mono 4.9.2** binary from: [wine-mono-bin-4.9.2.tar.gz](http://dl.winehq.org/wine/wine-mono/4.9.2/wine-mono-bin-4.9.2.tar.gz)
under: 

    /usr/share/wine/mono/wine-mono-4.9.2/

Then when I boot a clean **32-bit WINEPREFIX**, I get:

    WINEDEBUG=-all wine uninstaller --list
    {BB9BDB97-C247-4F20-B710-8B6765F67105}|||Wine Gecko (32-bit)
    {C1E530E6-B87C-5982-8173-5417B92E5F56}|||Wine Mono Windows Support

    wine --version
    wine-4.14

I tested manually running the **msi** installer at:

    cd /usr/share/wine/mono/wine-mono-4.9.2/support/
    wine msiexec /i winemono-support.msi

with the same result. Obviously I uninstalled any pre-existing wine-mono packages, installed in the test WINEPREFIX.

This appears to match up with other folks using the newer wine-mono packaging...
So I'm not seeing the second package. What steps are you using to install the new wine-mono package?

austin987:
@bobwya I'm not using a prepacked version; I'm building/installing Wine with:
`./configure && make -j$(nproc) && sudo make install`

so the binaries are in /usr/local. When I run `wineboot`, I'm prompted to download the installers, which then end up in ~/.cache/wine.

I looked into it, it seems this behavior was introduced in wine-mono-4.8.1 (wine wine-4.6). Looks like this is intended, a shared install will only have 'Wine Mono Windows Support', while the msi installer will also install 'Wine Mono Runtime'. See https://github.com/madewokherd/wine-mono/releases/tag/wine-mono-4.8.1

I've committed my fix here: https://github.com/Winetricks/winetricks/commit/710c154225358e10964968ddbae7dd9679f8cb2f

I appreciate your attempt, but I think this approach is simpler and more future proof (at least until uninstaller fails for unknown installers).

PR #1304, opened 2019-08-27
tests/self-update-test: use oldest commit instead of previous commit

austin987:
The broken commit didn't fail for me on my pesronal repo when testing; so I'm PRing this to make sure it works for main repo.


austin987:
Didn't, so using a hardcoded version instead.

PR #1325, opened 2019-09-18
w_set_winver: Attempt to improve version uncheck

Kreyren:
Removed w_unset_winver and defined default for w_set_winver

Signed-off-by: Jacob Hrbek <werifgx@gmail.com>

### DIFF
--- a/src/winetricks
+++ b/src/winetricks
@@ -2399,7 +2399,7 @@ _EOF_
             csdversion_hex=dword:00000200
             "$WINE" reg add "HKLM\\System\\CurrentControlSet\\Control\\ProductOptions" /v ProductType /d "WinNT" /f
             ;;
-        win7)
+        win7|default)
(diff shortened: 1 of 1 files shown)


austin987:
Adding a default is a good idea, but I don't think removing `w_unset_winver()` is a good idea; it's a wider change and would break private verbs that use it. Instead have it call `w_set_winver default`, and remove the fixme.

Kreyren:
@austin987 Assuming that 'private verbs' are depending on the `w_unset_winver()` that is expected to set default wine environment, then i'm proposing this solution:

```bash
w_unset_winver()
{
  w_warn 'w_unset_winver() is deprecated, use 'w_set_winver default' instead'
  w_set_winver default
}
```

austin987:
Thanks, that's what I had in mind.

PR #1326, opened 2019-09-18
winetricks: Attempt to improve compactness

Kreyren:
This shrinks whole script by 663 lines without any negative effect

Signed-off-by: Jacob Hrbek <werifgx@gmail.com>

### DIFF
--- a/src/winetricks
+++ b/src/winetricks
@@ -182,8 +182,7 @@ fi
 #---- Public Functions ----
 
 # Ask permission to continue
-w_askpermission()
-{
+w_askpermission() {
     echo "------------------------------------------------------"
     echo "$@"
     echo "------------------------------------------------------"
@@ -219,8 +218,7 @@ w_askpermission()
 }
 
 # Display info message.  Time out quickly if user doesn't click.
-w_info()
-{
+w_info() {
     # If $WINETRICKS_SUPER_QUIET is set, w_info is a no-op:
     if [ -z "$WINETRICKS_SUPER_QUIET" ] ; then
         echo "------------------------------------------------------"
@@ -240,8 +238,7 @@ w_info()
 }
 
 # Display warning message to stderr (since it is called inside redirected code)
-w_warn()
-{
+w_warn() {
     # If $WINETRICKS_SUPER_QUIET is set, w_info is a no-op:
     if [ -z "$WINETRICKS_SUPER_QUIET" ] ; then
         echo "------------------------------------------------------"
@@ -265,8 +262,7 @@ w_warn()
 # Display warning message to stderr (since it is called inside redirected code)
 # And give gui user option to cancel (for when used in a loop)
 # If user cancels, exit status is 1
-w_warn_cancel()
-{
+w_warn_cancel() {
     echo "------------------------------------------------------" >&2
     echo "$@" >&2
     echo "------------------------------------------------------" >&2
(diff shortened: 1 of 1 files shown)


Chiitoo:
I dislike this change greatly, as it makes things less readable.  It's not like I'm even contributing to the code though, so... just throwing the opinion out there.  :]

Kreyren:
> I dislike this change greatly, as it makes things less readable. It's not like I'm even contributing to the code though, so... just throwing the opinion out there. :]

```bash
# Proposed
function() {
  return 0
}
```

Compared to

```bash
# Current
function() 
{
  return 0
}
```

From my point of view:

proposed makes is more readable since there are no unnecessary spaces -> you can fit more code in one screen.

Or what is your argument?

Chiitoo:
I simply find code more readable when the opening and closing braces are aligned (similarly to what is often referred to as the "Allman style" [1]).  It's most helpful (to me at least) with larger blocks of code, and especially when there are several levels of nested 'if' statements.

The more things get clumped together, the more difficult it makes things to read to me, and I might use a fair bit more whitespace in general than most do due to that.

It's just my opinion and how I feel, though.  In no way it's going to be a blocker or anything for this, and I'm not exactly arguing which is better overall.  Rather I just wanted to point out that I like the old style.  If you and austin987 feel it's a change for the better, that's already 2 vs 1, and majority probably wins (that, and I trust austin987's decision-making quite a bit).

1. https://en.wikipedia.org/wiki/Indentation_style#Allman_style

Kreyren:
@Chiitoo My opinion is exact opposite meaning that i believe that using `K&R (aka 1TBS) and variants: Stroustrup, Linux kernel, BSD KNF` based on https://en.wikipedia.org/wiki/Indentation_style#Brace_placement_in_compound_statements feels way more readable and effective to me, but then my codestyle is usually something that none else wants to use alike:

```bash
if ! command -v "git" >/dev/null; then die 1 "command 'git' is not executable"; fi

while [ "$#" -gt 0 ]; do case "$1" in
	-f|--force) forced="KREYREN"; shitf '1';;
	-K|--kill) kill_chroot "$2"; exit '1';;
	*)
		mount_required_fs "$1"
		chroot_in "$1" "$2"
		shift
esac; done
```

While trying to prioritize shell syntax in case said code would be required to use on system without dynamic linking (shell only) 
\+ over-quoting like `exit '1'` to define where first non-executable starts and ends where reasoning being good practice and future proofing
+\  Over-sanitization trying to make `trigger -> action` instead of relying on other codeblocks alike:
```bash
# Sanitized git-clone
egit-clone() {
	# SYNOPSIS: $0 [repository] [path]

	# Sanitization
	if ! command -v "git" >/dev/null; then die 1 "command 'git' is not executable"; fi
	## Sanitization for $1
	[[ "$1" != https://*.git ]] && die 1 "${FUNCNAME[0]}: Argument '$1' doesn't match 'https://*.git'"
	# TODO: Sanitize $2

	[ ! -d "$2" ] && (git clone "$1" "$2" && edebug "${FUNCNAME[0]}: cloned '$1' in '$2'" || die 1 "${FUNCNAME[0]}: Unable to clone '$1' in '$2'") || edebug "${FUNCNAME[0]}: Directory '$2' already exists for '$1', skipping.."
}
```
\+ Trying to avoid user-input when possible alike:
```bash
# FUNCTIONS
checkroot () { # Check if executed as root, if not tries to use sudo
	if ! ((EUID)); then
		return
	elif [[ -x "$(command -v sudo)" ]] && [ -n "$KREYREN_CHECKROOT" ]; then
			info "Failed to aquire root permission, trying reinvoking with 'sudo' prefix"
			exec sudo "$0" "$@"
			exit 1
	elif [[ ! -x "$(command -v sudo)" ]] && [ -n "$KREYREN_CHECKROOT" ]; then
		info "Failed to aquire root permission, trying reinvoking as root user."
		exec su -c "$0 $*"
		exit 1
	else
		die 0 "Unable to elevate root access."
	fi
}
```
etc..

so i woudn't want to force my code style of majority prefers Allman since that seems to be favorite choice in UNIX.

austin987:
FWIW, I haven't looked too deeply yet, but I'm considering using `shfmt`, particularly since that lets us delegate formatting to an external utility without having to shim our own checkers:
https://github.com/mvdan/sh

Issue #1329, opened 2019-09-20
riched30 fetch url is dead


Kreyren:
> Can't access riched30: http://ftp.tw.vim.org/cpatch/msupdate/msi/source/instmsia.exe . It looks like server is not responding

I have same issue on my end

```console
kreyren@dreamon:~$ ping http://ftp.tw.vim.org
ping: http://ftp.tw.vim.org: Name or service not known
```

_Originally posted by @Kreyren in https://github.com/Winetricks/winetricks/pull/1030#issuecomment-533539694_

### Possible solution
Would this be a sufficient replacement?
http://www.bitron.ch/downloads/riched30.exe

---

Referencing: https://github.com/Winetricks/winetricks/pull/1030/files#diff-90b1fb64c3f0a6c10d15531fb2008296R9207

Kreyren:
ping @realgsong

Kreyren:
Assuming that we need `riched20.dll` and `msls31.dll` we could get them from https://www.dll-files.com/

austin987:
Fixed by https://github.com/Winetricks/winetricks/commit/5c99449fc6810ee0dedb5c7f06d0a7b10f009266

PR #1330, opened 2019-09-20
riched30: Fix URL using webarchive

Kreyren:
Fixes: https://github.com/Winetricks/winetricks/issues/1329

Signed-off-by: Jacob Hrbek <weriffx@gmail.com>


Kreyren:
Disclaimer: It's fetched from archived version of the website, i'm not sure how to verify if it's the correct file since i believe that i never used `riched30` before.

austin987:
Given that the sha256sum changes, I think it makes more sense to use archive.org to get what was on microsoft.com. I did that in https://github.com/Winetricks/winetricks/commit/5c99449fc6810ee0dedb5c7f06d0a7b10f009266

PR #1332, opened 2019-09-22
w_uplay_install_game: init

Kreyren:
Attempt to add backend for uplay, tested locally

Depends on: https://github.com/Winetricks/winetricks/pull/1249


Kreyren:
Seems that winetricks exits prematurely, but the wrapper still works

PR #1334, opened 2019-09-22
splintercell_blacklist_uplay: init

Kreyren:
Compatibility for 'Tom Clancy's Splinter Cell: Blacklist' for uplay

Tested using commits below

DEPENDS: https://github.com/Winetricks/winetricks/pull/1332
DEPENDS: https://github.com/Winetricks/winetricks/pull/1249

WARN: Game crashes on dx9 mode on environment loading 

### DIFF
--- a/src/winetricks
+++ b/src/winetricks
@@ -20182,15 +20182,24 @@ load_wog()
 #----------------------------------------------------------------
 
 w_metadata splintercell_blacklist_uplay games \
-    title="Tom Clancy Splinter Cell: Blacklist (Uplay)" \
+    title="Tom Clancy Splinter Cell: Blacklist (Non-free, Uplay)" \
     publisher="Ubisoft" \
     year="2013" \
     media="download" \
-    installed_exe1="${W_PROGRAMS_X86_WIN}/Ubisoft/Ubisoft Game Launcher/games/Splinter Cell Blacklist/Blacklist_Launcher.exe"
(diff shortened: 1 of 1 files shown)


Kreyren:
GAMEDIR/src/SYSTEM/Blacklist.ini has
```
[InitVideo]
Video[0]=Logo_Ubisoft.bik
```

Which can be overwritten to avoid wasting time on logos

Kreyren:
`-offline` is valid argument for uplay games

Kreyren:
GAMEDIR/drive_c/Program\ Files\ \(x86\)/Ubisoft/Ubisoft\ Game\ Launcher/cache/configuration/configuration has instructions for uplay to launch games.

Kreyren:
`GAMEDIR/src/SYSTEM/Blacklist_DX11_game.exe` Has executable for DX11 in case you do the same mistake as me and set the game in DX9 which prevents it from launching

Kreyren:
Seems that winetricks exits prematurely, but the wrapper still works

Kreyren:
Using `ed` instead of `sed`

```
<greybot> Some sed(1)s have -i which claims to "modify" files. It does not: sed is not a FILE editor. The -i flag re-writes the entire file and replaces the original with the new. This breaks open handles and hard-link sets, and fails to follow symlinks. -i is also unportable: valid sed on one system will write broken files on another. Use ed(1) or ex(1) instead: eg. ex -sc '%s/a/b/|wq' file
```

Kreyren:
Game seems to have issues on dxvk on some missions.

TODO: File a dxvk-bug

PR #1337, opened 2019-09-24
w_workaround_wine_bug: Improved Windows Detection

Kreyren:
Current is checking weather WINE variable is set and this method makes it more reliable to be detected on WSL assuming that `uname` on WSL returns `GNU/Linux` and that POSIX sh is concern

### DIFF
--- a/src/winetricks
+++ b/src/winetricks
@@ -2546,10 +2546,12 @@ w_wine_version_in()
 # the environment variable WINETRICKS_BLACKLIST to disable it.
 w_workaround_wine_bug()
 {
-    if test "$WINE" = ""; then
-        echo "No need to work around wine bug $1 on Windows"
+    case "$(uname -r)" in
+      *-Microsoft)
+        w_info "No need to work around wine bug $1 on Windows"
         return 1
-    fi
+    esac
+
     case "$2" in
         [0-9]*) w_die "bug: want message in w_workaround_wine_bug arg 2, got $2" ;;
         "") _W_msg="";;


Kreyren:
Note that output of `uname -r` on WSL is `4.4.0-18362-Microsoft`

austin987:
That would break things under cygwin.

Does Wine even work under WSL?

Kreyren:
> That would break things under cygwin.

Elaborate? 

> Does Wine even work under WSL?

I don't see a reason for it to not work, but from what i was told then the display is an issue on WSL

Kreyren:
If display is an issue i believe that i could make it work using few workarounds and provided feature already has logic for windows so i just tried to improve it so that it woudn't false trigger on linux since checking for WINE variable seems insane to me.

austin987:
Cygwin doesn't have the same uname as WSL.

I don't see the point of this 'improvement', in particular since it's for a theoretical usecase that you haven't actually tested.

Checking for WINE is completely valid, if the user doesn't set it explicitly, winetricks does. See:
https://github.com/Winetricks/winetricks/blob/master/src/winetricks#L5095
https://github.com/Winetricks/winetricks/blob/master/src/winetricks#L3028

if anything, you would want to extend winetricks_get_platform() to detect WSL as windows. But again, I don't see the point unless it actually works on WSL and has been tested. Cygwin is barely tested as it is.

Kreyren:
> Cygwin doesn't have the same uname as WSL.

Confirmed, output of `uname -r` on Cygwin is `3.0.7(0.338/5/3)`, Attempt to adapt logic will be made.

> I don't see the point of this 'improvement', in particular since it's for a theoretical usecase that you haven't actually tested.

Provided logic was tested on WSL where it works as expected. 

Note that i'm not making blind commits unless stated otherwise in commit message. 

winetricks_get_platform() is noted

PR #1318, opened 2019-09-14
Origin (DDP): init

Kreyren:
This attempts to make origin (digital distribution platform made by Electronic Arts) working on wine <4.15

NOTICE: Qt5Svg.dll is required for C:\\Program Files (x86)\\Origin\\imageformats\\qsvg.dll which is non-fatal for non-legacy launcher

NOTICE: Non-legacy is unable to connect to the internet -> Login is impossible and code is left as stub

References
Winebug for installer for winvista+ environment - https://bugs.winehq.org/show_bug.cgi?id=47772 
Winebug for origin-legacy OriginClientService.exe crashing with DXVK - https://bugs.winehq.org/show_bug.cgi?id=47773

Signed-off-by: Jacob Hrbek <werifgx@gmail.com>

PR #1338, opened 2019-09-24
winetricks_get_platform: Attempt to improve platform detection

Kreyren:
Attempts to improve winetricks_get_platform() to allow differenciating
of different platforms

Follows up on: #1337

---

#### BLOCKER 01
requesting reasoning for WINELOADERNOEXEC for windows-based currently
not used in logic

self-diag: Current commit shouldn't require it

#### BLOCKER 02
Is there a reason to provide logic for wineconsole which is running cmd?
Afaik there is no way to get this output in shell unless something like
`wine cmd ECHO %OS%` is used?

wine cmd:
```console
Z:\home\kreyren>ECHO "%OS%"
"Windows_NT"
```

Logic adapted since this seems to be used in logic

#### BLOCKER 03
TODO: Adapt whole script for this logic once returns are agreed on

Update: Resolved (checked four times)

---

### DIFF
--- a/src/winetricks
+++ b/src/winetricks
@@ -532,8 +532,11 @@ w_try_regedit32()
 {
     # on windows, doesn't work without cmd /c
     case "$W_PLATFORM" in
-        windows_cmd|wine_cmd) cmdc="cmd /c";;
-        *) unset cmdc ;;
+        windows_cygwin|wine_cmd) cmdc="cmd /c" ;;
+        nix_linux|nix_freebsd|nix_darwin) unset cmdc ;;
+        nix_redox) w_die "Platform '$W_PLATFORM' is not implemented in w_try_regedit32()" ;;
+        windows_windows) w_die "Platform '$W_PLATFORM' is not implemented in w_try_regedit32()" ;;
+        *) w_die "Unknown platform '$W_PLATFORM' parsed in w_try_regedit32()"
     esac
 
     # shellcheck disable=SC2086
@@ -544,8 +547,11 @@ w_try_regedit64()
 {
     # on windows, doesn't work without cmd /c
     case "$W_PLATFORM" in
-        windows_cmd|wine_cmd) cmdc="cmd /c";;
-        *) unset cmdc ;;
+        windows_cygwin|wine_cmd) cmdc="cmd /c" ;;
+        nix_linux|nix_freebsd|nix_darwin) unset cmdc ;;
+        nix_redox) w_die "Platform '$W_PLATFORM' is not implemented in w_try_regedit64()" ;;
+        windows_windows) w_die "Platform '$W_PLATFORM' is not implemented in w_try_regedit64()" ;;
+        *) w_die "Unknown platform '$W_PLATFORM' parsed in w_try_regedit64()"
     esac
 
     # shellcheck disable=SC2086
@@ -699,16 +705,19 @@ winetricks_wintounix()
 w_pathconv()
 {
     case "$W_PLATFORM" in
-        windows_cmd)
+        windows_cygwin)
             # for some reason, cygpath turns some spaces into newlines?!
             cygpath "$@" | tr '\012' '\040' | sed 's/ $//'
             ;;
-        *)
+        windows_windows) w_die "FIXME: w_pathconv for $1 not implemented" ;;
+        nix_redox) w_die "FIXME: w_pathconv for $1 not implemented" ;;
+        nix_linux|nix_freebsd|nix_darwin)
             case "$@" in
                 -u?c:\\*|-u?C:\\*|-u?c:/*|-u?C:/*) winetricks_wintounix "$2" ;;
                 *) winetricks_early_wine winepath "$@" ;;
             esac
         ;;
+        *) w_die "Unknown platform in w_pathconv() : $W_PLATFORM"
     esac
 }
 
@@ -1760,7 +1769,7 @@ _EOF_
 w_skip_windows()
 {
     case "$W_PLATFORM" in
-        windows_cmd)
+        windows_cygwin|windows_windows)
             echo "Skipping operation '$1' on Windows"
             return 0
             ;;
@@ -2546,10 +2555,10 @@ w_wine_version_in()
 # the environment variable WINETRICKS_BLACKLIST to disable it.
 w_workaround_wine_bug()
 {
-    if test "$WINE" = ""; then
-        echo "No need to work around wine bug $1 on Windows"
-        return 1
-    fi
+    case "$W_PLATFORM" in
+        windows_*) echo "No need to work around wine bug $1 on Windows" ; return 1
+    esac
+
     case "$2" in
         [0-9]*) w_die "bug: want message in w_workaround_wine_bug arg 2, got $2" ;;
         "") _W_msg="";;
@@ -2637,13 +2646,15 @@ w_metadata()
 
     # Do sanity check unless running on Cygwin, where it's way too slow.
     case "$W_PLATFORM" in
-        windows_cmd)
-            ;;
-        *)
+        windows_cygwin) w_warn "FIXME: Sanity check for w_metadata() is too slow, skipping.." ;;
+        windows_windows) w_die "FIXME: Platform $W_PLATFORM is not implemtented in w_metadata()" ;;
+        nix_redox) w_die "FIXME: Platform $W_PLATFORM is not implemtented in w_metadata()" ;;
+        nix_linux|nix_freebsd|nix_darwin)
             if grep '[^"]$' "$file"; then
                 w_die "bug: w_metadata $_W_md_cmd corrupt, might need forward slashes?"
             fi
-            ;;
+        ;;
+        *) w_die "Unknown W_PLATFORM '$W_PLATFORM' detected in w_metadata()"
     esac
     unset _W_md_cmd
 
@@ -2729,8 +2740,8 @@ w_do_call()
 
         # If needed, set the app's wineprefix
         case "$W_PLATFORM" in
-            windows_cmd|wine_cmd) ;;
-            *)
+            windows_cygwin|windows_windows|wine_cmd) ;;
+            nix_linux|nix_freebsd|nix_darwin)
                 # shellcheck disable=SC2154
                 case "${category}-${WINETRICKS_OPT_SHAREDPREFIX}" in
                     apps-0|benchmarks-0|games-0)
@@ -2742,6 +2753,8 @@ w_do_call()
                         ;;
                 esac
             ;;
+            nix_redox) w_die "FIXME: Platform '$W_PLATFORM' is not implemented in w_do_call()" ;;
+            *) w_die "Unknown platform detected in w_do_call() : $W_PLATFORM"
         esac
 
         test "$W_OPT_NOCLEAN" = 1 || rm -rf "$W_TMP"
@@ -3025,20 +3038,66 @@ winetricks_get_sha256sum_prog() {
     fi
 }
 
+# Get running platform in W_PLATFORM and return it in a form of [nix|windows]_[kernel_used]
 winetricks_get_platform()
 {
-    if [ "${OS}" = "Windows_NT" ]; then
-        if [ -n "${WINELOADERNOEXEC}" ]; then
-            # Windows/Cygwin
-            export W_PLATFORM="windows_cmd"
-        else
-            # wineconsole/cmd under wine
-            export W_PLATFORM="wine_cmd"
-        fi
-    else
-        # Normal Unix shell
-        export W_PLATFORM="wine"
-    fi
+  # Linux - `uname` returns 'GNU/Linux'
+  if [ "$(uname -s)" = "Linux" ]; then
+    export W_PLATFORM="nix_linux"
+    printf "%s\n" "$W_PLATFORM"
+    return 0
+
+  # FreeBSD - `uname` returns 'FreeBSD'
+  elif [ "$(uname -s)" = "FreeBSD" ]; then
+    export W_PLATFORM="nix_freebsd"
+    printf "%s\n" "$W_PLATFORM"
+    return 0
+
+  # Redox - `uname` returns 'Redox'
+  elif [ "$(uname -s)" = "Redox" ]; then
+    export W_PLATFORM="nix_redox"
+    printf "%s\n" "$W_PLATFORM"
+    return 0
+
+  # Cygwin - `uname -o` returns 'Cygwin'
+  # `uname -o` on macOS is invalid, do not output error if invoked (https://github.com/Winetricks/winetricks/pull/1415#issuecomment-549769127)
+  elif [ "$(uname -o 2>/dev/null)" = "Cygwin" ]; then
+    export W_PLATFORM="windows_cygwin"
+    printf "%s\n" "$W_PLATFORM"
+    return 0
+
+  # MacOS - `uname` returns 'Darwin'
+  elif [ "$(uname -s)" = "Darwin" ]; then
+    export W_PLATFORM="nix_darwin"
+    printf "%s\n" "$W_PLATFORM"
+    return 0
+
+  # Windows - Impossible to get?
+  elif [ "$(uname)" = "Windows" ]; then
+    w_warn "How did you execute POSIX shell on native Windows? File an issue for compatibility"
+    export W_PLATFORM="windows_windows"
+    printf "%s\n" "$W_PLATFORM"
+    return 0
+
+  # WineCMD/CMD
+  elif [ "$OS" = "Windows_NT" ] && [ -z "$(uname)" ]; then
+    export W_PLATFORM="wine_cmd"
+    printf "%s\n" "$W_PLATFORM"
+    return 0
+
+  else
+   # WSL - `uname -r` returns `4.4.0-18362-Microsoft` where 4.4.0 is kernel version and 18362 windows build(?)
+   ## $OS is blank on WSL (10/2019)
+   case "$(uname -r)" in
+     *-Microsoft)
+      export W_PLATFORM="windows_wsl"
+      printf "%s\n" "$W_PLATFORM"
+      return 0
+   esac
+  fi
+
+  # Logic-check
+  [ -z "$W_PLATFORM" ] && w_die "Function winetricks_get_platform is unable to get W_PLATFORM variable"
 }
 
 winetricks_latest_version_check()
@@ -4050,7 +4109,7 @@ winetricks_is_installed()
     fi
 
     case "$W_PLATFORM" in
-        windows_cmd|wine_cmd)
+        windows_cygwin|windows_windows|wine_cmd)
             # On Windows, there's no wineprefix, just check if file's there
             _W_file_unix="$(w_pathconv -u "$_W_file")"
             if test -f "$_W_file_unix"; then
@@ -4058,7 +4117,7 @@ winetricks_is_installed()
                 return 0  # installed
             fi
             ;;
-        *)
+        nix_linux|nix_freebsd|nix_darwin)
             # Compute wineprefix for this app
             case "${category}-${WINETRICKS_OPT_SHAREDPREFIX}" in
                 apps-0|benchmarks-0|games-0)
@@ -4085,6 +4144,8 @@ winetricks_is_installed()
              IFS="$_W_IFS"
             fi
             ;;
+        nix_redox) w_die "Platform '$W_PLATFORM' is not implemented in winetricks_is_installed()" ;;
+        *) w_die "Unknown platform '$W_PLATFORM' used in winetricks_is_installed"
     esac
     unset _W_file _W_prefix _W_IFS  # leak _W_file_unix for caller. Is this wise?
     return 1  # not installed
@@ -4812,10 +4873,13 @@ winetricks_set_wineprefix()
     esac
 
     case "$W_PLATFORM" in
-        windows_cmd)
+        windows_cygwin)
             W_DRIVE_C="/cygdrive/c" ;;
-        *)
+        windows_windows) w_die "FIXME: Platform '$W_PLATFORM' is not implemtented in winetricks_set_wineprefix()" ;;
+        nix_redox) w_die "FIXME: Platform '$W_PLATFORM' is not implemtented in winetricks_set_wineprefix()" ;;
+        nix_linux|nix_freebsd|nix_darwin)
             W_DRIVE_C="$WINEPREFIX/dosdevices/c:" ;;
+        *) w_die "Unknown platform '$W_PLATFORM' was used in winetricks_set_wineprefix()"
     esac
 
     # Kludge: use Temp instead of temp to avoid \t expansion in w_try
@@ -4831,8 +4895,8 @@ winetricks_set_wineprefix()
     fi
 
     case "$W_PLATFORM" in
-        "windows_cmd|wine_cmd") W_CACHE_WIN="$(w_pathconv -w "$W_CACHE")" ;;
-        *)
+        windows_cygwin|windows_windows|wine_cmd) W_CACHE_WIN="$(w_pathconv -w "$W_CACHE")" ;;
+        nix_linux|nix_freebsd|nix_darwin)
             # For case where Z: doesn't exist or / is writable (!),
             # make a drive letter for W_CACHE.  Clean it up on exit.
             test "$WINETRICKS_CACHE_SYMLINK" && rm -f "$WINETRICKS_CACHE_SYMLINK"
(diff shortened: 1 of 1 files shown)


Kreyren:
ping @austin987 attempting to improve requested function in https://github.com/Winetricks/winetricks/pull/1337#issuecomment-534694058, more info required

Kreyren:
Ready for review

Kreyren:
Reference for logic on Redox https://gitlab.redox-os.org/redox-os/kernel/blob/master/src/scheme/sys/uname.rs

austin987:
A) fails Travis, so not going in
B) I'm unclear on what this is needed for. I.e., what bug does this fix?

src/winetricks
```diff
@@ -2546,10 +2555,10 @@ w_wine_version_in()
 # the environment variable WINETRICKS_BLACKLIST to disable it.
 w_workaround_wine_bug()
 {
-    if test "$WINE" = ""; then
-        echo "No need to work around wine bug $1 on Windows"
-        return 1
-    fi
```
Kreyren: > A) fails Travis, so not going in

Resolved now and ready for review and merge

> B) I'm unclear on what this is needed for. I.e., what bug does this fix?

Referenced line was the main motivation for this commit since checking if WINE variable is zero seems unreliable and insane to me assuming that winetricks can be used on cygwin and WSL1/2 now

This way we have more control over logic per platform since we can reliably detect the platform in question and provide logic e.i: https://github.com/Winetricks/winetricks/pull/1318 which seems to require different workarounds for freebsd, darwin and linux. 

And assuming that these kernels are very different from each other i would expect using this logic to adapt wineapps for them and personally i'm trying to migrate from GNU/Linux on GNU/Redox which is going to need it to make WINE working.

austin987: > Referenced line was the main motivation for this commit since checking if WINE variable is zero seems unreliable and insane to me assuming that winetricks can be used on cygwin and WSL1/2 now

"seems unreliable" isn't a bug. If you've got an actual bug, that's a different story. WINE is unset on windows, so that when a command line "$WINE" foo.exe is executed, it won't try to execute a non-existent binary.

I'm still not convinced winetricks actually works on WSL1/2 (even with this patch). I'd be happy to be proven wrong on that.

src/winetricks
```diff
@@ -532,8 +532,11 @@ w_try_regedit32()
 {
     # on windows, doesn't work without cmd /c
     case "$W_PLATFORM" in
-        windows_cmd|wine_cmd) cmdc="cmd /c";;
-        *) unset cmdc ;;
+        windows_cygwin|wine_cmd) cmdc="cmd /c" ;;
+        nix_linux|nix_freebsd|nix_darwin) unset cmdc ;;
+        nix_redox) w_die "Platform '$W_PLATFORM' is not implemented in w_try_regedit32()" ;;
+        windows_windows) w_die "Platform '$W_PLATFORM' is not implemented in w_try_regedit32()" ;;
```
austin987: Why on earth are you dying here? Windows works just fine as is.

src/winetricks
```diff
@@ -532,8 +532,11 @@ w_try_regedit32()
 {
     # on windows, doesn't work without cmd /c
     case "$W_PLATFORM" in
-        windows_cmd|wine_cmd) cmdc="cmd /c";;
-        *) unset cmdc ;;
+        windows_cygwin|wine_cmd) cmdc="cmd /c" ;;
+        nix_linux|nix_freebsd|nix_darwin) unset cmdc ;;
+        nix_redox) w_die "Platform '$W_PLATFORM' is not implemented in w_try_regedit32()" ;;
```
austin987: If a platform legitimately doesn't have regedit, it's completely broken and there's no reason for adding 'support' for it.

src/winetricks
```diff
@@ -3025,20 +3038,65 @@ winetricks_get_sha256sum_prog() {
...
+    export W_PLATFORM="nix_darwin"
+    printf "%s\n" "$W_PLATFORM"
+    return 0
+
+  # Windows - Impossible to get?
+  elif [ "$(uname)" = "Windows" ]; then
+    w_warn "How did you execute POSIX shell on native Windows? File an issue for compatibility"
+    export W_PLATFORM="windows_windows"
+    printf "%s\n" "$W_PLATFORM"
+    return 0
```
austin987: If this is impossible, why add it and clutter the code elsewhere?

src/winetricks
```diff
@@ -3025,20 +3038,65 @@ winetricks_get_sha256sum_prog() {
...
+
+  # WineCMD/CMD
+  elif [ "$OS" = "Windows_NT" ] && [ -z "$(uname)" ]; then
+    export W_PLATFORM="wine_cmd"
+    printf "%s\n" "$W_PLATFORM"
+    return 0
+
+  else
```
austin987: If you're worried about unknown platforms, make this an elif, then add an else afterward that sets W_PLATFORM=unknown.

src/winetricks
```diff
@@ -699,16 +705,19 @@ winetricks_wintounix()
...
 {
     case "$W_PLATFORM" in
-        windows_cmd)
+        windows_cygwin)
             # for some reason, cygpath turns some spaces into newlines?!
             cygpath "$@" | tr '\012' '\040' | sed 's/ $//'
             ;;
-        *)
+        windows_windows) w_die "FIXME: w_pathconv for $1 not implemented" ;;
+        nix_redox) w_die "FIXME: w_pathconv for $1 not implemented" ;;
```
austin987: Not needed.

src/winetricks
```diff
@@ -2546,10 +2555,10 @@ w_wine_version_in()
 # the environment variable WINETRICKS_BLACKLIST to disable it.
 w_workaround_wine_bug()
 {
-    if test "$WINE" = ""; then
-        echo "No need to work around wine bug $1 on Windows"
-        return 1
-    fi
+    case "$(winetricks_get_platform)" in
```
austin987: $W_PLATFORM

src/winetricks
```diff
@@ -2637,13 +2646,15 @@ w_metadata()
 
     # Do sanity check unless running on Cygwin, where it's way too slow.
     case "$W_PLATFORM" in
-        windows_cmd)
-            ;;
-        *)
+        windows_cygwin) w_warn "FIXME: Sanity check for w_metadata() is too slow, skipping.." ;;
+        windows_windows) w_die "FIXME: Platform $W_PLATFORM is not implemtented in w_metadata()" ;;
+        nix_redox) w_die "FIXME: Platform $W_PLATFORM is not implemtented in w_metadata()" ;;
```
austin987: Not needed.

src/winetricks
```diff
@@ -2637,13 +2646,15 @@ w_metadata()
 
     # Do sanity check unless running on Cygwin, where it's way too slow.
     case "$W_PLATFORM" in
-        windows_cmd)
-            ;;
-        *)
+        windows_cygwin) w_warn "FIXME: Sanity check for w_metadata() is too slow, skipping.." ;;
+        windows_windows) w_die "FIXME: Platform $W_PLATFORM is not implemtented in w_metadata()" ;;
```
austin987: Not needed.

src/winetricks
```diff
@@ -2742,6 +2753,8 @@ w_do_call()
                         ;;
                 esac
             ;;
+            nix_redox) w_die "FIXME: Platform '$W_PLATFORM' is not implemented in w_do_call()" ;;
```
austin987: Not needed.

src/winetricks
```diff
@@ -3025,20 +3038,65 @@ winetricks_get_sha256sum_prog() {
...
-        else
-            # wineconsole/cmd under wine
-            export W_PLATFORM="wine_cmd"
-        fi
-    else
-        # Normal Unix shell
-        export W_PLATFORM="wine"
-    fi
+  # Linux - `uname` returns 'GNU/Linux'
+  if [ "$(uname)" = "Linux" ]; then
```
austin987: Please use explicit `uname -s`

src/winetricks
```diff
@@ -3025,20 +3038,65 @@ winetricks_get_sha256sum_prog() {
...
+  elif [ "$(uname)" = "FreeBSD" ]; then
+    export W_PLATFORM="nix_freebsd"
+    printf "%s\n" "$W_PLATFORM"
+    return 0
+
+  # Redox - `uname` returns 'Redox'
+  elif [ "$(uname)" = "Redox" ]; then
+    export W_PLATFORM="nix_redox"
+    printf "%s\n" "$W_PLATFORM"
+    return 0
```
austin987: Not needed.

src/winetricks
```diff
@@ -4085,6 +4143,8 @@ winetricks_is_installed()
              IFS="$_W_IFS"
             fi
             ;;
+        nix_redox) w_die "Platform '$W_PLATFORM' is not implemented in winetricks_is_installed()" ;;
```
austin987: Not needed.

src/winetricks
```diff
@@ -4812,10 +4872,13 @@ winetricks_set_wineprefix()
     esac
 
     case "$W_PLATFORM" in
-        windows_cmd)
+        windows_cygwin)
             W_DRIVE_C="/cygdrive/c" ;;
-        *)
+        windows_windows) w_die "FIXME: Platform '$W_PLATFORM' is not implemtented in winetricks_set_wineprefix()" ;;
+        nix_redox) w_die "FIXME: Platform '$W_PLATFORM' is not implemtented in winetricks_set_wineprefix()" ;;
```
austin987: Not needed.

src/winetricks
```diff
@@ -5037,7 +5102,8 @@ winetricks_init()
 
     winetricks_get_sha256sum_prog
 
-    winetricks_get_platform
+    # winetricks_get_platform got updated and in addition to W_PLATFORM export it outputs the platform which seems unexpected here
```
austin987: So remove the printf's you added..

src/winetricks
```diff
@@ -5094,15 +5160,16 @@ winetricks_init()
...
+        windows_cygwin)
             WINE=""
             WINE64=""
             WINE_ARCH=""
             WINE_MULTI=""
             WINESERVER=""
             W_DRIVE_C="C:/"
             ;;
-        *)
+        windows_windows) w_die "Platform '$W_PLATFORM' is not implemented in winetricks_init()" ;;
```
austin987: Not needed.

src/winetricks
```diff
@@ -5154,6 +5221,8 @@ winetricks_init()
                 fi
                 unset _abswine
                 ;;
+        nix_redox) w_die "Platform '$W_PLATFORM' is not implemented in winetricks_init()" ;;
```
austin987: Not needed.

src/winetricks
```diff
@@ -8312,9 +8383,11 @@ load_dotnet35()
         w_package_unsupported_win64
     fi
 
+    # Platform compatibility
     case "$W_PLATFORM" in
-        windows_cmd) ;;
-        *) w_warn "dotnet35 does not yet fully work or install on wine.  Caveat emptor." ;;
+        windows_cygwin|windows_windows) ;;
+        nix_linux|nix_freebsd|nix_darwin|nix_redox) w_warn "dotnet35 does not yet fully work or install on wine.  Caveat emptor." ;;
+        *) w_die "Unknown platform '$W_PLATFORM' parsed in load_dotnet35()"
```
austin987: I highly doubt this is needed, if you're on an unknown platform it would've died already.

src/winetricks
```diff
@@ -8367,8 +8440,9 @@ load_dotnet35sp1()
     fi
 
     case "$W_PLATFORM" in
-        windows_cmd) ;;
-        *) w_warn "dotnet35sp1 does not yet fully work or install on wine.  Caveat emptor." ;;
+        windows_cygwin|windows_windows) ;;
+        nix_linux|nix_freebsd|nix_darwin|nix_redox) w_warn "dotnet35sp1 does not yet fully work or install on wine.  Caveat emptor." ;;
+        *) w_die "Unknown platform parsed in load_dotnet35sp1()"
```
austin987: Again, unnecessary

src/winetricks
```diff
@@ -8416,8 +8490,9 @@ load_dotnet40()
     w_package_warn_win64
 
     case "$W_PLATFORM" in
-        windows_cmd) ;;
-        *) w_warn "dotnet40 does not yet fully work or install on wine.  Caveat emptor." ;;
+        windows_cygwin|windows_windows) ;;
+        nix_linux|nix_freebsd|nix_darwin|nix_redox) w_warn "dotnet40 does not yet fully work or install on wine.  Caveat emptor." ;;
+        *) w_die "Unknown platform '$W_PLATFORM' used in load_dotnet40()"
```
austin987: Again, unnecessary

src/winetricks
```diff
@@ -18550,11 +18625,14 @@ load_masseffect2_demo()
...
             cp "$W_CACHE/$W_PACKAGE/MassEffect2DemoEN.exe" "$W_TMP"
-            chmod +x "$W_TMP"/MassEffect2DemoEN.exe ;;
-        *)
-            ln -sf "$W_CACHE/$W_PACKAGE/MassEffect2DemoEN.exe" "$W_TMP" ;;
+            chmod +x "$W_TMP"/MassEffect2DemoEN.exe
+        ;;
+        windows_windows) w_die "Platform '$W_PLATFORM' is not implemented in load_masseffect2_demo()" ;;
+        nix_linux|nix_freebsd|nix_darwin) ln -sf "$W_CACHE/$W_PACKAGE/MassEffect2DemoEN.exe" "$W_TMP" ;;
+        nix_redox) w_die "Platform '$W_PLATFORM' is not implemented in load_masseffect2_demo()" ;;
+        *) w_die "Unknown platform '$W_PLATFORM' parsed in load_masseffect2_demo()"
```
austin987: Again, unnecessary

src/winetricks
```diff
@@ -21567,8 +21645,11 @@ winetricks_stats_log_command()
 
     # and for the user's own reference later, when figuring out what he did
     case "$W_PLATFORM" in
-        windows_cmd) _W_LOGDIR="$W_WINDIR_UNIX"/Temp ;;
-        *) _W_LOGDIR="$WINEPREFIX" ;;
+        windows_cygwin) _W_LOGDIR="$W_WINDIR_UNIX"/Temp ;;
+        windows_windows) w_die "Platform '$W_PLATFORM' is not implemented in winetricks_stats_log_command()" ;;
+        nix_linux|nix_freebsd|nix_darwin) _W_LOGDIR="$WINEPREFIX" ;;
+        nix_redox) w_die "Platform '$W_PLATFORM' is not implemented in winetricks_stats_log_command()" ;;
+        *) w_die "Unknown platform '$W_PLATFORM' parsed in winetricks_stats_log_command()"
```
austin987: Again, unnecessary

src/winetricks
```diff
@@ -21761,13 +21842,13 @@ then
...
         "")
             if [ -z "$DISPLAY" ]; then
-                if [ "$(uname -s)" = "Darwin" ]; then
-                    echo "Running on OSX, but DISPLAY is not set...probably using Mac Driver."
-                else
-                    echo "DISPLAY not set, not defaulting to gui"
-                    winetricks_usage
-                    exit 0
-                fi
+                case "$(winetricks_get_platform)" in
```
austin987: $W_PLATFORM

src/winetricks
Kreyren: @austin987 I don't have way to test this and currently windows is unable to run shell/bash by itself afaik.

Note that `windows_windows` assumes native implementation where `windows_cygwin` and `windows_wsl` are cygwin and wsl

src/winetricks
Kreyren: Agree, but assuming that i'm rewriting winetricks_get_platform and that i was using `windows_windows` for testing i implemented it too assuming it being efficient on resources and my general ethicacy is to support everything.

src/winetricks
Kreyren: elaborate? 

> windows_windows) w_die "FIXME: w_pathconv for $1 not implemented" ;;
This will trigger only if shell is executed on windows kernel and i was told that there is a 'hacky' way to get shell on native windows + sanitization doesn't hurt i believe

src/winetricks
Kreyren: Presented makes sure that W_PLATFORM is exported assuming that `winetricks_get_platform` now outputs which platform is used or do you want me to use $W_PLATFORM instead?

src/winetricks
Kreyren: @austin987 WSL1 has shell/bash running on what seems to be linux kernel so it's able to execute it for WSL2 don't know since it's not released yet from what i found i believe that WSL1 is something like WINE for linux and WSL2 is direct implementation of linux kernel.

> "seems unreliable" isn't a bug. If you've got an actual bug, that's a different story. 

```bash
load_something() { unset WINE ; w_workaround_wine_bug (windows is false triggered) ; }
```

src/winetricks
austin987: It may not exist, it's not supported, and it clutters up the code.

src/winetricks
Kreyren: > windows_windows) w_die "FIXME: Platform $W_PLATFORM is not implemtented in w_metadata()" ;;

same reasoning as in https://github.com/Winetricks/winetricks/pull/1338/files/7c0eed4e8a25108c4f9e345657356343a07cbf78#r332377677

src/winetricks
austin987: Use $W_PLATFORM

src/winetricks
Kreyren: @austin987 `nix_redox` is not tested, do you want me to assume that it works?

src/winetricks
Kreyren: Same reasoning as https://github.com/Winetricks/winetricks/pull/1338/files/7c0eed4e8a25108c4f9e345657356343a07cbf78#r332380161

src/winetricks
austin987: Yes, but it lacks X support. All it can run is wine cmd and other command line apps.

src/winetricks
austin987: No, I don't want extra platforms added for no reason, particularly if they're untested. There are literally hundreds of possibilities, there's absolutely no reason to detect all of them unless they need special treatment somewhere.

src/winetricks
Kreyren: Needed to allow logic per platform as referenced on https://github.com/Winetricks/winetricks/pull/1318

src/winetricks
Kreyren: @austin987 Since current is using it, but i wasn't able to access these variables from shell/bash unless something like `wineconsole %OS% >> ~/test ; cat ~/test` is used?  -> requesting sanity check

src/winetricks
Kreyren: @austin987 Proposing to w_die if unknown platform is used assuming that not everything in winetricks is sanitized to avoid unexpected results.

src/winetricks
Kreyren: Same reasoning as https://github.com/Winetricks/winetricks/pull/1338/files/7c0eed4e8a25108c4f9e345657356343a07cbf78#r332380161

src/winetricks
Kreyren: Same reasoning as https://github.com/Winetricks/winetricks/pull/1338/files/7c0eed4e8a25108c4f9e345657356343a07cbf78#r332380161 for nix_redox 

Same reasoning as https://github.com/Winetricks/winetricks/pull/1338/files/7c0eed4e8a25108c4f9e345657356343a07cbf78#r332377677 for windows_windows

src/winetricks
Kreyren: i believe that calling the function directly is more sane assuming situation in which W_PLATFORM is not exported.

src/winetricks
Kreyren: Same reasoning as https://github.com/Winetricks/winetricks/pull/1338/files/7c0eed4e8a25108c4f9e345657356343a07cbf78#r332377677 for windows_windows

windows_cygwin used in previous and required to avoid attempting to run WINE.

src/winetricks
Kreyren: Same reasoning as https://github.com/Winetricks/winetricks/pull/1338/files/7c0eed4e8a25108c4f9e345657356343a07cbf78#r332380161

src/winetricks
Kreyren: if the function is called directly it woudn't die + sanitization

src/winetricks
austin987: I don't understand your question.

src/winetricks
austin987: Sure, that's fine.

src/winetricks
Kreyren: Same reasoning as https://github.com/Winetricks/winetricks/pull/1338/files/7c0eed4e8a25108c4f9e345657356343a07cbf78#r332385126 for `*)`

src/winetricks
Kreyren: Same reasoning as https://github.com/Winetricks/winetricks/pull/1338/files/7c0eed4e8a25108c4f9e345657356343a07cbf78#r332385126 for `*)`

src/winetricks
Kreyren: Same reasoning as https://github.com/Winetricks/winetricks/pull/1338/files/7c0eed4e8a25108c4f9e345657356343a07cbf78#r332385126 for `*)`



Same reasoning as https://github.com/Winetricks/winetricks/pull/1338/files/7c0eed4e8a25108c4f9e345657356343a07cbf78#r332380161 for nix_redox

Same reasoning as https://github.com/Winetricks/winetricks/pull/1338/files/7c0eed4e8a25108c4f9e345657356343a07cbf78#r332377677 for windows_windows

src/winetricks
Kreyren: Same reasoning as https://github.com/Winetricks/winetricks/pull/1338/files/7c0eed4e8a25108c4f9e345657356343a07cbf78#r332385126 for `*)`

Same reasoning as https://github.com/Winetricks/winetricks/pull/1338/files/7c0eed4e8a25108c4f9e345657356343a07cbf78#r332380161 for nix_redox

Same reasoning as https://github.com/Winetricks/winetricks/pull/1338/files/7c0eed4e8a25108c4f9e345657356343a07cbf78#r332377677 for windows_windows

src/winetricks
Kreyren: Same reasoning as https://github.com/Winetricks/winetricks/pull/1338/files/7c0eed4e8a25108c4f9e345657356343a07cbf78#r332378232

src/winetricks
austin987: What do you mean called directly?

src/winetricks
Kreyren: Seems to be possible to run shell without subsystem and cygwin: 

Referencing https://stackoverflow.com/a/21580885

Referencing https://youtu.be/0AiqegwnskE?t=107 (http://win-bash.sourceforge.net/)

src/winetricks
Kreyren: @austin987 

```shell
# Assuming W_PLATFORM=""
load_something() { w_try_regedit32 ;} 
```

Script should fail if W_PLATFORM is not exported to prevent unexpected or if something unset W_PLATFORM

austin987: By the time load_something runs, winetricks_init() would already have run.

Rather than hypotheticals, do you have a testcase demonstrating it?

src/winetricks
Kreyren: @austin987 I was told that there are ways to get X support on the WSL1 from quick search: https://www.ctrl.blog/entry/how-to-x-on-wsl.html which seems to export Cygwin X/vcXsrv into WSL1 

image

Assuming that pty/tty is missing on WSL1 i would feel confident in making a franken-wsl and passing the X through SSH or something alike, but i'm not windows user so i won't be testing this.

Aldo there is probably no reason (excluding performance benefit?) to use WINE on windows so i would assume that WSL1 users would use it to install apps using CLI since winetricks has logic for them and is using AHK to make things faster to install.

src/winetricks
Kreyren: winecmd is used in current code and in logic for functions:
image

Kreyren: ah you meant windows_windows?

Same reasoning as https://github.com/Winetricks/winetricks/pull/1338/files/7c0eed4e8a25108c4f9e345657356343a07cbf78#r332377677

src/winetricks
Kreyren: Resolved on demand https://github.com/Winetricks/winetricks/pull/1338/commits/90690b501d6362c95f6645b66f3bc082b93e410c

Disclaimer: i still think that said function should be called directly instead of using variables.

src/winetricks
Kreyren: # Kreytricks rebase
i'm working on https://github.com/Winetricks/winetricks/pull/1359 so that winetricks woudn't make a new wineprefix on init so that https://github.com/Winetricks/winetricks/pull/1348 could be used on https://github.com/Winetricks/winetricks/pull/1347 (and others) without the need to reinvoke winetricks which could conflict with it where W_PLATFORM is not exported.

# Hypothetical
Other then that assuming that it supports WSL1, Native Windows and Cygwin i would assume there being forks that tries to unset this to make it working on their system and so it should encourage contrib

# Practical
image
image

Kreyren:
uname -o is nit valid on macos

https://github.com/Winetricks/winetricks/pull/1415#issuecomment-549769127

PR #1345, opened 2019-09-27
Add Igor Pro

ukos-git:
Igor Pro by WaveMetrics uses QT as GUI. The fonts are required for UI
compatibility. 32bit is recommended for the installer.

Although the program is is mostly compatible to lower wine versions, a recent wine
version > 3 is recommended.

Movie export will not work.


Kreyren:
```
In src/winetricks line 14129:

	w_download http://www.wavemetrics.net/Downloads/Win/${file1}

                                                            ^-- SC2086: Double quote to prevent globbing and word splitting.

In src/winetricks line 14131:

	IGORPRO_LOGFILE=$(mktemp --dry-run --tmpdir ${file1%.*}XXX.log)

                                                    ^-- SC2086: Double quote to prevent globbing and word splitting.
```

Kreyren:
Please follow current commit naming in your case it should be:

```
igopro: new verb

<optionally some message here>

<optionally signature>
```

Referencing example https://github.com/Winetricks/winetricks/commit/f54242371e4917c94bfc01b87ea34727ad9ab09c

Kreyren:
```
In src/winetricks line 14131:

	IGORPRO_LOGFILE="$(mktemp --dry-run --tmpdir ${file1%.*}XXX.log)"

                                                     ^-- SC2086: Double quote to prevent globbing and word splitting.
```

Recommends checking your contribution in https://www.shellcheck.net/ which should output all problems in real time there are also plugins in text editor https://github.com/koalaman/shellcheck#in-your-editor

---

Please squash commits

Kreyren:
on my end this fails to open on 64-bit wineprefix (https://gist.githubusercontent.com/Kreyren/69432ca5ebdbdac7c8f4c7f4b390ddb5/raw/9cbc9824deafa79829139c05a0dc36bc4e8e57c9/sdagasdg)

32-bit wineprefix results in following, but continues to install:
image

log: https://gist.github.com/Kreyren/5a63528002ff3b18f378acf46493d817

Note that on win32 it installs in wine `$WINEPREFIX/drive_c/Program\ Files/WaveMetrics/Igor\ Pro\ 8\ Folder/IgorBinaries_Win32/Igor.exe`

win32 works on my end:
image

Consider forcing win32 using `w_package_unsupported_win64`

src/winetricks
```diff
@@ -14105,6 +14105,37 @@ _EOF_
...
+file1="setupIgor8.exe" \
+installed_exe1="C:/Program Files/WaveMetrics/Igor Pro 8 Folder/IgorBinaries_Win32/Igor.exe"
+
+load_igorpro()
+{
+        if test "$W_ARCH" = "win64"; then
+            w_warn "Installation is recommended to be done on a 32bit wine prefix."
+        fi
+
+	w_set_winver win7
```
Kreyren: Is there any reason for win7? Note that you can use `default` argument to use default wine environment (future-proofing)
https://github.com/Winetricks/winetricks/blob/1a3c72846a63c659b0b98073f201394ae2d08c36/src/winetricks#L2402

Kreyren:
Proposing:
```
load_igorpro()
{
        if test "$W_ARCH" = "win64"; then
            w_warn "Installation is recommended to be done on a 32bit wine prefix."
        fi

	w_call corefonts
	w_call tahoma # menu font
	w_call droid  # replacement for lucida
	w_call fontsmooth=rgb

	w_download "http://www.wavemetrics.net/Downloads/Win/${file1}"

  w_call vcrun2017

  w_set_winver default

	IGORPRO_LOGFILE="$(mktemp --dry-run --tmpdir ${file1%.*}XXX.log)"
	w_try "$WINE" "${W_CACHE}/${W_PACKAGE}/${file1}" /install "/log:$IGORPRO_LOGFILE" $W_UNATTENDED_SLASH_Q
}
```

This seems to avoid error https://github.com/Winetricks/winetricks/pull/1345#issuecomment-535996700 

`w_set_winver default` is used since `vcrun2017` sets the environment on winXP which should probably be fixed.(confirmation needed)

ukos-git:
Thank you for the review.

### architecture

The program itself runs just fine on 32bit and 64bit wine. Forcing to `w_package_unsupported_win64` would not allow to install the 64bit binaries. I just tested the install without `wine32` and it seems to work on ubuntu using wine 4.2. However, I know that there are issues. I tried with a docker container using debian buster and wine-development with `4.2` and it only works on 32bit architecture. 

### vcrun

I cannot remember why I put the `vcrun2017` after the installer. I think, it was messing around with something and as the installer did not directly require it but only starting the program itself requires it, I put it after the install procedure. If the windows version is XP, the installer will not work but win10 is also fine.

Kreyren:
> I tried with a docker container using debian buster and wine-development with 4.2 and it only works on 32bit architecture.

Recommends making a winebug and referencing it in the winetricks using `w_workaround_wine_bug()` so that the installer would fail on debian if 64-bit is used and if bug is still not resolved. (willing to contrib this logic assuming winebug provided since i'm not using this software)

ukos-git:
Just tested with `wine-4.2.r0.gb1b1abaf` and this works.
I cannot help with testing below 4.2 but if I remember correctly, wine 3?? also worked and only wine 2 was a bad choice. Unless there is a test system that will run a verb in different wine versions I think this is ready for merge.

PR #1353, opened 2019-09-30
winetricks: Improved wording in source

Kreyren:
Improved wording to prevent freenode.net/#bash and other scripters from freaking out about 25K+ long POSIX shell script that is a package manager.

image

### DIFF
--- a/src/winetricks
+++ b/src/winetricks
@@ -15,7 +15,7 @@ WINETRICKS_VERSION=20190912-next
 
 #--------------------------------------------------------------------
 #
-# Winetricks is a package manager for Win32 dlls and applications on POSIX.
+# Winetricks is POSIX shell frontend for WINE to configure wineapps and it's components
 # Features:
 # - Consists of a single shell script - no installation required
 # - Downloads packages automatically from original trusted sources


austin987:
A) I don't see this as an improvement. Just because you call a dog a cat doesn't make it a cat.
B) I especially don't see the point if it's to appease some trolls on IRC.

Kreyren:
> I especially don't see the point if it's to appease some trolls on IRC.

Those are legit scripters not trolls and it's quite common that people freak out when they see it assuming that they are not familiar with winetricks. 

> Just because you call a dog a cat doesn't make it a cat.

I don't thinks that this is a case since winetricks is not really a package manager in a sence of linux since it manages only WINE, it's components and wineapps which to me clarifies as presented.

Chiitoo:
To quote Wikipedia a little: "A package manager deals with packages, distributions of software and data in archive files." and "A package manager or package-management system is a collection of software tools that automates the process of installing, upgrading, configuring, and removing computer programs for a computer's operating system in a consistent manner."

I guess it more or less fits, aside from perhaps the 'removing' part.

I'm not sure if the change is an improvement either, or if there needs to be any in the first place, but I will say that the official spelling of Wine is Wine, not WINE.  :]

Kreyren:
@Chiitoo alright, noted

PR #1370, opened 2019-10-11
VulkanSDK: Attempt to resolve #1368

### LINKED ISSUE #1368: Vulkan Runtime changes that will require winetricks modifications
I exchanged email offline with @austin987 but wanted to document here in the issue tracker to make sure this doesn't get lost.

In the future, the Vulkan SDK for windows as downloaded from vulkan.lunarg.com will no longer include the Vulkan Runtime Installer (which was used to install the Vulkan Loader on the system). The reason for this is that the Vulkan Loader is included with IHV driver packages on Windows and is the recommended method to get the Vulkan Loader. 

Although the Vulkan Loader install will no longer be part of the Windows SDK, a separate Vulkan Runtime installer package (for the latest version only) will still be made available on vulkan.lunarg.com for solutions such as Wine. A permament link already exists for this package:  https://vulkan.lunarg.com/sdk/home#sdk/downloadConfirm/latest/windows/vulkan-runtime.exe

Winetricks is currently hard wired to use an older version of the Vulkan SDK to get the loader (by installing an SDK). Should Winetricks want to move ahead to a newer version, they should not install the Vulkan SDK to get the loader but rather install the Vulkan Runtime Installer that will be at vulkan.lunarg.com

Kreyren:
Fixes: https://github.com/Winetricks/winetricks/issues/1368

### DIFF
--- a/src/winetricks
+++ b/src/winetricks
@@ -11785,24 +11785,28 @@ load_vjrun20()
 
 #----------------------------------------------------------------
 
+# File changes https://github.com/Winetricks/winetricks/pull/1370#issuecomment-540879256
+
 w_metadata vulkanrt dlls \
     title="Vulkan Runtime Latest" \
     publisher="LunarG" \
+    homepage="https://vulkan.lunarg.com/sdk/home" \
     year="2018" \
     media="download" \
     file1="vulkan-runtime.exe" \
-    installed_exe1="$W_SYSTEM32_DLLS_WIN/vulkaninfo.exe" \
-    installed_exe2="$W_SYSTEM32_DLLS_WIN/vulkaninfo-1-999-0-0-0.exe" \
-    installed_exe3="$W_SYSTEM32_DLLS_WIN/vulkaninfo-1-999-0-0-0.exe" \
-    installed_exe4="C:/windows/syswow64/vulkaninfo.exe"
+    installed_file1="C:\windows\system32\vulkan-1-999-0-0-0.dll" \
+    installed_file2="C:\windows\syswow64\vulkan-1-999-0-0-0.dll" \
+    installed_file3="C:\windows\syswow64\vulkaninfo-1-999-0-0-0.dll" \
+    installed_exe1="C:\windows\system32\vulkaninfo-1-999-0-0-0.exe" \
+    installed_exe2="C:\windows\system32\vulkaninfo.exe" \
+    installed_exe3="C:\windows\syswow64\vulkaninfo.exe"
 
 load_vulkanrt()
 {
-    # https://vulkan.lunarg.com/sdk/home
-    # https://github.com/Winetricks/winetricks/issues/1368#issue-505379537
+    # vulkanrt is no longer present in vulkansdk (https://github.com/Winetricks/winetricks/issues/1368#issue-505379537)
     # TODO: Checksum
     w_download "https://sdk.lunarg.com/sdk/download/latest/windows/vulkan-runtime.exe?u=" '' 'vulkan-runtime.exe'
-    w_try "$WINE" "$W_CACHE/$W_PACKAGE/$file1" $W_UNATTENDED_SLASH_S
+    w_try "$WINE" "$W_CACHE/$W_PACKAGE/$file1" "$W_UNATTENDED_SLASH_S"
 }
 
 #----------------------------------------------------------------
@@ -11810,6 +11814,7 @@ load_vulkanrt()
 w_metadata vulkansdk dlls \
     title="Vulkan SDK 1.1.73.0 (developers only)" \
     publisher="LunarG" \
+    homepage="https://vulkan.lunarg.com/sdk/home" \
     year="2018" \
     media="download" \
     file1="VulkanSDK-1.1.73.0-Installer.exe" \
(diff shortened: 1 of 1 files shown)


Kreyren:
Changed files for vulkanrt
```
drive_c/Program Files (x86)/VulkanRT/
drive_c/windows/system32/vulkan-1-999-0-0-0.dll
drive_c/windows/system32/vulkaninfo-1-999-0-0-0.exe
drive_c/windows/system32/vulkaninfo.exe
drive_c/windows/syswow64/vulkan-1-999-0-0-0.dll
drive_c/windows/syswow64/vulkaninfo-1-999-0-0-0.dll
drive_c/windows/syswow64/vulkaninfo.exe
```

Kreyren:
ping @KarenGhavam-lunarG: Can you provide me an raw file with sha256sum for https://sdk.lunarg.com/sdk/download/latest/windows/vulkan-runtime.exe ?

Kreyren:
@KarenGhavam-lunarG you said that vulkanSDK won't have vulkan runtime in the future -> Is there a way to check if vulkanSDK has vulkanrt so that i could make a logic:

`If VulkanSDK doesn't have vulkan runtime -> Install runtime`

KarenGhavam-lunarG:
@Kreyren You don't need to check if an SDK includes the runtime installer package. There already is and always will be a runtime installer package at the permanent link: https://vulkan.lunarg.com/sdk/home#sdk/downloadConfirm/latest/windows/vulkan-runtime.exe

Kreyren:
@KarenGhavam-lunarG But for the SDK assuming that it provided said runtime before and now it won't (assuming that i understand correctly)

KarenGhavam-lunarG:
On Vulkan.lunarg.com there are two packages you can download:
1. The SDK (which includes a runtime installer)
2. The runtime installer

In the future the SDK will NOT include the runtime installer. I believe Wine has been installing an SDK to get both the SDK and the runtime installer.  To get the runtime installer you can get it from the permanent link (that has existed for several years now).

If you go to vulkan.lunarg.com you will see that there is a permanent link for the latest SDK (https://vulkan.lunarg.com/sdk/home#sdk/downloadConfirm/latest/windows/vulkan-sdk.exe) and for the latest runtime (https://vulkan.lunarg.com/sdk/home#sdk/downloadConfirm/latest/windows/vulkan-runtime.exe). You could always put the latest SDK and the latest runtime on a Wine installation by using these permanent links.

Kreyren:
> I believe Wine has been installing an SDK to get both the SDK and the runtime installer.

Confirmed which is fixed in this Merge Request.

> In the future the SDK will NOT include the runtime installer. 

This commit is for this scenario assuming that it's expected to get Vulkan SDK with Vulkan runtime.

a) If we want vulkan SDK using winetricks we need to first install the Vulkan runtime and then the VulkanSDK -- OR -- Is it irelevant and should be kept as separete verb?

b) if VulkanSDK requires vulkan runtime -- and -- if VulkanSDK will be in future provided without vulkan runtime; then i want to merge it now with logic `If VulkanSDK doesn't have Vulkan Runtime; then install Vulkan Runtime` -> How should i do it? What is the difference in VulkanSDK that would be sufficient for this logic?

c) Checksum? (review above >  Can you provide me an url on raw file with sha256sum for https://sdk.lunarg.com/sdk/download/latest/windows/vulkan-runtime.exe ? <)

KarenGhavam-lunarG:
From this conversation I conclude that the permanent link to the latest runtime installer is not useful because you also want to know the SHA and since the latest runtime installer will keep changing about every weeks I can't give you a SHA that won't be changing.

So you can choose a version of the SDK and runtime you want installed on Wine (you are currently at 1.1.73.0 which is quite old...) and download those and reference their SHA codes accordingly. Although it shouldn't matter, install the runtime installer first. Then install the SDK.

Kreyren:
> I can't give you a SHA that won't be changing.

SHA that would change to match latest release is expected so that we could download latest and check if checksum matches (currently not possible based on available info).

Some SHA in a form of raw file would also be welcomed for the other versions so that SHA would be handled by upstream (LunarG) instead of us generating new SHA for each version manually and hardcoding it in.

> Although it shouldn't matter, install the runtime installer first. Then install the SDK.

Noted, so change in winetricks is **not** required for future compatibility based on available info.

If this is a wine issue filed to be archived then you should probably file a winebug since winetricks is separate project (afaik).

> you are currently at 1.1.73.0 which is quite old...

I can make commit for other versions of `vulkanrt` if needed.. this is probably up to @austin987 to decide since i don't know if older version is used for a reason.

KarenGhavam-lunarG:
> Some SHA in a form of raw file would also be welcomed for the other versions so that SHA would be handled by upstream (LunarG) instead of us generating new SHA for each version manually and hard-coding it in.

So to avoid hard coding the SHA in your code, you would use a permanent link to the runtime installer package and a permanent link to a text file that contained the SHA for the runtime installer? If yes, I could get that added to vulkan.lunarg.com. It will take a while before that change is made at vulkan.lunarg.com. Once it is ready I will update the [original issue](https://github.com/Winetricks/winetricks/issues/1368) I submitted.

> If this is a wine issue filed to be archived then you should probably file a winebug since winetricks is separate project (afaik).

I originally submitted an issue to WineHQ Bugzilla and was told that since it wasn't a bug and had nothing to do with WineHQ packages I should submit to this project instead.

Kreyren:
>  If yes, I could get that added to vulkan.lunarg.com. 

Thanks!

> I originally submitted an issue to WineHQ Bugzilla and was told that since it wasn't a bug and had nothing to do with WineHQ packages I should submit to this project instead.

Based on current code winetricks is not using vulkansdk to get vulkan runtime -> invalid

Reference the winebug please

austin987:
Superseded by https://github.com/Winetricks/winetricks/commit/8418d3c32e1fad024ef463b5a930a509b0d7b07b

Issue #1132, opened 2018-11-12
Media Fountain needs enhancements


ahmed-elsayed2017:
When I try installing MF using Winetricks, it installs only mf.dll file (native, builtin), but some games needs more files like:

mferror.dll
mfplat.dll
mfreadwrite.dll
msmpeg2adec.dll
msmpeg2vdec.dll
sqmapi.dll

You can get more information from this issue:
https://github.com/ValveSoftware/Proton/issues/1102

ahmed-elsayed2017:
https://github.com/doitsujin/dxvk/issues/761

lieff:
I've already tried to make winetricks PR, but I've failed to find KB update or other download containing this files. Original media foundation downloads contains delta files except for win10. But win10 versions do not works with wine.

ahmed-elsayed2017:
What if I changed Windows version to 10 from Wine configurations and tried win10 version of MF?

lieff:
Nope, It uses functions unsupported by wine like RtwqUnlockWorkQueue.

ahmed-elsayed2017:
Any new news regarding MF?

james-munson:
https://github.com/jzengc/wine/tree/proton_3.16_mf

ahmed-elsayed2017:
@james-munson Thanks, but did you try it yourself?

Glog78:
Lutris seem to use win7-sp1 for extraction of the files -> https://lutris.net/games/install/10999/view . Maybe that could help.

ahmed-elsayed2017:
This workaround works for some games like Danganronpa V3: Killing Harmony, Resident Evil 7 Biohazard, Resident Evil 2 Remake, and Darksiders Warmastered Edition.

There is another workaround mentioned here >>> 

**So this MF workaround provides:**

>  mf.dll
> mfplate.dll
> mfreadwrite.dll
> wmadmod.dll
> wmvdecod.dll

and the other MF workaround provides:

> mf.dll
> mferror
> mfplate.dll
> mfreadwrite.dll
> msmpeg2adec.dll
> msmpeg2vdec.dll
> sqmapi.dll

and it also some registry files to be imported as well, and that works with other games like Shadows: Awakening, and Thronebreaker: The Witcher Tales ... etc

All the info are in this post >>> https://github.com/ValveSoftware/Proton/issues/1102#issuecomment-419678627

[...]

PR #1384, opened 2019-10-19
mf: Added missing and mirror

Kreyren:
DO NOT MERGE

Disclaimer: i didn't mess with regedit4, grapped from @austin987 's 
branch https://github.com/austin987/winetricks/commit/5d653bfd9f2282a1d77edd3e4ed71205be1c4894 and @z0z0z 's repo (https://github.com/z0z0z/mf-install)

Relevant: https://github.com/Winetricks/winetricks/issues/1132

Signed-off-by: Jacob Hrbek

### DIFF
--- a/src/winetricks
+++ b/src/winetricks
@@ -10409,7 +10409,7 @@ load_mf()
 		[HKEY_CLASSES_ROOT\CLSID\{48e2ed0f-98c2-4a37-bed5-166312ddd83f}\InprocServer32]
 		@="mfreadwrite.dll"
 		"ThreadingModel"="Both"
-		_EOF_
+_EOF_
 
 		# Generate wmf.reg
 		cat > "$W_CACHE/$W_PACKAGE/wmf.reg" <<_EOF_
@@ -10625,7 +10625,7 @@ load_mf()
 
 		[HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Windows Media Foundation\SchemeHandlers\sdp:]
 		"{E9F4EBAB-D97B-463e-A2B1-C54EE3F9414D}"="Net Scheme Handler"
-		_EOF_
+_EOF_
 
 		w_try_regedit "$W_CACHE/$W_PACKAGE/mf.reg"
 		w_try_regedit "$W_CACHE/$W_PACKAGE/wmf.reg"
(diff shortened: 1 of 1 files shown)


Kreyren:
It exports the dlls correctly, but it doesn't work in test wineapp where script from z0z0z works no issue.. i have no idea what is different and i'm going to break something now to feel better and revisit later

src/winetricks
```diff
@@ -10174,25 +10174,485 @@ _EOF_
...
+		@="mf.dll"
+		"ThreadingModel"="Both"
+
+		[HKEY_CLASSES_ROOT\CLSID\{48e2ed0f-98c2-4a37-bed5-166312ddd83f}]
+		@="MFReadWrite Class Factory"
+
+		[HKEY_CLASSES_ROOT\CLSID\{48e2ed0f-98c2-4a37-bed5-166312ddd83f}\InprocServer32]
+		@="mfreadwrite.dll"
+		"ThreadingModel"="Both"
+		_EOF_
```
m42a: This does not work because `_EOF_` is indented, so the shell doesn't terminate the here-document on this line
```suggestion
_EOF_
```

src/winetricks
```diff
@@ -10174,25 +10174,485 @@ _EOF_
...
+
+		[HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Windows Media Foundation\SchemeHandlers\rtspt:]
+		"{E9F4EBAB-D97B-463e-A2B1-C54EE3F9414D}"="Net Scheme Handler"
+
+		[HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Windows Media Foundation\SchemeHandlers\rtspu:]
+		"{E9F4EBAB-D97B-463e-A2B1-C54EE3F9414D}"="Net Scheme Handler"
+
+		[HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Windows Media Foundation\SchemeHandlers\sdp:]
+		"{E9F4EBAB-D97B-463e-A2B1-C54EE3F9414D}"="Net Scheme Handler"
+		_EOF_
```
m42a: This also does not work because `_EOF_` is indented, so the shell doesn't terminate the here-document on this line
```suggestion
_EOF_
```

Kreyren:
@m42a There is current issue with it not working compared to z0z0z's script this seems to be the issue (Thanks so much if that's the case!!! I spent so much time trying to debug this and eventually ended up depressed on rocket league) i try to setup an environment for it to test since i'm on vacation with limited options.

PR #1390, opened 2019-10-20
w_check_exec: New function

Kreyren:
I'm using this locally since `[ -x $(command -v 'something') ] && { 
something :} || nothing` is unreliable

proof:
`[ -x $(command -v 'kreyren') ] && { printf "Unexpected" ;} || printf 
"Expecting fail"`
-- OR --
`[ -x $(which 'kreyren') ] && { printf "Unexpected" ;} || printf 
"Expecting fail"`
^ Notice that if conditions output zero the logic considers it as true 
assuming system without 'kreyren' executable in PATH which based on my 
experience is the most used logic for this

### DIFF
--- a/src/winetricks
+++ b/src/winetricks
@@ -718,6 +718,27 @@ w_expand_env()
     winetricks_early_wine cmd.exe /c echo "%$1%"
 }
 
+# Check if $1 is executable (output custom message using w_debug if $2 is specified) and return 126 (standart non-executable exit code) or 0 if executable
+w_check_exec()
+{
+    w_executable="$1"
+    w_message="$2"
+
+    if ! command -v "$w_executable" 2>/dev/null; then
+        if [ -n "$w_message" ]; then
+            printf '%s\n' "$w_message"
+            exit 126
+        elif [ -z "$w_message" ]; then
+            printf '%s\n' "Command $w_executable is not executable"
+            exit 126
+        fi
+    elif command -v "$w_executable" 2>/dev/null; then
+        w_debug "Command '$w_executable' is executable"
+    fi
+
+    unset w_executable w_message
+}
+
 # Get the latest tagged release from github.com API
 w_get_github_latest_release()
 {


src/winetricks
```diff
@@ -718,6 +718,26 @@ w_expand_env()
...
 
+# Check if $1 is executable (output custom message using w_debug if $2 is specified) and return 126 (standart non-executable exit code) or 0 if executable
+w_check_exec()
+{
+    w_executable="$1"
+    w_message="$2"
+
+    if ! command -v "$w_executable" 2>/dev/null; then
+        if [ -n "$w_message" ]; then
+            printf '%s\n' "$w_message"
```
austin987: This should be w_warn

src/winetricks
```diff
@@ -718,6 +718,26 @@ w_expand_env()
...
+            return 126
+        elif [ -z "$w_message" ]; then
+            printf '%s\n' "Command $w_executable is not executable"
```
austin987: Also w_warn

src/winetricks
```diff
@@ -718,6 +718,26 @@ w_expand_env()
...
+            return 126
+        fi
+    elif command -v "$w_executable" 2>/dev/null; then
```
austin987: You don't need to run command again; either passed or it failed the first time.

src/winetricks
Kreyren: It's expected to use this for logic for example if p7zip (linux) is not executable -> Use windows 7-Zip

So it's made to return `0` or `126` for logic.  -> I don't think that we should bother user with `w_warn()` assuming GUI used.

src/winetricks
Kreyren: Same reasoning as https://github.com/Winetricks/winetricks/pull/1390#discussion_r336954402

src/winetricks
Kreyren: `else` is going to trigger on anything unexpected where provided logic might return unexpected on non-standard system -> Reasoning to use `elif`

Or do you want me to use `else` anyway?

austin987: Unless you have an actual example of what something unexpected could be, yes, use else.

In other words, it's either executable or it's not.

PR #1418, opened 2019-11-07
w_do_call: always set the wineprefix

austin987:
Previously dlls/settings weren't setting this. This causes a problem if
prefix=foobar is used for a dll; ~/.wine (or $WINEPREFIX) would get used
for winetricks_is_installed() instead of $HOME/.local/share/wineprefixes/$cmd

This broke things when the WINEARCH didn't match.

### DIFF
--- a/src/winetricks
+++ b/src/winetricks
@@ -2754,14 +2754,13 @@ w_do_call()
             *)
                 # shellcheck disable=SC2154
                 case "${category}-${WINETRICKS_OPT_SHAREDPREFIX}" in
-                    apps-0|benchmarks-0|games-0)
-                        winetricks_set_wineprefix "$cmd"
-                        # If it's a new wineprefix, give it metadata
-                        if test ! -f "$WINEPREFIX"/wrapper.cfg; then
-                            echo ww_name=\""$title"\" > "$WINEPREFIX"/wrapper.cfg
-                        fi
-                        ;;
+                    apps-0|benchmarks-0|games-0) winetricks_set_wineprefix "$cmd";;
+                    *) winetricks_set_wineprefix;;
                 esac
+                # If it's a new wineprefix, give it metadata
+                if test ! -f "$WINEPREFIX"/wrapper.cfg; then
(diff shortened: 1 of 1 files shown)


src/winetricks
```diff
@@ -2754,14 +2754,13 @@ w_do_call()
...
                 case "${category}-${WINETRICKS_OPT_SHAREDPREFIX}" in
-                    apps-0|benchmarks-0|games-0)
-                        winetricks_set_wineprefix "$cmd"
-                        # If it's a new wineprefix, give it metadata
-                        if test ! -f "$WINEPREFIX"/wrapper.cfg; then
-                            echo ww_name=\""$title"\" > "$WINEPREFIX"/wrapper.cfg
-                        fi
-                        ;;
+                    apps-0|benchmarks-0|games-0) winetricks_set_wineprefix "$cmd";;
+                    *) winetricks_set_wineprefix;;
```
Kreyren: > *) winetricks_set_wineprefix;;

I don't think that this is sane since it may trigger if wrong category is parsed..

austin987: Do you have an example?

Kreyren: Negative, i don't have working wine on this system due to the damage caused vacation where i'm at still, will test once i return (ETA 2,5 weeks)

In theory parsing dll verb is going to trigger `*)` or using upcoming `tool` category which i'm trying to contrib

from a sanitization point of view in general using wild cards is likely to break on other systems, is prone to hide issues and makes debugging harder thus reasoning for me to suggest this.

austin987: Comments like this are not helpful at all. You're welcome to criticize or point out issues, but saying you *think* things may not work with *no* evidence, when the change *fixes* an *actual problem* is categorically not helpful and mildly insulting.

Parsing dll is *exactly* the problem. If a dll or any other category of verb is called, winetricks_set_wineprefix won't be called here, which means prefix=foo isn't respected.

Kreyren: Noted, sorry

PR #1429, opened 2019-11-21
reorder -f|--force options, required by bash-completion script

bobwya:
* winetricks.bash-completion: requries that short options are
   listed before long options, in the BASH case statement entries.
   This is due to simplified regex handling employed in the
   bash-completion script.
 * fixes a regression introduced by:
   f40b4655f8f3951730b271ee02ebe9dbcc619ecc


austin987:
May be worth adding a check for that to `./tests/shell-checks`.

PR #1440, opened 2019-11-27
Add powershell app based on PowerShell Core 6.1,

theli-ua:
functionality enough to satisfy installers requiring powershell

For example this allows Magic Arena ( https://magic.wizards.com/en/mtgarena ) installer work in wine

Note: https://docs.microsoft.com/en-us/powershell/scripting/install/installing-powershell-core-on-windows?view=powershell-6  says WMF is a prerequisite but I have not found a way to install it. And installing it this way is at least enough for installer to work (thus warns). 
It is also not clear what dotnet version it requires ... it works with 462 but probably any 45+ will work (not sure how to reflect such dependency)

### DIFF
--- a/src/winetricks
+++ b/src/winetricks
@@ -14982,8 +14982,7 @@ load_powershell()
     w_try_7z  "$W_SYSTEM32_DLLS"/WindowsPowerShell/v1.0/ "${W_CACHE}/${W_PACKAGE}/${file1}"
     w_try_cd "$W_SYSTEM32_DLLS"/WindowsPowerShell/v1.0
     w_try cp -f pwsh.exe powershell.exe
-    for file in *\\*
-    do
+    for file in *\\*; do
         target=$(printf '%s\n' "$file" | sed 's#\\#/#g')
         w_try mkdir -p "${target%/*}"
         w_try mv -v "$file" "$target"
--- a/src/winetricks
+++ b/src/winetricks
@@ -14963,17 +14963,13 @@ w_metadata powershell apps \
 
 load_powershell()
 {
-    if test "$W_ARCH" = "win64"; then
-        w_die "Only implemented for 32bit arch at the moment"
-    fi
+    w_package_unsupported_win64
 
-    w_warn "Installing Powershell into $W_SYSTEM32_DLLS_WIN/WindowsPowerShell/v1.0/"
-    w_warn "So that apps that require psh scripts to be executed would find it"
-    w_warn "Note that some functionality may not work without Windows Management Framework"
+    w_warn "Installing Powershell into $W_SYSTEM32_DLLS_WIN/WindowsPowerShell/v1.0/. Note that some functionality may not work without Windows Management Framework"
 
     w_call dotnet462
 
-    w_download https://github.com/PowerShell/PowerShell/releases/download/v6.1.6/PowerShell-6.1.6-win-x86.zip 1dc690c9a1091d0ed5b71de903ab3cd7b58ba9638acdfeb9f9fe7031abaf47e9 PowerShell-6.1.6-win-x86.zip
+    w_download https://github.com/PowerShell/PowerShell/releases/download/v6.1.6/PowerShell-6.1.6-win-x86.zip 1dc690c9a1091d0ed5b71de903ab3cd7b58ba9638acdfeb9f9fe7031abaf47e9
 
     w_override_dlls native powershell.exe
 
@@ -14986,7 +14982,12 @@ load_powershell()
     w_try_7z  "$W_SYSTEM32_DLLS"/WindowsPowerShell/v1.0/ "${W_CACHE}/${W_PACKAGE}/${file1}"
     w_try_cd "$W_SYSTEM32_DLLS"/WindowsPowerShell/v1.0
     w_try cp -f pwsh.exe powershell.exe
-    for file in *\\*; do target="${file//\\//}"; mkdir -p "${target%/*}"; mv -v "$file" "$target"; done
+    for file in *\\*
+    do
+        target=$(printf '%s\n' "$file" | sed 's#\\#/#g')
+        w_try mkdir -p "${target%/*}"
+        w_try mv -v "$file" "$target"
+    done
     w_append_path "$W_SYSTEM32_DLLS"/WindowsPowerShell/v1.0
 }
 
--- a/src/winetricks
+++ b/src/winetricks
@@ -14977,7 +14977,7 @@ load_powershell()
     rm -f "$W_SYSTEM32_DLLS"/powershell.exe
 
 
-    mkdir -p "$W_SYSTEM32_DLLS"/WindowsPowerShell/v1.0
+    w_try mkdir -p "$W_SYSTEM32_DLLS"/WindowsPowerShell/v1.0
 
     w_try_7z  "$W_SYSTEM32_DLLS"/WindowsPowerShell/v1.0/ "${W_CACHE}/${W_PACKAGE}/${file1}"
     w_try_cd "$W_SYSTEM32_DLLS"/WindowsPowerShell/v1.0
--- a/src/winetricks
+++ b/src/winetricks
@@ -14953,6 +14953,45 @@ load_openwatcom()
 
 #----------------------------------------------------------------
 
+w_metadata powershell apps \
+    title="Windows Powershell" \
+    publisher="Microsoft" \
+    year="2019" \
+    media="download" \
+    file1="PowerShell-6.1.6-win-x86.zip" \
+    installed_file1="$W_SYSTEM32_DLLS_WIN/WindowsPowerShell/v1.0/powershell.exe"
+
+load_powershell()
+{
+    if test "$W_ARCH" = "win64"; then
+        w_die "Only implemented for 32bit arch at the moment"
+    fi
+
+    w_warn "Installing Powershell into $W_SYSTEM32_DLLS_WIN/WindowsPowerShell/v1.0/"
+    w_warn "So that apps that require psh scripts to be executed would find it"
+    w_warn "Note that some functionality may not work without Windows Management Framework"
+
+    w_call dotnet462
+
+    w_download https://github.com/PowerShell/PowerShell/releases/download/v6.1.6/PowerShell-6.1.6-win-x86.zip 1dc690c9a1091d0ed5b71de903ab3cd7b58ba9638acdfeb9f9fe7031abaf47e9 PowerShell-6.1.6-win-x86.zip
+
+    w_override_dlls native powershell.exe
+
+    # remove builtin placeholders
+    rm -f "$W_SYSTEM32_DLLS"/powershell.exe
+
+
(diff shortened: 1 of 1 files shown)


src/winetricks
```diff
@@ -14953,6 +14953,45 @@ load_openwatcom()
...
+    rm -f "$W_SYSTEM32_DLLS"/powershell.exe
+
+
+    mkdir -p "$W_SYSTEM32_DLLS"/WindowsPowerShell/v1.0
+
+    w_try_7z  "$W_SYSTEM32_DLLS"/WindowsPowerShell/v1.0/ "${W_CACHE}/${W_PACKAGE}/${file1}"
+    w_try_cd "$W_SYSTEM32_DLLS"/WindowsPowerShell/v1.0
+    w_try cp -f pwsh.exe powershell.exe
+    for file in *\\*; do target="${file//\\//}"; mkdir -p "${target%/*}"; mv -v "$file" "$target"; done
+    w_append_path "$W_SYSTEM32_DLLS"/WindowsPowerShell/v1.0
```
theli-ua: Reason for this is this is a non-compliant zip file that has back slashes. While unzip detects that and extracts archive reversing slashes it fails to successfully extract archive due to a different reasn (for w/e reason it creates  one of directories without executable bit and then fails to extract files into it with permission denied)

Kreyren:
So you don't need WMF to satisfy MTGA?  o.O

theli-ua:
No, you don't

theli-ua:
> So you don't need WMF to satisfy MTGA? o.O

no, you don't

src/winetricks
```diff
@@ -14953,6 +14953,45 @@ load_openwatcom()
...
+    year="2019" \
+    media="download" \
+    file1="PowerShell-6.1.6-win-x86.zip" \
+    installed_file1="$W_SYSTEM32_DLLS_WIN/WindowsPowerShell/v1.0/powershell.exe"
+
+load_powershell()
+{
+    if test "$W_ARCH" = "win64"; then
+        w_die "Only implemented for 32bit arch at the moment"
+    fi
```
austin987: You can use w_package_unsupported_win64 for that.

src/winetricks
```diff
@@ -14953,6 +14953,45 @@ load_openwatcom()
...
+
+    w_warn "Installing Powershell into $W_SYSTEM32_DLLS_WIN/WindowsPowerShell/v1.0/"
+    w_warn "So that apps that require psh scripts to be executed would find it"
+    w_warn "Note that some functionality may not work without Windows Management Framework"
```
austin987: This will give 3 popups in GUI mode, please combine it into one.

src/winetricks
```diff
@@ -14953,6 +14953,45 @@ load_openwatcom()
...
+
+    w_call dotnet462
+
+    w_download https://github.com/PowerShell/PowerShell/releases/download/v6.1.6/PowerShell-6.1.6-win-x86.zip 1dc690c9a1091d0ed5b71de903ab3cd7b58ba9638acdfeb9f9fe7031abaf47e9 PowerShell-6.1.6-win-x86.zip
```
austin987: Since you're not changing the filename, there's no need to specify it as the last argument.

src/winetricks
```diff
@@ -14953,6 +14953,45 @@ load_openwatcom()
...
+    # remove builtin placeholders
+    rm -f "$W_SYSTEM32_DLLS"/powershell.exe
+
+
+    mkdir -p "$W_SYSTEM32_DLLS"/WindowsPowerShell/v1.0
+
+    w_try_7z  "$W_SYSTEM32_DLLS"/WindowsPowerShell/v1.0/ "${W_CACHE}/${W_PACKAGE}/${file1}"
+    w_try_cd "$W_SYSTEM32_DLLS"/WindowsPowerShell/v1.0
+    w_try cp -f pwsh.exe powershell.exe
+    for file in *\\*; do target="${file//\\//}"; mkdir -p "${target%/*}"; mv -v "$file" "$target"; done
```
austin987: mkdir/mv should be behind w_try().

This also introduces a bashism, causing travis to fail.

src/winetricks
theli-ua: I can technically add win64 support if you would prefer that. I don't really see an issue adding it.
However if you would like that indeed - what would be the best example of handling separate x86/64 archives?

austin987: That would be preferable, but not required.

There are several examples, e.g., amstream:
https://github.com/Winetricks/winetricks/blob/master/src/winetricks#L5586

you'll want to unconditionally install win32 (as win64 always supports win32), then selectively install win64 if $W_ARCH = win64.

theli-ua: I am honestly not sure what is the right way to install this properly and integrate into system and how PATH is supposed to be set up in the right way if both 32 and 64 bit things are installed so I guess I'll just leave it as is here

tannisroot:
64-bit prefix support would be really nice! There's an app (MTGA) that needs powershell for the installation process but it seems to crash in 32-bit prefix for some reason.
And it's also very likely that other potentials apps that will depend on powershell are going to be 64-bit. 
Also, perhaps powershell could be categorized not as an app but as a Windows component, since it is one?

theli-ua:
> 64-bit prefix support would be really nice! There's an app (MTGA) that needs powershell for the installation process but it seems to crash in 32-bit prefix for some reason.

well, MTGA installer is actually what this PR has been tested with. There is no reason it should require 64bit prefix

tannisroot:
> There is no reason it should require 64bit prefix

oh, weird, you're right. My prefix must have been borked during testing.
In that case, I guess 64-bit support can be added whenever a new 64-bit app will need powershell.

PietJankbal:
Hi, i attached script that also installs 64 bit powershell here:
 https://bugs.winehq.org/attachment.cgi?id=65974
Again , script has hard coded "Program Files" (not localized) because i dunno how to get around that issue. After installation powershell.exe can be run to start it

tannisroot:
@PietJankbal does this actually work? because it specified `w_package_unsupported_win64`, and I am pretty sure that won't allow you to install the verb into 64-bit prefix.

tannisroot:
I think I might have found a valid usecase for a 64-bit powershell verb.
I've discovered that MTGA doesn't seem to render properly with WineD3D on Ultra settings, but it works just fine with DXVK. However, to work correctly, DXVK requires you to launch 32-bit apps in forced Large Address Aware mode so that it doesn't run out of memory. However, that option doesn't work in 32-bit prefixes, only in 64-bit ones.
So it would be really lovely to add 64-bit support to avoid these issues that might occur with 32-bit prefixes.

PietJankbal:
> @PietJankbal does this actually work? because it specified `w_package_unsupported_win64`, and I am pretty sure that won't allow you to install the verb into 64-bit prefix.

That is commented out, it should be reoved really

tannisroot:
Oh, didn't notice it's commented out.

src/winetricks
```diff
@@ -14953,6 +14953,45 @@ load_openwatcom()
...
+    title="Windows Powershell" \
+    publisher="Microsoft" \
+    year="2019" \
+    media="download" \
+    file1="PowerShell-6.1.6-win-x86.zip" \
+    installed_file1="$W_SYSTEM32_DLLS_WIN/WindowsPowerShell/v1.0/powershell.exe"
+
+load_powershell()
+{
+    w_package_unsupported_win64
```
austin987: Can be removed, as you said (proper 64-bit support can be a followup PR).

src/winetricks
```diff
@@ -14953,6 +14953,45 @@ load_openwatcom()
...
+    w_warn "Installing Powershell into $W_SYSTEM32_DLLS_WIN/WindowsPowerShell/v1.0/. Note that some functionality may not work without Windows Management Framework"
+
+    w_call dotnet462
+
+    w_download https://github.com/PowerShell/PowerShell/releases/download/v6.1.6/PowerShell-6.1.6-win-x86.zip 1dc690c9a1091d0ed5b71de903ab3cd7b58ba9638acdfeb9f9fe7031abaf47e9
+
+    w_override_dlls native powershell.exe
+
+    # remove builtin placeholders
+    rm -f "$W_SYSTEM32_DLLS"/powershell.exe
```
austin987: Prepend with `w_try`

src/winetricks
```diff
@@ -14953,6 +14953,45 @@ load_openwatcom()
...
+
+
```
austin987: Please remove the extra newline.

src/winetricks
theli-ua: I wasn't planning on adding win64 support

theli-ua:
> A couple minor things, then this can go in as far as I'm concerned (unless you want to get 64-bit working before merging).

frankly, I'm not sure anymore this has much value. It does work after being installed but I'm not sure why anyone would use it. Original intent was to make Magic Arena installer work but as it happens just moving wine's stub to a proper location fixes that (in 5.0-rc4).

PietJankbal:
Hi,

A few remarks. As theli-ua already pointed out wine`s stub has moved so this line is not up to date any more:
 rm -f "$W_SYSTEM32_DLLS"/powershell.exe

Anyway, i still think it`s useful to have powershell installed and (a bit of)-functional so here`s my five cents (did also a bit of googling on this):

Powershell 6 and up are really kind of "add-ons" to Windows so those versions of powershell should be called by "wine pwsh.exe". The installers for powershell ver. 6 and up work fine if you put a fake pwrshlplugin.dll (with high enough version resource) in system32/wow64. 
But the pwsh.exe`s should really  remain in Program Files/Powershell and not be copied  (or linked) to system32/WindowsPowerShell/v1.0/powershell.exe 

The right way to go is i think install powershell2.0 (i have a script for this and it installs and with a bit of aftermath repair work because of bug 25740 it has at least minimal functionality like reporting version correctly) Version 2.0 is  btw also the version that is reported on my win7 partition when you start powershell (see e.g. https://appuals.com/how-to-check-powershell-version-on-windows-10-8-and-7/)

I`d first like to finish my internet explorer server_2003 script but after that i`d like to pick this up , ok?

Regards

theli-ua:
@PietJankbal sure, feel free to pick this up

PietJankbal:
See https://raw.githubusercontent.com/PietJankbal/temporary_stuff/master/pwsh20.diff

I`ll do PR if my other Pull request are settled down

tannisroot:
@PietJankbal the last link is now broken as you removed your repository

PietJankbal:
> @PietJankbal the last link is now broken as you removed your repository

Hi, i rearranged some files, it`s here: https://github.com/PietJankbal/storage/blob/master/pwsh20.diff
I stopped trying to get things into winetricks (too frustating...) if anyone would like to pick this up please go ahead

allentiak:
Hi, @austin987, @theli-ua 

I could take over here. Should I copy this branch into a new PR? Or I can someone assign this one to me?

austin987:
@allentiak I'm not sure if GH will let you push over another user's branch (it tells me I can, but I'm the maintainer).

I'd suggest making your own PR.

allentiak:
Then, I'll make a new PR out of this one. Give me a few days.

PR #1450, opened 2019-12-16
new verb: d9vk040

bobwya:


axredneck:
Then better add dxvk15 which contains D9VK 0.40.1

PR #1457, opened 2019-12-26
Add vcrun2019 verb

tannisroot:
Got links from here https://support.microsoft.com/en-us/help/2977003/the-latest-supported-visual-c-downloads

### DIFF
--- a/src/winetricks
+++ b/src/winetricks
@@ -12051,7 +12051,7 @@ w_metadata vcrun2015 dlls \
     publisher="Microsoft" \
     year="2015" \
     media="download" \
-    conflicts="vcrun2017" \
+    conflicts="vcrun2017 vcrun2019" \
     file1="vc_redist.x86.exe" \
     installed_file1="$W_SYSTEM32_DLLS_WIN/mfc140.dll"
 
@@ -12114,6 +12114,7 @@ w_metadata vcrun2017 dlls \
     publisher="Microsoft" \
     year="2017" \
     media="download" \
+    conflicts="vcrun2019" \
     file1="vc_redist.x86.exe" \
     installed_file1="$W_SYSTEM32_DLLS_WIN/mfc140.dll"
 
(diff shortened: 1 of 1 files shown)


src/winetricks
```diff
@@ -12114,6 +12114,7 @@ w_metadata vcrun2017 dlls \
     publisher="Microsoft" \
     year="2017" \
     media="download" \
+    conflicts="vcrun2019" \
```
austin987: Nice catch, but shouldn't it also include vcrun2015?

src/winetricks
```diff
@@ -12176,6 +12177,37 @@ load_vcrun2017()
 
 #----------------------------------------------------------------
 
+w_metadata vcrun2019 dlls \
+    title="Visual C++ 2015-2019 libraries (concrt140.dll,mfc140.dll,mfc140u.dll,mfcm140.dll,mfcm140u.dll,msvcp140.dll,vcamp140.dll,vccorlib140.dll,vcomp140.dll,vcruntime140.dll)" \
+    publisher="Microsoft" \
+    year="2019" \
+    media="download" \
```
austin987: Missing conflicts here.

src/winetricks
tannisroot: I don't think newer versions conflict with older ones. The vcrun installers don't seem to complain about already installed vcrun2015 and I am pretty sure I've seen both vcrun2015 and vcrun2017 installed on actual Windows machines.

src/winetricks
tannisroot: See comment in the previous discussion.

src/winetricks
austin987: They both install mfc140.dll..

tannisroot: Yeah, which gets replaced with a newer version when you install a newer visual c++ runtime.

austin987: On windows; winetricks will detect that the file is already there and bail out unless --force is used.

tannisroot: I mean, I don't mind adding it in the conflict, but it could be useful to have an upgrade path for existing prefixes for whatever potential reason users might have.

tannisroot: Oh, I see, didn't know that :D Will add the conflict then.

austin987: It's out of the scope of this PR, but it may be doable to do like we do in newer .Net versions and touch a file (e.g., c:/windows/vcrun2019.installed) and use that for the installed file rather than mfc140.dll. That assumes of course that the installer's overwrite feature works..

PR #1464, opened 2020-01-04
new verb: dxtrans

oleg-dubinskiy:
Add a new dxtrans verb, which extracts dxtrans.dll from downloaded WindowsXP-KB936929-SP3-x86-ENU.exe, copies it to system32 and registers. That dll is required by Windows Movie Maker 2.0 installer and missing in Wine by default.
Just checked: after my chages, installing MS dxtrans works correctly.
I preferred using already implemented XPSP3 helper instead of adding a new DirectX 8.x helper, because it is a bit hard to find original DirectX 8.2 XP installer in the Web, since it was removed from the MS website.
Also I doubt about publishing year: should I use a year when the file has been changed last time or a year when that DirectX version has been released? :thinking:


austin987:
The published year isn't super important, more for reference (mostly for games/applications). Whenever it last changed or whenever winxpsp2 was released is fine.

PR #1472, opened 2020-01-10
new verb: dxvk151

bobwya:


src/winetricks
```diff
@@ -7924,6 +7924,26 @@ load_dxvk150()
     helper_dxvk_d9vk "$file1" "4.20" "1.1.113" "dxgi,d3d9,d3d10,d3d11"
 }
 
+w_metadata dxvk151 dlls \
```
austin987: Missing separator here (`#----------------------------------------------------------------`)

bobwya: @austin987 There aren't separators between the previous **dxvk???** verb versions. There are separators between the **dxvk???** verb versions group and the **dxvk** and **dxvk_master** verbs.

PR #1482, opened 2020-01-25
Fix dxvk_master verb downloading from d9vk repository

AlexTMjugador:
The referenced D9VK repository for this verb is achieved and unmaintained, as it was merged with DXVK. It no longer receives the latest builds of DXVK.

### DIFF
--- a/src/winetricks
+++ b/src/winetricks
@@ -7986,12 +7986,12 @@ w_metadata dxvk_master dlls \
 
 load_dxvk_master()
 {
-    # https://git.froggi.es/joshua/d9vk
+    # https://git.froggi.es/doitsujin/dxvk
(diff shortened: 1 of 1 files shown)


danger89:
oopsy :package:

bobwya:
> LGTM. @bobwya?

@austin987 

Sure it's fine (well accepting the that fact Joshie's Gitlab instance is flaky as hell!)

The artifact file generated by the Gitlab build bot has the form: "**master-_\<commit hash\>_.zip**" now. So I guess we want a way to preserve the archive name - to avoid re-downloading it and reloading the verb with a newer commit. This would probably need some way for **w_download_to** to pass back a filename, that has actually been downloaded, following redirections (perhaps)?

PR #1497, opened 2020-03-10
WIP: w_try_ms_installer: wrapper for microsoft installers

austin987:
I have only done the refactor, very little testing.

I also want to do a more generic solution, for vb6run/nook/etc, but it's late. PR'ing now for visibility.


austin987:
Ended up pushing directly after refactoring a bit, along with some other cleanups.

PR #1498, opened 2020-03-11
Use an unified Source Han Sans font for CJK Fonts

huupoke12:
This PR make the `fakechinese`, `fakejapanese` and `fakekorean` use the SuperOTC variant of Source Han Sans (all in one) to reduce storage and network due to less duplicated glyphs, and to resolve issue of some different regions glyph is missing in a region font.

I also added some CJK fonts replacement listed at https://en.wikipedia.org/wiki/List_of_typefaces_included_with_Microsoft_Windows.

I also removed `fakejapanese_vlgothic` from `cjkfonts` since it would override the existing replacements in `fakejapanese` (I didn't edit anything in `fakejapanese_vlgothic`). People can still install it by calling it directly. Please let me know if this change is not suitable and/or more change is needed.

### DIFF
--- a/src/winetricks
+++ b/src/winetricks
@@ -13731,7 +13731,6 @@ load_sourcehansans()
     w_register_font sourcehansans.ttc "Source Han Sans SC Medium"
     w_register_font sourcehansans.ttc "Source Han Sans SC Bold"
     w_register_font sourcehansans.ttc "Source Han Sans SC Heavy"
-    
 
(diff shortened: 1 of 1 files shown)


austin987:
Thanks, I manually squashed the trailing space fix and merged.

I appreciate your patience and working on this; it's much better for a native speaker to work on this than me ;).

PR #1504, opened 2020-03-23
wmp11: extract the main installer and then run subinstallers

austin987:
@soredake can you try this with proton? I made some changes to allow installing this on wine-devel, and I want to make sure it doesn't break things for proton.


soredake:
Works fine with re5+proton 5.0-5.

austin987:
Great, thanks!

PR #1511, opened 2020-04-08
New verb: dotnetcore2

Aliervo:
Mount and Blade 2: Bannerlord requires the .NET Core Runtime 2.1

I attempted to follow the other dotnet verbs as well as vcrun verbs as a guide/template.


src/winetricks
```diff
@@ -9439,6 +9439,34 @@ verify_dotnet48()
 
 #----------------------------------------------------------------
 
+w_metadata dotnetcore2 dlls \
+    title="MS .NET Core Runtime 2.1 LTS" \
+    publisher="Microsoft" \
+    year="2020" \
+    media="download" \
+    conflicts="" \
```
austin987, line 9447: Nitpick, but you don't need to define this if there's not a conflict (but good on you for paying attention ;)).

src/winetricks
```diff
@@ -9439,6 +9439,34 @@ verify_dotnet48()
...
+
+load_dotnetcore2()
+{
+    # Official version, see https://dotnet.microsoft.com/download/dotnet-core/2.1
+    w_download https://download.visualstudio.microsoft.com/download/pr/73a0ea97-57d6-4263-ac36-ea3a9b373bcf/cd5f7e08e749c1d3468cdae89e4518de/dotnet-runtime-2.1.17-win-x86.exe 706a7f0ad998c3c1b321e89e4bcd6304bef31c95c83eda119e8d4adccccbc915
+
+    w_try_cd "$W_CACHE"/"$W_PACKAGE"
+    w_try "$WINE" dotnet-runtime-2.1.17-win-x86.exe
+
+    case "$W_ARCH" in
```
austin987, line 9459: Not a blocker (and I know you copied this from others) but I'd prefer `if [ "$W_ARCH" = "win64"; then ...`.

src/winetricks
```diff
@@ -9439,6 +9439,34 @@ verify_dotnet48()
...
+    file1="dotnet-runtime-2.1.17-win-x86.exe" \
+    installed_file1="$W_PROGRAMS_WIN/dotnet/dotnet.exe"
+
+load_dotnetcore2()
+{
+    # Official version, see https://dotnet.microsoft.com/download/dotnet-core/2.1
+    w_download https://download.visualstudio.microsoft.com/download/pr/73a0ea97-57d6-4263-ac36-ea3a9b373bcf/cd5f7e08e749c1d3468cdae89e4518de/dotnet-runtime-2.1.17-win-x86.exe 706a7f0ad998c3c1b321e89e4bcd6304bef31c95c83eda119e8d4adccccbc915
+
+    w_try_cd "$W_CACHE"/"$W_PACKAGE"
+    w_try "$WINE" dotnet-runtime-2.1.17-win-x86.exe
```
austin987, line 9457: Generally we try to support quiet mode if possible (particularly since it makes it easier for running the tests). It looks like this supports a /quiet flag (run the installer with `/?` to see). Could you please add support for quiet mode to x86/64 installers?

For example, see https://github.com/Winetricks/winetricks/blob/master/src/winetricks#L12833 et al

Aliervo:
I think I have everything fixed up now. Only way I could get quiet mode working was ``wine <installer>.exe /quiet`` and I couldn't get ``$W_UNATTENDED_SLASH_QUIET`` to inject, so for right now, I threw "/quiet" right in. If there is a better way, let me know.

As for dll overrides, simply installing .Net Core seems to fix many crashes and, more importantly, makes the game's crash reporter work.  There may be a few that I need to add as we troubleshoot some more now that we have better stack traces, but the install alone is enough to have an impact.

As for wine-mono, I've been testing in fresh 64-bit prefixes each time I change the script and don't have it installed as a system package, but we most of us are using Proton to play the game, which I believe does bundle wine-mono.  Does it make a difference script wise?

Aliervo:
Alright. Everything should be good to merge now.

src/winetricks
```diff
@@ -9439,6 +9439,33 @@ verify_dotnet48()
...
+    # Official version, see https://dotnet.microsoft.com/download/dotnet-core/2.1
+    w_download https://download.visualstudio.microsoft.com/download/pr/73a0ea97-57d6-4263-ac36-ea3a9b373bcf/cd5f7e08e749c1d3468cdae89e4518de/dotnet-runtime-2.1.17-win-x86.exe 706a7f0ad998c3c1b321e89e4bcd6304bef31c95c83eda119e8d4adccccbc915
+    unattended_args="/quiet"
+
+    w_try_cd "$W_CACHE"/"$W_PACKAGE"
+    w_try "$WINE" "$file1" ${unattended_args}
+
+    if [ "$W_ARCH" = "win64" ]; then
+      # Also install the 64-bit version
+      file1="dotnet-runtime-2.1.17-win-x64.exe"
```
austin987, line 9461: There's no need to reset the variable here.

Side note: if you got this idea from some documentation, please let me know where, I'd like to update it).

src/winetricks
```diff
@@ -9439,6 +9439,33 @@ verify_dotnet48()
...
+    year="2020" \
+    media="download" \
+    file1="dotnet-runtime-2.1.17-win-x86.exe" \
+    installed_file1="$W_PROGRAMS_WIN/dotnet/dotnet.exe"
+
+load_dotnetcore2()
+{
+    # Official version, see https://dotnet.microsoft.com/download/dotnet-core/2.1
+    w_download https://download.visualstudio.microsoft.com/download/pr/73a0ea97-57d6-4263-ac36-ea3a9b373bcf/cd5f7e08e749c1d3468cdae89e4518de/dotnet-runtime-2.1.17-win-x86.exe 706a7f0ad998c3c1b321e89e4bcd6304bef31c95c83eda119e8d4adccccbc915
+    unattended_args="/quiet"
```
austin987, line 9454: No need, use `$W_UNATTENDED_SLASH_QUIET` instead

Aliervo: I got a little confused by the comment above the `W_UNATTENDED_`s. Fixed.

src/winetricks
Aliervo: So I may have gone a bit rogue here.  `shellcheck` started to throw errors when I tried to call the 64-bit installer `file2` in the metadata, but stopped complaining when I moved the declaration into the if block.  I then noticed that there where no other `$file2` expansions anywhere in the code, so I decided to just reset the variable I already had.

I have noticed that not many verbs use the `$file1` expansion. Should I change both of these to the executable name and avoid this issue entirely?

austin987: It's used by winetricks_is_cached(), which is used to show verbs already have their files downloaded. It only supports one file (currently, though I don't really have plans to expand it, I don't think anyone really uses that feature).

I would say keep file1 in the metadata, and you can (optionally) use it for the 32-bit install. In the 64-bit block, don't set a variable, just do:
```
      w_download ...
      w_try "$WINE" executable.exe ...
```

austin987: Actually, don't worry about it, I've got some changes that I'm about to push that will conflict, so I'll fix this up for you and push.

Thanks for working on this!

austin987:
https://github.com/Winetricks/winetricks/commit/caf8b1bfc018e9b8b78a83c972b6a684b4874711

Thanks again!

PR #1519, opened 2020-04-12
Various speechsdk fixes

### LINKED ISSUE #1517: speechsdk verb for SAPI 5.1 should not restore default WinVer to Windows 7
A rather lengthy analysis here:

https://bugs.winehq.org/show_bug.cgi?id=48919

I've also mentioned it in:

https://bugs.winehq.org/show_bug.cgi?id=43841

In short: 

SAPI 5.1 doesn't exist on Windows Vista+ or later because it's shipped as part of OS.

* SAPI 5.3 = shipped with Windows Vista
* SAPI 5.4 = shipped with Windows 7 or later

There are apps that check the Windows version and automatically assume SAPI 5.3 or later. I've tracked/analyzed one of these in bug 48919 ("Balabolka (TTS app) reports SAPI Error Code 0x80045004 since Wine 2.18 with native SAPI 5.1 via 'winetricks -q speechsdk'").

'Windows XP' should be safe as compromise.

----

Side issue: when using Wine < 2.18, the recipe fails because `c:/windows/syswow64/Speech/Common` doesn't exist before 2.18 which magically fixes apps because due to winetricks abort the Windows version remains at Windows 2000 (used to install SAPI 5.1). The recipes should have a check for older Wine versions and not fail.

----

Follow-up issue: Consider if it's possible to extract SAPI 5.3 from Windows SDK.

gverm:
This should fix #1517. (except for the request for newer speechsdk version, but I think that could/should be a new issue)

I also removed the W_32BIT_DLLS variable as it was the same as W_SYSTEM32_DLLS and only used in speechsdk.

I added a warn for 64bit as it seems only a 32bit dll is installed.

If any of the above is not desirable, let me know.

### DIFF
--- a/src/winetricks
+++ b/src/winetricks
@@ -11645,7 +11645,8 @@ load_speechsdk()
 
     w_override_dlls native sapi
 
-    w_set_winver 'default'
+    # SAPI 5.1 doesn't work on vista and newer
+    w_set_winver winxp
 }
 
 #----------------------------------------------------------------
--- a/src/winetricks
+++ b/src/winetricks
@@ -11618,6 +11618,8 @@ w_metadata speechsdk dlls \
 
 load_speechsdk()
 {
+    w_package_warn_win64
+
     # https://www.microsoft.com/en-us/download/details.aspx?id=10121
     w_download https://download.microsoft.com/download/B/4/3/B4314928-7B71-4336-9DE7-6FA4CF00B7B3/SpeechSDK51.exe 520aa5d1a72dc6f41dc9b8b88603228ffd5d5d6f696224fc237ec4828fe7f6e0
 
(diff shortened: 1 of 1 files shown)


src/winetricks
```diff
@@ -11640,11 +11640,12 @@ load_speechsdk()
...
     # If sapi.dll isn't in original location, applications won't start, see
     # e.g., https://bugs.winehq.org/show_bug.cgi?id=43841
-    w_try ln -s "$W_COMMONFILES_X86/Microsoft Shared/Speech/sapi.dll" "$W_32BIT_DLLS/Speech/Common"
+    mkdir -p "$W_SYSTEM32_DLLS/Speech/Common/"
+    w_try ln -s "$W_COMMONFILES_X86/Microsoft Shared/Speech/sapi.dll" "$W_SYSTEM32_DLLS/Speech/Common"
 
     w_override_dlls native sapi
 
-    w_set_winver 'default'
+    w_set_winver winxp
```
austin987, line 11648: Please add a comment here saying why we need winxp. Otherwise LGTM, thanks!

gverm: @austin987 done!

austin987:
Thanks!

PR #1523, opened 2020-04-15
Add separate mfc verbs

### LINKED ISSUE #1455: Provide other mfc* verbs
It would be nice to have the mfc1* dlls in separate verbs. Mostly for debugging purposes (as to only install/override the things you absolutely need).

I might have a look at this myself if I have time, but anyone can feel free to have a go at it.

gverm:
@austin987 I haven't finished, but this is what I have now, the rest will look similar. The reason I'm creating this draft PR now is so that, if you have time, you can have a look at the approach, so I don't have to rework more than necessary.

Some notes:

- I'm not completely sure what installed_file{1,2,etc} are for, but the old wiki seems to suggest they're used to check if a verb is already installed. This will be a problem for the mfc verbs, since the vcrun* verbs use the mfc dlls as installed_file1. Am I correct here and if so, what would be a good way to fix this?

- I opted against using a separate helper function since it didn't seem worth it, but if you know some more ways to avoid code duplication a little, let me know.

Fixes #1455

### DIFF
--- a/src/winetricks
+++ b/src/winetricks
@@ -11965,6 +11965,21 @@ load_vcrun2003()
     w_try_7z "$W_SYSTEM32_DLLS" "${W_CACHE}/vcrun2003/BZEditW32_1.6.5.exe" "mfc71.dll" "msvcp71.dll" "msvcr71.dll" -y
 }
 
+w_metadata mfc71 dlls \
+    title="Visual C++ 2003 mfc71 library; part of vcrun2003" \
+    publisher="Microsoft" \
+    year="2003" \
+    media="download" \
+    file1="BZEditW32_1.6.5.exe" \
+    installed_file1="$W_SYSTEM32_DLLS_WIN/mfc71.dll"
+
+load_mfc71()
+{
+    w_download_to vcrun2003 https://sourceforge.net/projects/bzflag/files/bzedit%20win32/1.6.5/BZEditW32_1.6.5.exe 84d1bda5dbf814742898a2e1c0e4bc793e9bc1fba4b7a93d59a7ef12bd0fd802
+
+    w_try_7z "$W_SYSTEM32_DLLS" "${W_CACHE}/vcrun2003/BZEditW32_1.6.5.exe" "mfc71.dll" -y
+}
+
 #----------------------------------------------------------------
 
 # Temporary fix for bug 169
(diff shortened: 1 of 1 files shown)


src/winetricks
```diff
@@ -11976,13 +11976,26 @@ w_metadata vcrun2003 dlls \
...
 {
-    # Load the Visual C++ 2003 runtime libraries
     # Sadly, I know of no Microsoft URL for these
-    echo "Installing BZFlag (which comes with the Visual C++ 2003 runtimes)"
     # winetricks-test can't handle ${file1} in url since it does a raw parsing :/
     w_download https://sourceforge.net/projects/bzflag/files/bzedit%20win32/1.6.5/BZEditW32_1.6.5.exe 84d1bda5dbf814742898a2e1c0e4bc793e9bc1fba4b7a93d59a7ef12bd0fd802
-    w_try "$WINE" "$W_CACHE/vcrun2003/${file1}" "$W_UNATTENDED_SLASH_S"
-    w_try cp "$W_PROGRAMS_X86_UNIX/BZEdit1.6.5"/m*71* "$W_SYSTEM32_DLLS"
+
+    w_try_7z "$W_SYSTEM32_DLLS" "${W_CACHE}/mfc71/BZEditW32_1.6.5.exe" "mfc71.dll" "msvcp71.dll" "msvcr71.dll"
```
austin987, line 11983: This drops msvcp71.dll (fyi, I used `find ~/.wine/drive_c/ -type f -exec sha256sum {} \; | sort > /tmp/before` to find that)

austin987:
Hi Gijs,

Yes, installed_file1 (and only 1, there's a FIXME for 2..x} is used to see if a verb is already installed.

There's a similar problem in newer vvcruntimes/dotnet. Some verbs have a workaround (touching a file named ${verb}.workaround, e.g., dotnet48.

src/winetricks
gverm: @austin987 That should be fixed now. I checked both a 64bit and 32bit prefix and all 3 files were put successfully in syswow64 and system32 respectively.

gverm:
@austin987 Everything up until mfc140 (the vcrun2015 one) is added now. For the installed_file1 stuff I just decided to use the mfc*u.dll for the mfc* verbs. This makes sure that at least vcrun can be installed after mfc, the drawback is that mfc can still be installed (once) even if vcrun is already installed, but in my opinion, that doesn't matter much.

PR #1537, opened 2020-04-27
readme: add missing escape backslashes in installing&updating scripts

180254:
When you run the specified installation script in bash, the evaluated `mktemp -d` statement appears in the update_winetricks file. So the update_winetricks file contains e.g. `cd "/tmp/tmp.htySgyRnuY"`.

Such a folder may no longer exist, e.g. the next time you start your computer. I think the goal was to create a new temporary directory every time. For this to happen, the statement must be escaped.


austin987:
Nice find, thanks.

PR #1555, opened 2020-05-23
Fix wmp9 on 64 bit prefix

rodrigo21:
It should be Quiet not quiet


austin987:
Nice find, thanks!

I tweaked the commit message and committed as https://github.com/Winetricks/winetricks/commit/49b22b56802da12d3fc6672a6a5cc11309b36c71

PR #1556, opened 2020-05-24
Update Chinese translations and Adobe Flash Player

hillwoodroc:


austin987:
The flash player commit should be its own pr (with commit message = "flash: update version to ..")

hillwoodroc:
> The flash player commit should be its own pr (with commit message = "flash: update version to ..")
Fixed

PR #1566, opened 2020-06-20
Replace the URL of "binkw32.dll".

Xianguang-Zhou:
The old URL "http://www.down-dll.com/dll/b/__32-binkw32.dll3.0.0.0.zip" is invalid.


austin987:
Thanks, but I had already staged a different fix (using archive.org, so that the current cache is still valid).

PR #1575, opened 2020-07-11
usage,man: add fonts category and synchronize

kparal:
Add missing fonts category to the usage string. Synchronize manpage
commands section with the usage string.


austin987:
The changes themselves are fine, but the commits should be split up/renamed. I.e.,:

commit1: winetricks.1: sync manpage
commit2: winetricks: document missing `fonts list` option

kparal:
Adjusted.

PR #1594, opened 2020-08-24
Use TRUE and FALSE instead of 0 and 1

ferion11:
Simple refractory to use TRUE and FALSE instead of 0 and 1
This will make the logic flow better and cause less confusion with other definitions of other languages.

### DIFF
--- a/src/winetricks
+++ b/src/winetricks
@@ -163,6 +163,10 @@ WINETRICKS_VERSION=20200412-next
 # - Pull Requests: https://github.com/Winetricks/winetricks/pulls
 #--------------------------------------------------------------------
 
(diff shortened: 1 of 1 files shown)


src/winetricks
```diff
@@ -163,6 +163,9 @@ WINETRICKS_VERSION=20200412-next
 # - Pull Requests: https://github.com/Winetricks/winetricks/pulls
 #--------------------------------------------------------------------
```
austin987, line 165: Please add a comment here explaining why.

austin987:
Good idea.

Please change the commit message to 'misc: return TRUE and FALSE instead of 0 and 1'

PR #1595, opened 2020-08-25
misc: SC2250 added braces around variable references

ferion11:
*  Added braces around variable references
From [SC2250](https://github.com/koalaman/shellcheck/wiki/SC2250):
```
If a variable gets called, and there is a string that gets appended to the variable that could get misinterpreted as possibly part of the name of the variable. Then it will not call the right variable.
```
But for me, is that you have a lot of variable that can be confused with others like the combination `${FontSmoothing}` and `${FontSmoothingType}`, so, knowing where the variable name ends, helps in reading the code.

### DIFF
--- a/src/winetricks
+++ b/src/winetricks
@@ -120,11 +120,11 @@ WINETRICKS_VERSION=20200412-next
 #   close curlies ('}') and 'fi' should line up with the matching { or if,
 #   cases indented 4 spaces from 'case' and 'esac'.  For instance,
 #
-#      if test "$FOO" = "bar"; then
+#      if test "${FOO}" = "bar"; then
 #         echo "FOO is bar"
 #      fi
 #
-#      case "$FOO" in
+#      case "${FOO}" in
 #          bar) echo "FOO is still bar" ;;
 #      esac
 #
@@ -147,7 +147,7 @@ WINETRICKS_VERSION=20200412-next
 # Internationalization / localization:
 # - Important or frequently used message should be internationalized
 #   so translations can be easily added.  For example:
-#     case $LANG in
+#     case ${LANG} in
 #         de*) echo "Das ist die deutsche Meldung" ;;
 #         *)   echo "This is the English message" ;;
 #     esac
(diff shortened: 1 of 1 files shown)


austin987:
I haven't tested/reviewed in depth, just glanced. The change itself is (probably) fine (it's a bit big, but obviously it's not trivial/meaningful to split), but if you want to fix this properly, it needs to get added to tests/shell-checks (i.e., add enable=require-variable-braces to shellcheck call) so that it gets enforced on future PRs/commits.

Of course, it will also need to be fixed in other parts of the repo (one commit per file, please).

austin987:
Actually, it would be best to put it in the files themselves:
```
diff --git a/src/winetricks b/src/winetricks
index 216b229c..c0aad73d 100755
--- a/src/winetricks
+++ b/src/winetricks
@@ -1,5 +1,5 @@
 #!/bin/sh
-# shellcheck disable=SC2030,SC2031
+# shellcheck disable=SC2030,SC2031 enable=require-variable-braces
```

ferion11:
> I haven't tested/reviewed in depth, just glanced. The change itself is (probably) fine (it's a bit big, but obviously it's not trivial/meaningful to split), but if you want to fix this properly, it needs to get added to tests/shell-checks (i.e., add enable=require-variable-braces to shellcheck call) so that it gets enforced on future PRs/commits.
> 
> Of course, it will also need to be fixed in other parts of the repo (one commit per file, please).

Instead of `enable=require-variable-braces` I'm using `$ shellcheck --enable=all --exclude=SC2154,SC2249,SC2248 ./src/winetricks` locally, because I thought about doing all the tests after this one (and just use `--enable=all`), fixing them one by one. But after doing this PR, and seeing the size of the diff file, I realized that maybe that kind of change isn't wanted at this point of time (a lot needs to be checked, like simple typos).

And now, I believe that just calling attention to this type of issue, is enough, and we could just close this PR without merge. But, first I would like to hear your opinion on the subject.

ferion11:
> Actually, it would be best to put it in the files themselves:
> 
> ```
> diff --git a/src/winetricks b/src/winetricks
> index 216b229c..c0aad73d 100755
> --- a/src/winetricks
> +++ b/src/winetricks
> @@ -1,5 +1,5 @@
>  #!/bin/sh
> -# shellcheck disable=SC2030,SC2031
> +# shellcheck disable=SC2030,SC2031 enable=require-variable-braces
> ```

By the way, I would recommend not to disable tests globally, but case by case (if necessary), because it can really be a bug in some PR.

PR #1603, opened 2020-09-06
w_wineserver: wait before "wineserver -w" to prevent race condition

ferion11:
Some installation operations take place in multiple steps. But when waiting with just `wineserver -w`, it happens that when the first step (which is usually the main and biggest) is finished, the closing signal is sent to the `wineserver`, but other steps can start right after the first, then creating a race condition. And that happens when installing dotnet48 (and maybe other versions and installers too).

### DIFF
--- a/src/winetricks
+++ b/src/winetricks
@@ -1045,12 +1045,34 @@ w_winepath()
 # It uses minimal buffering, so each line is output immediately
 # and the user can watch progress as it happens.
 
+# Wait to all process that is using the ${1} directory to finish, checking 3 times.
+winetricks_wait_process_using_dir() {
+    _W_verification_dir="${1}"
+    _W_verification_time=7
+    _W_verification_num=3
+
+    i=0 ; while true; do
+        i=$((i+1))
+        sleep "${_W_verification_time}"
+
+        _W_fist_pid="$(lsof -t "${_W_verification_dir}" | head -n 1)"
+        if [ -n "${_W_fist_pid}" ]; then
+            i=0
+            tail --pid="${_W_fist_pid}" -f /dev/null
+            continue
+        fi
+
+        [ "${i}" -lt "${_W_verification_num}" ] || break
+    done
+}
+
 # wrapper around wineserver, to let users know that it will wait indefinitely/kill stuff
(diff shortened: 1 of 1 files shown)


austin987:
What wine version are you seeing this with? I'm unable to reproduce.

ferion11:
> What wine version are you seeing this with? I'm unable to reproduce.

If you want to reproduce, install the 64-bit version of dotnet48 and then right after install a dotnet application in the same script.
Or you can do a stress test on the dotnet installation, checking if there are any processes left behind just after each installation (after the `wineserver -w` exit).
It's related to this bug report: https://bugs.winehq.org/show_bug.cgi?id=49549

I find this in the 32-bit and 64-bit versions of wine-staging v5.11 (and wine-staging v4.21). Although I believe it's present in all versions, as it isn't a bug in Wine but in `winetricks`. It isn't hard to confirm that the `wineserver -w` is being bypassed around 10% of run cases of dotnet48 installations (inside test servers like travis-ci and Github Actions, just show the process list rigth after the installation with a verification).
You can implement a windows application that calls another one right after shutting down too (in a chain). Which in the case of dotnet48, it still compiles a few more classes after the `wineserver -w` exit (leaving the installation incomplete, some more others less).

Many users of my install script: https://github.com/ferion11/LogosLinuxInstaller
had been encountering this issue, and in about 10% of my tests randomly failed to install (testing on both travis-ci and Github Actions):
https://github.com/ferion11/LogosLinuxInstallTests
Until I identified the issue and implemented this solution.

I still have the issue of randomly processes entering into deadlock, but this is another problem, with the bug being in the Wine code. And you already reported it here:
https://bugs.winehq.org/show_bug.cgi?id=48847

ferion11:
I will close this for now, as I'm getting different logs from `winetricks dotnet48` and `winetricks -q dotnet48`, and some strange results with the latest git version too.

austin987:
FWIW, I noticed a race when making prefixes, https://github.com/Winetricks/winetricks/commit/7a05e45518ceb30c7c81ea275863c7b93c6d4a43 may help with that.

PR #1613, opened 2020-09-14
EUupdate.EXE download link workaround

twdragon:
Found a workaround for SourceForge link that is not working now (it leads to the CDN delay page) - the downloaded file with the correct hash uploaded to IPFS persistent storage system (CID `QmT1VxiU7HzcX6BNGiCCtpUr6pxaLDTjTxy8EvPw4X1EaD` EUupdate.EXE).

The root URI is made for the public IPFS directory CID `QmXLYcBn5AKzdwcDwzD4rvNExZCuo7qjNzEQtxoizfyBtw` to preserve filename.

### DIFF
--- a/src/winetricks
+++ b/src/winetricks
@@ -13899,7 +13899,8 @@ load_eufonts()
     # https://www.microsoft.com/en-us/download/details.aspx?id=16083
     # Previously at https://download.microsoft.com/download/a/1/8/a180e21e-9c2b-4b54-9c32-bf7fd7429970/EUupdate.EXE
     # 2020/09/11: https://sourceforge.net/projects/mscorefonts2/files/cabs/EUupdate.EXE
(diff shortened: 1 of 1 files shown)


austin987:
I'm unclear on why this is needed. Could you elaborate on why you need this?

twdragon:
@austin987 I discovered that the original link to the file `EUupdate.EXE` was unaccessible from some countries (for me now in Italy) at the time of this pull request. I downloaded and verified the file via VPN then uploaded it into IPFS exporting the key. For IPFS we can at least consider it accessible persistently.

austin987:
I don't like the idea of using a random 3rd party link to avoid ISP issues..winetricks should (already) fall back to archive.org if the file is unavailable.

Alternatively, if you have torify installed, you can use --torify to download via tor, assuming your ISP doesn't block that ;).

PR #1626, opened 2020-10-03
fixed vc_redist sha256 checksums (2020-09-29)

slaecker:
Microsoft has updated the vc_redist packages on 2020-09-29, see here:

https://docs.microsoft.com/de-de/visualstudio/releases/2019/release-notes

I updated the winetricks checksums so there is no mismatch error anymore


gverm:
@slaecker There's already #1625 for that.

PR #1652, opened 2020-11-26
Add travis job for ppc64le. Install shellcheck

gururajrkatti:
This PR adds CI support for the IBM Power Little Endian (ppc64le) architecture. The idea is to ensure that the builds on this architecture are continuously tested along with the Intel builds (amd64) as this is part of the ubuntu distro on that architecture as well and detecting (and fixing) any issues or failures early would help to ensure that we are always up to date. Since ppc64le doesn't have shellcheck by default, installing it before_install

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,9 @@ language: shell
 os:
   - linux
   - osx
+arch:
+  - amd64
+  - ppc64le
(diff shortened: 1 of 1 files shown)


gururajrkatti:
Thanks !

PR #1669, opened 2020-12-16
vcrun improvements

### LINKED ISSUE #1635: [vcrun2015 / vcrun2017] ucrtbase.dll not getting installed properly
Hey,

The issue has been found using the latest winetricks version (`61509663b824d72daafec6a95986e5b256e19e0e8ec1d2b7c4e89cf52665d63b`).

We tried to find the cause for a crash in [Foundation ](https://store.steampowered.com/app/690830/Foundation/)over at the [Proton issue tracker](https://github.com/ValveSoftware/Proton/issues/2318). After some tinkering & [a tip](https://github.com/Winetricks/winetricks/issues/1631#issuecomment-711906091) by @gverm it became clear that there is an issue with the vcrun2015 & vcrun2017 installers not properly replacing the ucrtbase.dll due to issues with the file permission. There is no hint in the log that the file couldn't be replaced properly.

[This recently removed workaround](https://github.com/Winetricks/winetricks/commit/279bd357cf5baa780807acc436624c22218e3131) for wine bug 30713 (via #1631 ) fixed the issue for some users by manually copy/pasting the dlls into the prefix system32 directory, [but it didn't work for everyone](https://github.com/ValveSoftware/Proton/issues/2318#issuecomment-711980001).

A manual workaround for Foundation is described [here](https://github.com/ValveSoftware/Proton/issues/2318#issuecomment-712030139).

Since proton automatically installs vcrun2015 into the game prefix and the game still crashes, I assumed that the vcrun2015 installer has to be affected as well.

### LINKED ISSUE #1667: [vcrun2017 / vcrun2019] msvcp140_1.dll and msvcp140_2.dll native overrides not set
As it says. Discovered in https://bugs.winehq.org/show_bug.cgi?id=50330

Information about these MS C++ Standard Library support libs:

https://stackoverflow.com/questions/62000672/what-are-the-msvcp140-1-dll-and-msvcp140-2-dll-files-in-the-microsoft-c-runtime

**NOTE:** As of `Wine 6.0-rc2`, only `msvcp140_1` exists but it doesn't harm to set `msvcp140_2` override as well.

https://source.winehq.org/git/wine.git/tree/04ddabfdff644561adf0e36050db1bbc63cad83a:/dlls/msvcp140_1

gverm:
Fixes #1667 and fixes #1635.

To clarify commit 3, the bug mentioned doesn't seem to be the reason why it has to run in XP mode anymore. In Win7 it will refuse to install ucrtbase.dll if it's missing, but it won't replace it if it's not removed before the installer is run.

I split the commits into chucks that made sense for me, but feel free to request more splitting/changes.

### DIFF
--- a/src/winetricks
+++ b/src/winetricks
@@ -12757,6 +12757,8 @@ load_vcrun2015()
             w_try "${WINE}" vc_redist.x64.exe ${W_OPT_UNATTENDED:+/q}
             ;;
     esac
+
+    w_set_winver 'default'
 }
 
 w_metadata mfc140 dlls \
@@ -12834,6 +12836,8 @@ load_vcrun2017()
             w_try "${WINE}" vc_redist.x64.exe ${W_OPT_UNATTENDED:+/q}
             ;;
     esac
+
+    w_set_winver 'default'
 }
 
 #----------------------------------------------------------------
@@ -12887,6 +12891,8 @@ load_vcrun2019()
             w_try "${WINE}" vc_redist.x64.exe ${W_OPT_UNATTENDED:+/q}
             ;;
     esac
+
+    w_set_winver 'default'
 }
 
 #----------------------------------------------------------------
--- a/src/winetricks
+++ b/src/winetricks
@@ -12745,11 +12745,17 @@ load_vcrun2015()
 
     w_set_winver winxp
 
+    # Otherwise ucrtbase doesn't get replaced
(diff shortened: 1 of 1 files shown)


austin987:
Thanks!

The splits make sense to me. As an aside, I wish more of my coworkers contributed to open souce/Wine, because the idea of atomic commits seems to be lost of some of them ;).

PR #1673, opened 2020-12-20
sapi: new verb

gverm:
This takes care of the last request in #1517.
It should also help to run apps that need SAPI in Win 7 mode.

### DIFF
--- a/src/winetricks
+++ b/src/winetricks
@@ -11942,6 +11942,38 @@ load_richtx32()
 
 #----------------------------------------------------------------
 
+w_metadata sapi dlls \
+    title="MS Speech API" \
+    publisher="Microsoft" \
+    year="2011" \
+    media="download" \
+    conflicts="speechsdk" \
+    file1="../win7sp1/windows6.1-KB976932-X86.exe" \
+    installed_file1="${W_SYSTEM32_DLLS_WIN}/sapi.dll"
+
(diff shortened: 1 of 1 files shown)


gverm:
I've converted this to a draft for now as it seems wine is unable to load the native dll, but I didn't see it before because I used native,builtin instead of native only.

gverm:
This should be good to go now. Apparently this version of native SAPI needs to be in system32/syswow64. I changed the verb to remove the directory where builtin SAPI is located. I also put the native dll into system32/syswow64 now and changed the override to native only.

src/winetricks
```diff
@@ -11919,6 +11919,37 @@ load_richtx32()
...
+    publisher="Microsoft" \
+    year="2011" \
+    media="download" \
+    conflicts="speechsdk" \
+    file1="../win7sp1/windows6.1-KB976932-X86.exe" \
+    installed_file1="${W_SYSTEM32_DLLS_WIN}/sapi.dll"
+
+load_sapi()
+{
+    for stub in "${W_SYSTEM32_DLLS}/Speech" "${W_SYSTEM64_DLLS}/Speech"; do
```
austin987, line 11933: Please add a comment here about why this is needed.

gverm: @austin987 Done.

PR #1678, opened 2020-12-30
Directplay improvements

gverm:
Stubs for these dlls were adding in the staging patchset dxdiag-new-dlls. This was causing issues with Total Annihilation and Moonbase Commander, so installing the native versions in the directplay verb seems like a good idea.


austin987:
Thanks!

Issue #1654, opened 2020-11-28
Why use Zenity on KDE Plasma instead of kdialog?


ryu-ketsueki:
This is more of an annoyance because Steam forces the instalation of zenity for their dialogue windows and does not use kdialog in any way. I imagine that for winetricks offer an option to use kdialog, shouldn't it be priorized in KDE Plasma? It integrates better. I can't find anything on command line that gives the option to run winetricks with kdialog if zenity is installed. Zenity will always be installed if Steam is installed. It's a hard dependency.

austin987:
The logic isn't that smart, it just looks to see which of those helpers is in $PATH, and it seems zenity gets checked for first.

The logic could be improved, i.e., similar to what is done for WINETRICKS_DOWNLOADER, where one can specify aria2c/curl/fetch/wget, to override the selection order.

Patches welcome. I don't have time/motivation for GUI issues.

Ruedii:
The main reason is that Zenity is widely used in almost every distro, and hence likely installed.  

Allowing the use of an external GUI plugin script might be a better way to handle this, though.

hikagura99:
In openSUSE it seems to use Kdialog, i think it depends on the distro

ryu-ketsueki:
> In openSUSE it seems to use Kdialog, i think it depends on the distro

I am using openSUSE Tumbleweed and it is prioritizing Zenity over Kdialog if Zenity is installed because of Steam

austin987:
As I said (https://github.com/Winetricks/winetricks/issues/1654#issuecomment-735284624), it's based on what you have installed. If you have both, zenity will get used.

ryu-ketsueki:
> As I said ([#1654 (comment)](https://github.com/Winetricks/winetricks/issues/1654#issuecomment-735284624)), it's based on what you have installed. If you have both, zenity will get used.

I know, I only wanted to point out for @austin987 that it's not distro dependent.

ThisNekoGuy:
Even after that supposed fix (https://github.com/Winetricks/winetricks/pull/1680), I get this inconsistency as well.

JoseskVolpe:
I agree it should priorize kdialog over zenity if the user is using KDE

ThisNekoGuy:
Waiting on https://github.com/Winetricks/winetricks/pull/1867 to fix this now; I found the issue and fixed it.

austin987:
[...]

PR #1680, opened 2021-01-03
WIP: ability to switch between zenity and kdialog

mscatto:
Hello, after seeing an open issue about this(#1654), I've decided to create a patch.
As it is now, I've changed the --gui flag to accept an argument as well as keep it's previous behaviour intact (--gui without arguments meaning using zenity).
It is also defaulting to kdialog when using kde by checking the XDG_CURRENT_DESKTOP env var.

I'm not 100% sure about the last point though, given how kdialog seemed to be advised against.

### DIFF
--- a/src/winetricks
+++ b/src/winetricks
@@ -3458,24 +3458,32 @@ winetricks_early_wine_arch()
 
 winetricks_detect_gui()
 {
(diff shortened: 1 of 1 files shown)


austin987:
I manually rebased/squashed it and committed it as https://github.com/Winetricks/winetricks/commit/86978c45c36f9cb17ba98ddb6ab4d1d4644da6a5; thanks for working on this.

PR #1685, opened 2021-01-13
steam: disable libglesv2

gverm:
This is necessary because otherwise the Store and Library tabs will be black. It "works" with libglesv2 diabled. I say "works" because it's functional enough to be useful, but still suffers from flickering/repainting issues.

### DIFF
--- a/src/winetricks
+++ b/src/winetricks
@@ -16355,6 +16355,10 @@ load_steam()
     if w_workaround_wine_bug 22053 "Disabling gameoverlayrenderer to prevent game crashes on some machines."; then
         w_override_dlls disabled gameoverlayrenderer
     fi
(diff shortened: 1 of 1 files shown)


PietJankbal:
Hi Gijs, just curious: does adding "--disable-gpu" to the command line  work for Steam to get rid of black window? Might be less "drastic" soltion (if it works at all, cannot test myself atm)

gverm:
@PietJankbal I tried and it doesn't seem to work for me. I thought doing it this way was nice anyway because we have a reference to a bug report and people can remove the override if they want to.

Gcenx:
I can confirm this works, it gotten Steam working again even in my CX builds

PR #1691, opened 2021-01-29
DirectDrawRenderer/renderer updates

gverm:
I think it's time to deprecate ddr=*. It doesn't work properly anymore ("opengl" isn't a valid value anymore, it was changed to "gl"). The renderer key also supports more values, namely "no3d" (alias of "gdi") and "vulkan", which will become more important in the future.

EDIT + Offtopic: Austin, how would you feel about getting rid of the case in winetricks_set_wined3d_var()? Seems a little unnecessary to me, since w_metadata will already make sure the args are correct. Mostly asking because UseGLSL was deprecated in 5.0 and will be removed in 3.1, so I have to touch that code again soon anyway.

### DIFF
--- a/src/winetricks
+++ b/src/winetricks
@@ -22078,7 +22078,7 @@ winetricks_set_wined3d_var()
         enable*) arg=enabled;;
         hard*) arg=hardware;;
         repack) arg=repack;;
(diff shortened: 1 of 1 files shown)


austin987:
I don't have strong feelings (I haven't had a chance to look too deep, and it's been a while since I looked around that area), but IIRC it was there to support enable/enabled?

Not sure it's worth the extra complexity for that alone, mostly throwing it out there as an FYI / reminder to myself for later.

PR #1703, opened 2021-02-11
Update sourcehansans font to 2.002R

jack6th:


austin987:
Could you please fix the commit message to reflect the actual change? I.e., something like 'sourcehansans: update to 2.002R'

PR #1707, opened 2021-02-21
Update dxvk verbs

bobwya:
@austin987 

I'd probably like to drop the verb:

`dxvk_master`

as it's clear that Joshua isn't really bothered about resurrecting his CI builds.

I'm not sure this verb is terribly useful now anyway...
**dxvk** is very stable these days anyway!

Is OK just to nuke it?  :-)

Bob


austin987:
Yes, please do (also remove it from tests/winetricks-tests blacklist)

PR #1726, opened 2021-03-27
steam: update sha265sum

gverm:


gverm:
I added a commit to make sure the steam verb also calls corefonts, because otherwise the steam store and library won't work.

PR #1738, opened 2021-04-06
Manually install ucrtbase in vcrun* verbs instead of deleting it

### LINKED ISSUE #1736: vcrun2015, 2017 and 2019 are broken (ucrtbase is removed prior to running installer)
Some change (can't seem to be able to bisect because of unrelated problems) after wine-6.5 is breaking the vcrun verbs that remove ucrtbase before running the installer.

According to Julliard (on IRC) this is to be expected with the recent loader changes, so we'll have to find another way to fix it.

I would suggest instead of removing ucrtbase, extracting it from the setup and replacing builtin before running the setup.

gverm:
Fixes #1736


austin987:
Thanks!

PR #1750, opened 2021-04-24
Update vcrun2010 hashes

gverm:
No idea why MS decided to change these old installers, but here we are...

### DIFF
--- a/src/winetricks
+++ b/src/winetricks
@@ -12591,7 +12591,9 @@ w_metadata vcrun2010 dlls \
 load_vcrun2010()
 {
     # See https://www.microsoft.com/en-us/download/details.aspx?id=5555
-    w_download https://download.microsoft.com/download/5/B/C/5BC5DBB3-652D-4DCE-B14A-475AB85EEF6E/vcredist_x86.exe 8162b2d665ca52884507ede19549e99939ce4ea4a638c537fa653539819138c8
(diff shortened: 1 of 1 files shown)


src/winetricks
```diff
@@ -12591,7 +12591,8 @@ w_metadata vcrun2010 dlls \
 load_vcrun2010()
 {
     # See https://www.microsoft.com/en-us/download/details.aspx?id=5555
-    w_download https://download.microsoft.com/download/5/B/C/5BC5DBB3-652D-4DCE-B14A-475AB85EEF6E/vcredist_x86.exe 8162b2d665ca52884507ede19549e99939ce4ea4a638c537fa653539819138c8
+    # 2021/04/24: 8162b2d665ca52884507ede19549e99939ce4ea4a638c537fa653539819138c8
```
austin987, line 12594: The comment has the old sha, not the updated one? (But please include the original sha as a comment, as well), i.e.,:
# originally, 8162b2d665ca52884507ede19549e99939ce4ea4a638c537fa653539819138c8
# 2021/04/24: 31d32fa39d52cac9a765a43660431f7a127eee784b54b2f5e2af3e2b763a1af8

same for below.

gverm: Done. For some reason I was confused and thought the previous hash was put after the date it was changed.

PR #1760, opened 2021-05-08
ev3 verb change link to archive.org

Botspot:
The original download site for Lego Mindstorms EV3 Home Edition is down. I've changed the link to archive.org.
Strangely enough, the original download had a sha256 of `c47341f08242f0f6f01996530e7c93bda2d666747ada60ab93fa773a55d40a19`, but this archive.org version has a sha256 of `c2e65d3ca09f71d52931c5a0ff7cb0554b86b30ae80d1c61167dd68893a8237f`.
I have no explanation for why the shasum changed, but this discrepancy is consistent and the exe does launch fine.

Sidenote: Where's the best place to get help for this ev3 verb? I've tried various combinations for days to get it working. Right now there's a [$30 bounty](https://github.com/ptitSeb/box86/issues/296) on it.

### DIFF
--- a/src/winetricks
+++ b/src/winetricks
@@ -14901,7 +14901,8 @@ load_ev3()
 
     # 2016/03/22: LMS-EV3-WIN32-ENUS-01-02-01-full-setup.exe c47341f08242f0f6f01996530e7c93bda2d666747ada60ab93fa773a55d40a19
 
-    w_download http://esd.lego.com.edgesuite.net/digitaldelivery/mindstorms/6ecda7c2-1189-4816-b2dd-440e22d65814/public/LMS-EV3-WIN32-ENUS-01-02-01-full-setup.exe c47341f08242f0f6f01996530e7c93bda2d666747ada60ab93fa773a55d40a19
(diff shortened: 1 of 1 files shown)


CleanMachine1:
Considering the sha256sum has changed, there is a good chance that either it has broken something, or worse. The first letter of it being a c is just coincidence.

It should be looked into rather than providing a potentially dangerous exe

PietJankbal:
This $30 bounty made me curious ,hahaha, so i gave it a try (just in wine on linux ). 

I got it running , but the installer got a strange crash in an unimplemented function. Not sure if a virus is involved, but it is weird anyway, looks a bit suspicious. I`ll send a patch (and open bug) for that crash anyway, see if gets accepted.

Furthermore needed ie8 and silverlight and i installed dotnet48; Then the program starts and i can click things; not sure if it`s fully functional as i don`t know the program

austin987:
As @CleanMachine1 said, this PR isn't just changing the URL, it's getting an entirely different binary (though I can't endorse reverse engineering it).

The proper place to get support for it would be https://forum.winehq.org/

Updating the sha isn't necessarily 'wrong', though if it's being updated, it would make sense to update it to whatever the authoritative source currently provides (and adjusting the verb accordingly to keep it working).

Botspot:
@austin987, please consider alternatives regarding this PR. Lego's hosting for this software has been down. Archive.org seems like the only option for the ev3 verb to work at all.

Issue #1593, opened 2020-08-20
IE8 is broken since wine-5.7


DarkShadow44:
Wine-5.7 (commit 0a12f258690b46f996b4890837ffdb4cea94ba70) changed the way fonts are loaded.
They now go through "dosdevices/z:", but there is a workaround that renames that to "dosdevices/z:.bak_wt". The result is, that the window is infinitely small and can't be used - also breaking silent install.

We could either fix this in winetricks, unless you think that fonts should not be read through "z:". What do you say?

alessivs:
> https doesn't work for some reason

I have a WIP fix for this, for which I will open a PR in the next days. It requires a new Windows component (msi installer). For guidance, should this be in a new prefix name or inside ie8?

austin987:
What's the component? Does it have any value on its own or is it an IE plugin/service pack/hotfix?

alessivs:
> What's the component? Does it have any value on its own or is it an IE
> plugin/service pack/hotfix?

Two hotfixes:

1. Add support for TLS 1.1 and 1.2 (KB4019276) - comes from Win XP POS
edition

2. Cumulative security update for Internet Explorer 8 (KB4230450)

Plus two registry entries that enable control of the feature from IE 8
Internet Settings.

I currently don't remember why 2 was necessary, but I will revisit.

As far as I remember they weren't available for x64.

It does not have (much) value on its own, since it augments IE 8 MSHTML
or the IE 8 application, and besides the installer it doesn't provide a GUI.

austin987:
I'd say a separate verb, A) so users with existing installs can install it, and B) 64-bit would be added by #1716 (when I get around to  testing/reviewing).

alessivs:
This issue seems solved by Wine 5.15 to Wine 6.0, the current stable. More details after further testing.

alessivs:
I can't tell with certainty, but some font issues could be related to the interaction of Wine and the packaging of 32-bit freetype2. A given Wine edition can experience the tiny installer window issue with an older freetype2, only to vanish with an update to the library. Sorry for the lack of details, as this comes from memory, and it's been tricky to try out different freetype packages without breaking the desktop.

Anyhow, I experienced font-related issues between Wine 5.7 and 5.14 as packaged by Arch, where IE8 installs but doesn't finish launching. The problem is absent in Wine 5.15 up to 6.4 (tried every Wine in between), where IE8 does function.

[...]

PR #1761, opened 2021-05-09
ie8_tls12: new verb

alessivs:
This verb adds TLS 1.1 and TLS 1.2 support for IE 8, configurable through the UI in *Tools : Internet Options : Advanced*.

image

Changes to the configuration affect the registry value at:

*  `HKCU\Software\Microsoft\Windows\CurrentVersion\Internet Settings\SecureProtocols`

Any configuration made prior to installing this verb influences the initial value, so the verb doesn't concern itself with modifying the registry to enable the new protocols without user interaction.

Also, the UI (IE 8 menu bar) isn't visible with the `Windows 7` setting, so this verb doesn't issue a final `w_set_winver 'default'` invocation. FWIW the menu bar isn't the only part of IE's UI that breaks with that setting.

**Before enabling both protocols**

image

**After enabling both protocols**

image

**Added:** As discussed in #1593 ([comment](https://github.com/Winetricks/winetricks/issues/1593#issuecomment-822183828))

### DIFF
--- a/src/winetricks
+++ b/src/winetricks
@@ -10773,6 +10773,44 @@ load_ie8_kb2936068()
 
 #----------------------------------------------------------------
 
+w_metadata ie8_tls12 dlls \
+    title="TLS 1.1 and 1.2 for Internet Explorer 8" \
+    publisher="Microsoft" \
+    year="2017" \
+    media="download" \
+    file1="windowsxp-kb4019276-x86-embedded-enu_3822fc1692076429a7dc051b00213d5e1240ce3d.exe" \
+    file2="ie8-windowsxp-kb4230450-x86-embedded-enu_d8b388624d07b6804485d347be4f74a985d50be7.exe" \
+    installed_file1="c:/windows/KB4230450-IE8.log"
+
+load_ie8_tls12()
+{
+    w_package_unsupported_win64
+    w_call ie8
+    w_set_winver winxp
+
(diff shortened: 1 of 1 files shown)


src/winetricks
```diff
@@ -10773,6 +10773,44 @@ load_ie8_kb2936068()
...
+    w_call ie8
+    w_set_winver winxp
+
+    "${WINE}" reg add "HKLM\\System\\WPA\\PosReady" /v Installed /t REG_DWORD /d 0001 /f
+
+    w_download http://download.windowsupdate.com/c/msdownload/update/software/updt/2017/10/windowsxp-kb4019276-x86-embedded-enu_3822fc1692076429a7dc051b00213d5e1240ce3d.exe 381abded5dd70a02bd54d4e8926e519ca6b306e26cbf10c45bbf1533bf57a026
+
+    w_try_cd "${W_CACHE}/${W_PACKAGE}"
+
+    # Avoid permanent hang in attended mode; avoid long pause in unattended mode
```
austin987, line 10797: Is there a bug report for that?

alessivs: Not from me. I did notice `w_workaround_wine_bug` in Winetricks, but I'm unsure how to proceed with any report against Wine due to possible applicability of the *DLL overrides* rule at [this page](https://wiki.winehq.org/Bugs).

austin987: Sure, a fair concern. IE is definitely a corner case for that policy. An appdb report/winetricks bug would also be acceptable. While not everyone cares, it's useful for users to have somewhere to follow for more info/discussion.

Issue #1770, opened 2021-06-03
winetricks vcrun2019 fails.  [Note: SHA256 *does* match, 64-bit prefix]


GH-A-CC:
[consolelog.txt](https://github.com/Winetricks/winetricks/files/6594835/consolelog.txt)

GH-A-CC:
`[PUA@blackbird ~]$ winetricks vcrun2019
Executing mkdir -p /home/PUA
------------------------------------------------------
warning: You are using a 64-bit WINEPREFIX. Note that many verbs only install 32-bit versions of packages. If you encounter problems, please retest in a clean 32-bit WINEPREFIX before reporting a bug.
------------------------------------------------------
Using winetricks 20210206-next - sha256sum: e5457a7a984969443119a4c0c1a01607fc8bf8adca114f265669402b24b960b8 with wine-6.9 (Staging) and WINEARCH=win64
Executing w_do_call vcrun2019
Executing mkdir -p /home/PUA
------------------------------------------------------
warning: You are using a 64-bit WINEPREFIX. Note that many verbs only install 32-bit versions of packages. If you encounter problems, please retest in a clean 32-bit WINEPREFIX before reporting a bug.
------------------------------------------------------
Executing load_vcrun2019 
Using native,builtin override for following DLLs: api-ms-win-crt-private-l1-1-0 api-ms-win-crt-conio-l1-1-0 api-ms-win-crt-heap-l1-1-0 api-ms-win-crt-locale-l1-1-0 api-ms-win-crt-math-l1-1-0 api-ms-win-crt-runtime-l1-1-0 api-ms-win-crt-stdio-l1-1-0 api-ms-win-crt-time-l1-1-0 atl140 concrt140 msvcp140 msvcp140_1 msvcp140_2 msvcr140 ucrtbase vcomp140 vcruntime140
Executing wine regedit C:\windows\Temp\override-dll.reg
0208:err:module:import_dll Library ucrtbase.dll (which is needed by L"C:\\windows\\syswow64\\sechost.dll") not found
0208:err:module:import_dll Library sechost.dll (which is needed by L"C:\\windows\\syswow64\\advapi32.dll") not found
0208:err:module:import_dll Library ucrtbase.dll (which is needed by L"C:\\windows\\syswow64\\advapi32.dll") not found
0208:err:module:import_dll Library advapi32.dll (which is needed by L"C:\\windows\\syswow64\\shlwapi.dll") not found
[...]

PR #1773, opened 2021-06-06
vcrun2019 updates

gverm:
For now this only contains a commit to update the hashes, but I can change it to include a fix for #1770 once we figure out a solution for it.


mkaito:
This doesn't really fix the problem, does it? By defining the hash, we define a specific version of the installer that we want to use. A version that does not work. So what's the point of merging this exactly?

gverm:
> This doesn't really fix the problem, does it? By defining the hash, we define a specific version of the installer that we want to use. A version that does not work. So what's the point of merging this exactly?

While it does not fix all the issues, it fixes the first user-visible problem. This will prevent people from opening an issue for that issue at least. (shasum bugs get reported far more/easier than other problems)

We still need to decided how to fix/what to do about #1770.

Issue #1579, opened 2020-07-23
Idea for installing old pdh.dll for NT 4.0


WheezyE:
Hi,

I was recently getting [VARA 4.0.3](https://rosmodem.wordpress.com/) (released July 1, 2020) to work in Wine 5.0.1 - which is part of [Winlink Express](https://winlink.org/WinlinkExpress) (aka RMS Express).  I needed pdh, which winetricks installed dutifully with "winetricks pdh". But I found that I actually was getting errors with this version of pdh.dll and needed an older version of pdh to get VARA working - then it worked perfectly with no wine errors.  The old Win NT dll I used is downloadable direct from Microsoft.  I don't know how common of a use-case this is, but I wrote this command to auto-install it:

```
sudo apt install unzip -y && cd ~/Downloads && wget http://download.microsoft.com/download/winntsrv40/update/5.0.2195.2668/nt4/en-us/nt4pdhdll.exe && unzip nt4pdhdll.exe && cp pdh.dll ~/.wine/drive_c/windows/system32
```

I didn't end up needing to set the dll override in winecfg, but I'm sure it wouldn't hurt - and copying the code segment in winetricks from the other pdh to set the dll override would probably not take too much doing.

Anyway, I just wanted to submit the info in case it was helpful, or if you wanted to implement like a "pdhNT4" command or something.

--------------------------------------------
EDIT:
The error message I get with VARA (when the "winetricks pdh" dll is installed) before VARA crashes is: 
```
0009:fixme:advapi:RegisterEventSourceW ((null),L"PDH"): stub
wine: Call from 0x7bc73f39 to unimplemented function ADVAPI32.dll.PerfEnumerateCounterSet, aborting
wine: Unimplemented function ADVAPI32.dll.PerfEnumerateCounterSet called at address 7BC73F39 (thread 0009), starting debugger...
```
If you want to try to replicate the error: I installed these components 
`winetricks -q dotnet35sp1 corefonts vb6run vcrun2015 sound=alsa win7 ie6`
(and mono and gecko) and the old (NT 4) pdh.dll to get VARA to run.

austin987:
Feel free to s

PR #1781, opened 2021-06-17
pdh_nt4: new verb

WheezyE:
We discussed adding a verb to install the older WinNT 4.0 version of pdh.dll since it helps compatibility with some programs like VARA (part of Winlink / RMS Express).  I got the green light from austin987 to add it ( https://github.com/Winetricks/winetricks/issues/1579 ).  This is my first edit to winetricks, and I'm new to bash, so let me know if this implementation looks strange or isn't standard.
Note: I created an earlier pull request, but realized I had typos in it, so this is a new PR

### DIFF
--- a/src/winetricks
+++ b/src/winetricks
@@ -12934,18 +12934,15 @@ load_vcrun2019()
     # 2021/04/13: 14563755ac24a874241935ef2c22c5fce973acb001f99e524145113b2dc638c1
     # 2021/06/06: 91c21c93a88dd82e8ae429534dacbc7a4885198361eae18d82920c714e328cf9
 
-    w_override_dlls native,builtin api-ms-win-crt-private-l1-1-0 api-ms-win-crt-conio-l1-1-0 api-ms-win-crt-heap-l1-1-0 api-ms-win-crt-locale-l1-1-0 api-ms-win-crt-math-l1-1-0 api-ms-win-crt-runtime-l1-1-0 api-ms-win-crt-stdio-l1-1-0 api-ms-win-crt-time-l1-1-0 atl140 concrt140 msvcp140 msvcp140_1 msvcp140_2 msvcr140 ucrtbase vcomp140 vcruntime140
+    w_warn "ucrtbase.dll is no longer included in vcrun2019. For details see: https://github.com/Winetricks/winetricks/issues/1770"
+
+    w_override_dlls native,builtin api-ms-win-crt-private-l1-1-0 api-ms-win-crt-conio-l1-1-0 api-ms-win-crt-heap-l1-1-0 api-ms-win-crt-locale-l1-1-0 api-ms-win-crt-math-l1-1-0 api-ms-win-crt-runtime-l1-1-0 api-ms-win-crt-stdio-l1-1-0 api-ms-win-crt-time-l1-1-0 atl140 concrt140 msvcp140 msvcp140_1 msvcp140_2 msvcr140 vcomp140 vcruntime140
 
     w_download https://aka.ms/vs/16/release/vc_redist.x86.exe 91c21c93a88dd82e8ae429534dacbc7a4885198361eae18d82920c714e328cf9
 
     if w_workaround_wine_bug 50894 "Working around failing wusa.exe lookup via C:\windows\SysNative"; then
         w_set_winver winxp
     fi
-
-    # Setup will refuse to install ucrtbase because builtin's version number is higher, so manually replace it
-    # See https://bugs.winehq.org/show_bug.cgi?id=46317
-    w_try_cabextract --directory="${W_TMP}/win32"  "${W_CACHE}"/vcrun2019/vc_redist.x86.exe -F 'a10'
-    w_try_cabextract --directory="${W_SYSTEM32_DLLS}" "${W_TMP}/win32/a10" -F 'ucrtbase.dll'
 
     w_try_cd "${W_CACHE}"/"${W_PACKAGE}"
     w_try "${WINE}" vc_redist.x86.exe ${W_OPT_UNATTENDED:+/q}
@@ -12969,9 +12966,6 @@ load_vcrun2019()
             w_override_dlls native,builtin vcruntime140_1
 
             w_download https://aka.ms/vs/16/release/vc_redist.x64.exe a1592d3da2b27230c087a3b069409c1e82c2664b0d4c3b511701624702b2e2a3
-            # Also replace 64-bit ucrtbase.dll
-            w_try_cabextract --directory="${W_TMP}/win64"  "${W_CACHE}"/vcrun2019/vc_redist.x64.exe -F 'a10'
-            w_try_cabextract --directory="${W_SYSTEM64_DLLS}" "${W_TMP}/win64/a10" -F 'ucrtbase.dll'
             w_try "${WINE}" vc_redist.x64.exe ${W_OPT_UNATTENDED:+/q}
             ;;
     esac
--- a/src/winetricks
+++ b/src/winetricks
@@ -22173,6 +22173,29 @@ load_rtlm()
 {
     winetricks_set_wined3d_var RenderTargetLockMode "$1"
 }
+
+#----------------------------------------------------------------
+
+w_metadata set_mididevice settings \
+    title="Set MIDImap device to the value specified in the MIDI_DEVICE environment variable"
+
+load_set_mididevice()
+{
+    if [ -z "${MIDI_DEVICE}" ]; then
+        MIDI_DEVICE=$(w_question "Please specify MIDImap device: ")
+        [ -z "${MIDI_DEVICE}" ] && w_die "Please specify device in MIDI_DEVICE environment variable."
+    fi
+
+    echo "Setting MIDI device to \"${MIDI_DEVICE}\""
+    cat > "${W_TMP}"/set-mididevice.reg <<_EOF_
+REGEDIT4
+
+[HKEY_CURRENT_USER\Software\Microsoft\Windows\CurrentVersion\Multimedia\MIDIMap]
+"CurrentInstrument"="${MIDI_DEVICE}"
+_EOF_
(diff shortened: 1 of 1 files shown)


src/winetricks
```diff
@@ -11466,6 +11466,37 @@ load_pdh()
...
+}
+
+#----------------------------------------------------------------
+
+w_metadata pdh_nt4 dlls \
+    title="MS pdh.dll (Performance Data Helper); WinNT 4.0 Version" \
+    publisher="Microsoft" \
+    year="1997" \
+    media="download" \
+    file1="../pdh_nt4/nt4pdhdll.exe" \
```
austin987, line 11481: This should just be 'nt4pdhdll.exe'

src/winetricks
```diff
@@ -11466,6 +11466,37 @@ load_pdh()
...
+    installed_file1="${W_SYSTEM32_DLLS_WIN}/pdh.dll"
+
+load_pdh_nt4()
+{
+    #w_download_to pdh_nt4 http://download.microsoft.com/download/winntsrv40/update/5.0.2195.2668/nt4/en-us/nt4pdhdll.exe a0a45ea8f4b82daaebcff7ad5bd1b7f5546e527e04790ca8c4c9b71b18c73e32
```
austin987, line 11486: You can remove this.

src/winetricks
```diff
@@ -11466,6 +11466,37 @@ load_pdh()
...
+    w_download_to pdh_nt4 https://web.archive.org/web/20060826233132/http://download.microsoft.com/download/winntsrv40/update/5.0.2195.2668/nt4/en-us/nt4pdhdll.exe a0a45ea8f4b82daaebcff7ad5bd1b7f5546e527e04790ca8c4c9b71b18c73e32
+
+    w_try_unzip "${W_TMP}/${W_PACKAGE}" "${W_CACHE}/${W_PACKAGE}"/nt4pdhdll.exe
+    w_try cp "${W_TMP}/${W_PACKAGE}/pdh.dll" "${W_SYSTEM32_DLLS}/pdh.dll"
+
+    if [ "${W_ARCH}" = "win64" ]; then
+        echo "There is no 64-bit version of the WinNT 4.0 pdh.dll. Please either use a 32-bit wineprefix or try 'winetricks pdh' instead."
```
austin987, line 11493: This should be w_warn (otherwise it won't show for GUI users).

src/winetricks
```diff
@@ -11466,6 +11466,37 @@ load_pdh()
...
+    fi
+
+    w_override_dlls native,builtin pdh
+    
+    # We just installed the older version of pdh.dll and told winetricks that it was installed by logging it into winetricks.log.  If we want to be able to switch back and forth between the older (pdh_nt4) and the newer (pdh) pdh.dll's in the future, then we must tell winetricks to invalidate any log files of the version of pdh.dll that we just overwrote so that winetricks knows that that version is not installed anymore.
+    sed -i 's+\<pdh\>+_pdh+g' "${WINEPREFIX}/winetricks.log"
```
austin987, line 11499: The comment is _way_ too long (break it over multiple lines, if really needed), but I don't see the need. winetricks won't block you from installing pdh after this (if you intended that, you need to do add a conflicts to the metadata).

However, i don't think munging the install log is appropriate, but maybe I'm missing something. Why do you want to do that?

austin987:
Also, fyi, you don't need to close/open new PRs every time. You can just edit your current commit locally (i.e., `git rebase -i` and then force push to update the PR (`git push -f`).

src/winetricks
WheezyE: Thanks!  I was wondering, is it more preferable to download from internet archive or straight from Microsoft?

I saw a bunch of other winetricks downloads being changed to internet archive links instead of direct from Microsoft, though I also noticed that sometimes packages (like win7sp1/windows6.1-KB976932-X86.exe) still download straight from Microsoft.

src/winetricks
WheezyE: Thanks for this tip too. I'll change that when I get home too.

src/winetricks
WheezyE: I was wondering if messing with the log would be poor coding practices.  I'll check out the conflicts metadata option and break up or shorten the comment.

The intended use for editing the winetricks.log is so that users can install pdh, then install pdh_nt4 (to try that one instead to see if their application works better), then install pdh again if they want to switch back to the new pdh.dll again.  
For example: Without this line, if we first install pdh, then winetricks will know that pdh is installed.  Then if we later install pdh_nt4, winetricks will think pdh (new pdh.dll) is still installed AND pdh_nt4 (old pdh.dll) is also installed alongside - when in reality the pdh_nt4 (old pdh.dll) installation overwrote pdh (new pdh.dll).  If a user wants to switch back to pdh (new pdh.dll), winetricks will say that pdh is already installed.

WheezyE:
Thanks for all these comments!  This is super helpful and I'll get working on this as soon as I can.  I'm heading out of town for the next couple days, but if I don't get a chance to make these edits tonight, I'll take a look around Wednesday when I'm back.

src/winetricks
austin987: The original source (so microsoft.com in this case) is preferred, but they have lately been removing downloads for older windows versions, so for those downloads we've moved to archive.org.

src/winetricks
WheezyE: Thanks for the tip about `conflicts`, I've added that instead of the sed editing of winetricks.log

WheezyE:
Just got back and had a chance to make these changes.  (This ended up being easier than I thought - sorry for the long delay).  Let me know what you think!

src/winetricks
```diff
@@ -11470,6 +11471,31 @@ load_pdh()
...
+    publisher="Microsoft" \
+    year="1997" \
+    media="download" \
+    conflicts="pdh" \
+    file1="nt4pdhdll.exe" \
+    installed_file1="${W_SYSTEM32_DLLS_WIN}/pdh.dll"
+
+load_pdh_nt4()
+{
+    w_download_to pdh_nt4 http://download.microsoft.com/download/winntsrv40/update/5.0.2195.2668/nt4/en-us/nt4pdhdll.exe a0a45ea8f4b82daaebcff7ad5bd1b7f5546e527e04790ca8c4c9b71b18c73e32
```
austin987, line 11485: You can use w_download directly (then remove yhe

austin987:
Much better 🙂

Please adjust the w_download call, and squash the commits, and this should be good to go.

WheezyE:
Just wanted to check in to see if there was anything else I can do to help merge this little guy.  Thank you again for all your help

austin987 (commented):
Sorry, I had made this comment and apparently never submitted the review :(

WheezyE:
No worries :) and sorry for my slow reply too - busy week again. I'll move the command and then test it today or tomorrow.  Thank you again!!!

src/winetricks
```diff
@@ -11470,6 +11471,31 @@ load_pdh()
...
+load_pdh_nt4()
+{
+    w_download http://download.microsoft.com/download/winntsrv40/update/5.0.2195.2668/nt4/en-us/nt4pdhdll.exe a0a45ea8f4b82daaebcff7ad5bd1b7f5546e527e04790ca8c4c9b71b18c73e32
+
+    w_try_unzip "${W_TMP}/${W_PACKAGE}" "${W_CACHE}/${W_PACKAGE}"/nt4pdhdll.exe
+    w_try cp "${W_TMP}/${W_PACKAGE}/pdh.dll" "${W_SYSTEM32_DLLS}/pdh.dll"
+
+    if [ "${W_ARCH}" = "win64" ]; then
+        w_warn "There is no 64-bit version of the WinNT 4.0 pdh.dll. If your program doesn't work then try a 32-bit wineprefix or use 'winetricks pdh' instead."
+    fi
```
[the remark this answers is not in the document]
WheezyE: Nice catch!  I've moved this to the top of the pdh_nt4 module now

WheezyE (commented):
Sorry again for the slow reply.  I finally had a chance to install a 64-bit Debian VM on my laptop at test a 64-bit wineprefix with this and it seems to look pretty good.  If the extra warning here is redundant, I can just remove it too - since it looks like winetricks already warns 64-bit users that some packages only install 32-bit files.

```
Using winetricks 20210206-next - sha256sum: 2cdd52339d66884090e68a04b864d649a04732e91b06de9c470c0fecd60a8815 with wine-4.0 (Debian 4.0-2) and WINEARCH=win64
winetricks GUI enabled, using zenity 3.30.0
Executing w_do_call pdh_nt4
Executing mkdir -p /home/pi
warning: You are using a 64-bit WINEPREFIX. Note that many verbs only install 32-bit versions of packages. If you encounter problems, please retest in a clean 32-bit WINEPREFIX before reporting a bug.
Executing load_pdh_nt4 
warning: There is no 64-bit version of the WinNT 4.0 pdh.dll. If your program doesn't work then try a 32-bit wineprefix or use 'winetricks pdh' instead.
Executing mkdir -p /home/pi/.cache/winetricks/pdh_nt4
Executing cd /home/pi/.cache/winetricks/pdh_nt4
Downloading http://download.microsoft.com/download/winntsrv40/update/5.0.2195.2668/nt4/en-us/nt4pdhdll.exe to /home/pi/.cache/winetricks/pdh_nt4
Executing cd /home/pi
Executing cp /home/pi/.wine/dosdevices/c:/windows/temp/pdh_nt4/pdh.dll /home/pi/.wine/dosdevices/c:/windows/syswow64/pdh.dll
Using native,builtin override for following DLLs: pdh
Executing wine regedit C:\windows\Temp\override-dll.reg
it looks like wine32 is missing, you should install it.
multiarch needs to be enabled first.  as root, please
execute "dpkg --add-architecture i386 && apt-get update &&
apt-get install wine32"
Executing wine64 regedit C:\windows\Temp\override-dll.reg

```

PR #1794, opened 2021-08-01
fix case change operations on/in case-insensitive filesystems/directories

openglfreak:


austin987:
These should be separate commits prefixed with the verb, i.e. "protectionid: fix case change operations on/in case-insensitive filesystems/directories".

Did you find these with a script? If so, it would be helpful to add it to tests/winetricks-test so similar cases don't sneak in in the future.

openglfreak:
I did not find them with a script, I just searched for `w_try mv` with the text editor.

austin987:
I don't know that this is a common case worth fixing, but if you're so inclined, a generic could be made by introducing a w_try_mv() function, a la w_try_cd() to check for cases were only the case differs and handle it accordingly.

openglfreak:
Should I do that?

austin987:
I'm fine with the current changes, if you split the commits.

The suggestion would be a more permanent fix, but you don't have to do that. Up to you, I don't think this affects very many people 🙂

openglfreak:
https://gist.github.com/openglfreak/14b505758b05a7cb7112265f77c17954
Would something like this be ok?

openglfreak:
Meh, I'll leave it up to verb authors to not break something, just merge as-is for now

PR #1796, opened 2021-08-04
Update PhysX version

SveSop:
Newer than 9.14 is needed for Batman Arkham Knight to
render "Interactive smoke/fog". This will download and
install latest 9.19 version.

### DIFF
--- a/src/winetricks
+++ b/src/winetricks
@@ -11557,16 +11557,16 @@ load_peverify()
 w_metadata physx dlls \
     title="PhysX" \
     publisher="Nvidia" \
-    year="2014" \
+    year="2019" \
     media="download" \
-    file1="PhysX-9.14.0702-SystemSoftware.msi" \
-    installed_file1="${W_PROGRAMS_X86_WIN}/NVIDIA Corporation/PhysX/Engine/v2.8.3/PhysXCore.dll"
+    file1="PhysX-9.19.0218-SystemSoftware.exe" \
+    installed_file1="${W_PROGRAMS_X86_WIN}/NVIDIA Corporation/PhysX/Engine/86C5F4F22ECD/APEX_Particles_x64.dll"
 
 load_physx()
 {
-    w_download https://uk.download.nvidia.com/Windows/9.14.0702/PhysX-9.14.0702-SystemSoftware.msi 0a022e28accf5851be9d6577487cdcd3d3a3e2a8a21a64456b72b415c217f03c
+    w_download https://uk.download.nvidia.com/Windows/9.19.0218/PhysX-9.19.0218-SystemSoftware.exe 4f36389fcbfbdef4781fb85e7a68373542903235f65d93f7143693738324917c
     w_try_cd "${W_CACHE}/${W_PACKAGE}"
-    w_try "${WINE}" msiexec /i PhysX-9.14.0702-SystemSoftware.msi ${W_OPT_UNATTENDED:+/q}
+    w_try "${WINE}" PhysX-9.19.0218-SystemSoftware.exe ${W_OPT_UNATTENDED:+/s}
 }
 
 #----------------------------------------------------------------


src/winetricks
```diff
@@ -11557,16 +11557,16 @@ load_peverify()
...
+    file1="PhysX-9.19.0218-SystemSoftware.exe" \
     installed_file1="${W_PROGRAMS_X86_WIN}/NVIDIA Corporation/PhysX/Engine/v2.8.3/PhysXCore.dll"
 
 load_physx()
 {
-    w_download https://uk.download.nvidia.com/Windows/9.14.0702/PhysX-9.14.0702-SystemSoftware.msi 0a022e28accf5851be9d6577487cdcd3d3a3e2a8a21a64456b72b415c217f03c
+    w_download https://uk.download.nvidia.com/Windows/9.19.0218/PhysX-9.19.0218-SystemSoftware.exe 4f36389fcbfbdef4781fb85e7a68373542903235f65d93f7143693738324917c
     w_try_cd "${W_CACHE}/${W_PACKAGE}"
-    w_try "${WINE}" msiexec /i PhysX-9.14.0702-SystemSoftware.msi ${W_OPT_UNATTENDED:+/q}
+    w_try "${WINE}" PhysX-9.19.0218-SystemSoftware.exe ${W_OPT_UNATTENDED:+/q}
```
austin987, line 11569: The newer physx doesn't support /q; looks like it uses /s now.

austin987 (changes requested):

src/winetricks
SveSop: Yeah, it seems so. Still it will not actually allow you to do a true silent install, cos it seems you still have to click "I agree" button on the license agreement.
I updated it and added a filecheck that i missed. `v2.8.3/PhysXCore.dll` exists on both, although a different hash (and i dunno if you can do a hash-check with `installed_file`?)
The folder and file `86C5F4F22ECD/APEX_Particles_x64.dll` does not exist on the 9.14 version tho, so it will allow an upgrade without failing with "already installed".

Sorry i missed that.

austin987: It installs fine for me (i.e., without a popup for the license) with wine-6.12-195-gc586c57082d if I use -s.

austin987 (approved):

openglfreak:
This version needs msxml3, I think

austin987:
@SveSop do you want me to merge?

@openglfreak what is msxml3 needed for? Are you seeing a problem with it currently?

SveSop:
> @SveSop do you want me to merge?
> 
Depending on answer below:

> @openglfreak what is msxml3 needed for? Are you seeing a problem with it currently?

@openglfreak I have not tested this with older (5.x branch) of wine/wine-staging or proton. Is there issues with that you think?
Any testcase i could verify the need for msxml3? (Or spot excessive log spam due to this).

gverm:
You might be thinking of [WineHQ bug 40430](https://bugs.winehq.org/show_bug.cgi?id=40430), which has been fixed since wine-6.0.

For 64bit, there was [WineHQ bug 48879](https://bugs.winehq.org/show_bug.cgi?id=48879) as well, which has been fixed since wine-6.4.

SveSop:
> You might be thinking of [WineHQ bug 40430](https://bugs.winehq.org/show_bug.cgi?id=40430), which has been fixed since wine-6.0.
> 
> For 64bit, there was [WineHQ bug 48879](https://bugs.winehq.org/show_bug.cgi?id=48879) as well, which has been fixed since wine-6.4.
> 
> Might be good to "proof" the verb by adding workarounds for both those bugs for versions before 6.4.

I see 48879 mentioned 9.12 failiing. The PhysX version in winetricks has been 9.14 for a good while tho...
I have not tested this on older then 6.3 (Proton is based on 6.3), so i cannot say if it is still the case.

I will quickly compile 6.2 when i get home and see what happens tho.

SveSop:
> I see 48879 mentioned 9.12 failiing. The PhysX version in winetricks has been 9.14 for a good while tho...
> I have not tested this on older then 6.3 (Proton is based on 6.3), so i cannot say if it is still the case.
> 
> I will quickly compile 6.2 when i get home and see what happens tho.

@openglfreak I tested wine-6.2 with the PhysX-6.12 package mentioned from the bugpost, and sure enough, it failed with
```
00f8:trace:msi:MsiActiveScriptSite_OnScriptError (00CAA730)->(00C6A218)
00f8:err:msi:MsiActiveScriptSite_OnScriptError script error: L"ActiveX component can't create object"
```

However, the 9.14 version (.msi) and this new 9.19 (.exe) did not fail or have any issues with wine-6.2. Maybe they (nVidia) changed something with the installer?

Also tested wine-staging-6.14 against the 9.12 (.msi) package, and it worked, aswell as 9.14 and 9.19.

Conclusion: A winebug made PhysX-9.12 package fail with wine < 6.4, but seems solved with PhysX-9.14 (and PhysX-9.19) regardless of using wine-6.2.

There is seemingly no problems/errors installing PhysX-9.19 on wine-stable-6.0.1 either.

gverm (approved):
@SveSop Thanks for testing. In that case this is good to go in my opinion.
The msxml3 bug did apply to this version as well, but since it's been fixed since 6.0, that shouldn't matter.

SveSop:
@austin987 Anything more i need to provide for a merge?

PR #1812, opened 2021-08-27
Remove d9vk* verbs

bobwya:
@austin987
Let me know - if you want me to tidy up the **helper_dxvk_d9vk** function as well.
I'm not sure that would achieve much though.


austin987:
@bobwya that would be nice :). While it's not strictly necessary, it may be confusing to anyone reading the code after the d9vk verbs are gone.

PR #1816, opened 2021-08-31
README: Improved .verb documentation.

devdilso:
I was working on dockerize a windows application that used mdac28. 

But I did not want to download the file from the internet and instead maintain it inside my own repository.

I used a custom ´.verb´ to fix the issue and wanted to improve the documentation with the way I solved the problem.

### DIFF
--- a/README.md
+++ b/README.md
@@ -205,6 +205,27 @@ load_icecat()
 
 Note that the file prefix (icecat.verb) and command name (icecat) must match. All metadata fields are optional, only the command name and category are required.
 
+This feature is specially useful if you need to install a setup manually, instead of relying on internet downloads.
+
+Here is an example that installs mdac28 from a folder named `Libs`: 
+
+```
+load_mdac28()
+{
+    w_package_unsupported_win64
+    load_native_mdac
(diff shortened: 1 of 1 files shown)


austin987:
IMO that's pretty hacky and shouldn't be encouraged. You'd be better off copying the binary into the container before you start (or bindmounting the directory and setting W_CACHE/XDG_CACHE_HOME).

devdilso:
> IMO that's pretty hacky and shouldn't be encouraged. You'd be better off copying the binary into the container before you start (or bindmounting the directory and setting W_CACHE/XDG_CACHE_HOME).



> W_CACHE/XDG_CACHE_HOME

Good to know, thanks.

In my implementation I am also mouting the file and then just installing the .verb I created.

devdilso:
In case some need, here is what I did after @austin987 suggestion: 

Mounted in docker-compose: 
```
- "./Libs/MDAC_TYP.EXE:/root/.cache/winetricks/mdac28/MDAC_TYP.EXE"
```
Installed normally in my startup script:
```
winetricks -q mdac28
```

PR #1820, opened 2021-09-13
workaround SIP DYLD_ environment stripping

Gcenx:
After the introduction of SIP (OS X 10.11) DYLD_ environment variables are stripped.

See https://support.apple.com/en-us/HT204899

<br>

As `winetricks` is launched from a system provided utility DYLD_ variables will be stripped leaving them empty. 

`DYLD_FALLBACK_LIBRARY_PATH` used to provide additional library paths.

### DIFF
--- a/src/winetricks
+++ b/src/winetricks
@@ -5353,6 +5353,12 @@ winetricks_init()
 
     winetricks_get_platform
 
+    # Workaround SIP DYLD_stripping
+    # See https://support.apple.com/en-us/HT204899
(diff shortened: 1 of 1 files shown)


austin987:
Thanks @Gcenx, but could you please add the Apple support link in a comment? Otherwise, LGTM.

austin987:
Minor update:  shellcheck check failed (please add braces around the variable names).

Gcenx:
Done

Gcenx:
Just noticed a double “see” in the comment won’t be able to fix that until I’m in front of a computer.

PR #1827, opened 2021-09-29
cnc_ddraw: update to 4.4.4.0

Gcenx:
Bump cnc_ddraw to the latest version


Gcenx:
Not sure if changing the date to the current release year is correct, if the date should be the initial date added to winetricks or should it be the date of the first release?

austin987:
It _should_ reflect the year of the actual release (but the vast majority are inaccurate). It's optional though and can be omitted.

PR #1832, opened 2021-10-05
w_verify_7z_available: new helper

Gcenx:
Calling w_try_7z when falling back to windows 7-zip wipes $W_TMP that can cause issues, where using a native version of 7z won't wipe $W_TMP this can come in handy.

Sending this before a new verb as I'm using this function, if not desirable I'll need to alter the verb in question to extract to $W_CACHE/W_PACKAGE and wipe once finished.

### DIFF
--- a/src/winetricks
+++ b/src/winetricks
@@ -901,6 +901,15 @@ w_read_key()
     unset _W_keyfile _W_keymsg _W_nokeymsg
 }
 
+w_verify_7z_available()
+{
+    # If verb_a requires verb_b, then verba will fail when the dependency for verb_b is installed
+    # This should be called by verb_a, to give a proper warning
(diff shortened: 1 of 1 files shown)


austin987:
Could you push the branch that would use this somewhere I could see? (A github branch/pastebin/diff, whatever is easier).

Gcenx:
Just pushed the verb into this branch.

`originupdate` Needs the native 7z as it needs to extract in two steps, this could be changed to extract into another location if the helper isn’t desirable.

Alternatively I could change the verb to only support 6.7+ and drop the updated verb for upstream winetricks.

Gcenx:
The origin verbs are slightly modified from the version @gverm and myself were working on. I noticed while on the train to work I didn’t have Gverm added as co-author for that commit.

PR #1838, opened 2021-10-26
fix a wrong Chinese translation

lilydjwg:
Here `down` does not mean `download`.


src/winetricks
```diff
@@ -1,4 +1,4 @@
-#!/bin/sh
+Winetricks/winetricksWinetricks/winetricksWinetricks/winetricksWinetricks/winetricksWinetricks/winetricks#!/bin/sh
 # shellcheck disable=SC2030,SC2031
```
austin987, line 2: Please remove this change.

austin987 (changes requested):

src/winetricks
lilydjwg: Oops, I don't know why it's here. I was just using the GitHub web interface.... It's fixed now.

austin987 (approved):

PR #1849, opened 2021-11-25
msaa: new verb

alexhenrie:
This verb provides a workaround for [Wine's oleacc bugs](https://bugs.winehq.org/buglist.cgi?component=oleacc&list_id=749105&product=Wine&resolution=---). I can confirm that the NVDA screen reader works quite decently with the native DLL!

Download description page: https://www.microsoft.com/en-us/download/details.aspx?id=21596

### DIFF
--- a/src/winetricks
+++ b/src/winetricks
@@ -10938,6 +10938,28 @@ load_mfc70()
 
 #----------------------------------------------------------------
 
+w_metadata msaa dlls \
+    title="MS Active Accessibility (oleacc.dll, oleaccrc.dll, msaatext.dll)" \
+    publisher="Microsoft" \
+    year="2003" \
+    media="download" \
+    file1="MSAA20_RDK.exe" \
+    installed_file1="${W_SYSTEM32_DLLS_WIN}/oleacc.dll" \
+    installed_file2="${W_SYSTEM32_DLLS_WIN}/oleaccrc.dll" \
+    installed_file3="${W_SYSTEM32_DLLS_WIN}/msaatext.dll"
+
+load_msaa()
+{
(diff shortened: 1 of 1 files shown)


src/winetricks
```diff
@@ -10938,6 +10938,28 @@ load_mfc70()
 
 #----------------------------------------------------------------
 
+w_metadata msaa dlls \
+    title="MS Active Accessibility (oleacc.dll, oleaccrc.dll, msaatext.dll)" \
+    publisher="Microsoft" \
+    year="2003" \
+    media="download" \
+    file1="../msaa/MSAA20_RDK.exe" \
```
austin987, line 10946: You can just use the filename directly, since the file is in the directory with the same name as the package.

src/winetricks
```diff
@@ -10938,6 +10938,28 @@ load_mfc70()
...
+    installed_file1="${W_SYSTEM32_DLLS_WIN}/oleacc.dll" \
+    installed_file2="${W_SYSTEM32_DLLS_WIN}/oleaccrc.dll" \
+    installed_file3="${W_SYSTEM32_DLLS_WIN}/msaatext.dll"
+
+load_msaa()
+{
+    w_download_to msaa https://download.microsoft.com/download/c/1/c/c1cf13a6-4d7f-4b7d-9f67-51ef3a421fc7/MSAA20_RDK.exe 081e382f7e5b874ab143f0b073246fd31f84ae181df1838813b02935a951c9da
```
austin987, line 10953: Similarly, you can use `w_download https...` instead of `w_download to msaa https...`, since this is the `msaa` package.

src/winetricks
alexhenrie: Fixed, thanks!

src/winetricks
alexhenrie: Fixed, thanks!

alexhenrie:
Thank you!

PR #1866, opened 2022-01-09
setpath: new verb to set the user PATH variable in a wine prefix from the WINEPATH environment variable

rhabacker:


austin987:
What's the motivation for this? I can see that it's a useful code snippet, but I don't see why most/many users would need this.

rhabacker:
One use case is the use of Wine to run cross compiled applications for Windows. Corresponding applications and additional libraries are usually not located at the standard paths defined by Wine and must therefore be accessible via the PATH variable. 

An example of this is the DBus project, which uses Wine in its CI system for testing the created applications.

Until now, each project has to implement the setting of the path within Wine in a complex way (see e.g. https://gitlab.freedesktop.org/dbus/dbus/-/blob/master/tools/ci-build.sh#L39), which is simplified with this function.

austin987:
Makes sense, I'm happy to include it if it would help other FOSS projects :)

Could you please fix the shellcheck issues that the CI found? Thanks!

rhabacker:
> Could you please fix the shellcheck issues that the CI found?

Fixed, simplified the implementation of that verb as `winepath -w` supports more than one path and made the doc more specific.

Issue #1905, opened 2022-04-15
isolate_home leaves "Templates" symlink to $HOME


ichthyosaurus:
- winetricks: 20220411-next
- wine: 7.5

The `isolate_home` verb actually doesn't remove symlinks to `$HOME` but simply removes all symlinks (regardless of their target) in `$WINEPREFIX/drive_c/users/$USER`. This is not done recursively, so the link `$WINEPREFIX/drive_c/users/$USER/AppData/Roaming/Microsoft/Windows/Templates` pointing to `$HOME` is left untouched.

---

Ideas:

As a simple fix, it would be enough to explicitly check and remove `$WINEPREFIX/drive_c/users/$USER/AppData/Roaming/Microsoft/Windows/Templates`. E.g. insert below this line:

https://github.com/Winetricks/winetricks/blob/376351b965e14b231f3b75f86e9ed014698c56d3/src/winetricks#L22391

```bash
    x="${WINEPREFIX}/drive_c/users/${USER}/AppData/Roaming/Microsoft/Windows/Templates"
    if test -h "${x}" && test -d "${x}"; then
        rm -f "${x}"
        mkdir -p "${x}"
    fi
    unset x
```

A more elaborate fix could use `find` like this, replacing the whole loop:

https://github.com/Winetricks/winetricks/blob/376351b965e14b231f3b75f86e9ed014698c56d3/src/winetricks#L22386

```bash
LANG=C find . -type l -exec sh -c '
        prefix="$(realpath "$0")";
        target="$(realpath "$1")";
        if [[ "$target" == "$prefix/"* ]]; then
            echo "leaving symlink pointing inside the prefix: $1-> $target";
        elif [[ "$target" == "$HOME/"* || "$target" == "$HOME" ]]; then
            echo "removing symlink $1 -> $target...";
            rm -f "$1";
            mkdir -p "$1";
        else
            echo "leaving data directory symlink not pointing to \$HOME: $1 -> $target";
        fi' "${WINEPREFIX}" {} \;
```

I'm not sure if this works with plain `sh`, though. On my system, `sh` is linked to `bash`.

PR #1924, opened 2022-07-18
isolate_home: correctly handle subdirectory symlinks to $HOME

bobwya:
Remove symlinks that target the $HOME directory (or subdirectories
of this directory) within the current WINEPREFIX C:\users\<user>
directory.
Test for symlinks which are in subdirectories of the current
WINEPREFIX.
Ignore symlinks to files (as before).

Closes: #1905


austin987:
Thanks!

PR #1925, opened 2022-07-18
winetricks: interpret backslash escape sequences for console messsages

bobwya:
w_askpermission w_info w_warn w_warn_cancel: update functions
Avoid outputing raw '\n' sequences when winetricks is called from
a BASH shell.

### DIFF
--- a/src/winetricks
+++ b/src/winetricks
@@ -182,13 +182,21 @@ else
     W_TMP_EARLY="/tmp"
 fi
 
+# shellcheck disable=SC2039,SC3037
+W_ECHO_PARAM="$(echo "-e")"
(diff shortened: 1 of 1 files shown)


bobwya:
@austin987 does this look OK (I force pushed a fix a few days ago)?

Btw I've got 2 shellcheck overrides - both doing the same thing!
As with the latest shellcheck version (on my system: 0.8.0) - the utility has changed the SC code (vs the shellcheck version winetricks is using).
See: [SC3037](https://www.shellcheck.net/wiki/SC3037)

austin987:
@bobwya why not instead use printf?

bobwya:
> @bobwya why not instead use printf?

Opened a **printf_console_messages** branch for this...

Issue #1920, opened 2022-07-09
Confusing ""Mono does not appear to be installed" message when installing dotnet


matthijskooijman:
I just tried to install dotnet462 using winetricks, which shows a prompt ""Mono does not appear to be installed" three times (and then it breaks because my wine version is incompatible with dotnet45).

This message is quite confusing. I interpreted this as "You need to install mono before installilng dotnet462", which seemed weird but not entirely implausible. However, after reading the code I found that dotnet462 actually tries to *uninstall* mono using the `remove_mono` verb, and that is telling me it cannot uninstall mono because it is not installed.

It is extra confusing because dotnet462 depends on dotnet461, dotnet46, dotnet45 and dotnet40, and each of these show the same confusing prompt.

To prevent this confusion, I can imagine that dotnet462 (and others that call remove_mono) could skip the call if mono is not installed, or maybe they can pass a parameter so remove_mono will silently return if not installed, or maybe just changing the message to "Cannot remove mono: it does not appear to be installed" (which is not ideal, but should at least be a lot less confusing)?

PR #1928, opened 2022-07-25
remove_mono: support 'internal' argument

bobwya:
remove_mono internal: no warning when wine-mono is not installed
dotnet* vc2010express: update verbs to call remove_mono internal

Closes: #1920


austin987:
Thanks!

Issue #1720, opened 2021-03-24
[Feature Request] Being able to run winecmd


SlushThorsten:
That would be useful for batch scripts

austin987:
That's already a thing.

Toadfield:
How?
I can't find that option
![image](https://user-images.githubusercontent.com/68649672/112653288-8e3eb680-8e4e-11eb-9ca4-caf0bd0f8fa6.png)

austin987:
My mistake, commandline shell runs a native shell, not winecmd.

PR #1931, opened 2022-07-28
winecmd: new verb

bobwya:
Closes: #1720


austin987:
There's already a 'cmd' verb: https://github.com/Winetricks/winetricks/blob/master/src/winetricks#L5919

bobwya:
@austin987 Oops! Renamed as: **winecmd** - OK?

PR #1942, opened 2022-08-07
Vcrun use w_try_ms_installer & handle status 236

Gcenx:
vcrun verbs should also use `w_try_ms_installer` as these installers pass these exit codes.

Also added handling status 236, this being used to indicate a newer version of the dlls already exist. Seen this one after installing Origin then attempting to install vcrun2010.


Gcenx:
Strangely on the x86 installer give this status where x64 didn’t.

PR #1947, opened 2022-08-17
origin|originupdater: new verbs

Gcenx:
The `origin` verb will install the Origin client and exit cleanly after install and can also be installed silently.

Darwin requirements need to be higher due to most being on more modern versions of macOS (10.15+) so forced to use a winecx variant where Origin _doesn't_ function until the more recent WineCX21.x based builds (Wine-6.0+) due to some missing thunks within `wine32on64` related functions

OriginSetup automatically installs the listed vcrun installers so added those to ensure they get registered correctly, `d3dcompiler_47` is used by the Origin client.

The override of `gameux` being set to disabled made games correctly register when left enabled games will goto 100% and then not show as playable without rebooting Origin where having this disabled makes games register without issue.

The `originupdated` verb is used to workaround Bugzilla `44691` by manually unpacking `OriginSetup.exe` and installing into place, this can also be used to force update the Origin client.

### DIFF
--- a/src/winetricks
+++ b/src/winetricks
@@ -16029,6 +16029,84 @@ load_openwatcom()
 
 #----------------------------------------------------------------
 
+w_metadata origin apps \
+    title="EA Origin" \
+    publisher="EA" \
+    year="2011" \
+    media="download" \
+    file1="OriginSetup.exe" \
+    installed_file1="${W_PROGRAMS_X86_WIN}/Origin/Origin.exe" \
+    homepage="https://www.origin.com/"
+
+helper_origin_dl()
+{
+    # Skipping checksum as this changes too often
+    w_download_to origin https://origin-a.akamaihd.net/Origin-Client-Download/origin/live/OriginSetup.exe
+}
+
+load_origin()
+{
+    # Need to force wine-6.0 as Origin doesn't run below WineCX21 (wine32on64)
+    if [ "$(uname -s)" = "Darwin" ] && w_wine_version_in ,6.0 ; then
+        w_die "${W_PACKAGE} requires wine version 6.0 (or newer)"
+    fi
+
+    if [ "${WINETRICKS_FORCE}" != 1 ] && w_workaround_wine_bug 44691 "Installer fails under wine, manually unpacking it instead" 6.7,; then
+        w_call originupdater
+        w_warn "${W_PACKAGE} might fail, to update use the originupdate verb"
+    else
+        helper_origin_dl
+        w_try_cd "${W_CACHE}/${W_PACKAGE}"
+        w_try "${WINE}" "${file1}" /NoLaunch ${W_OPT_UNATTENDED:+/SILENT}
+    fi
+
+    if w_workaround_wine_bug 32342 "QtWebEngineProcess.exe crashes when updating or launching Origin (missing fonts)"; then
+        w_call corefonts
+    fi
+
+    if w_workaround_wine_bug 44985 "Disabling libglesv2 to make Store and Library function correctly."; then
+        w_override_dlls disabled libglesv2
+    fi
+
+    # Origin requirements
+    w_call vcrun2010
+    w_call vcrun2019
+    w_call d3dcompiler_47
+
+    # Avoids "An unexpected error has occurred. Please try again in a few moments. Error: 327684:3"
+    # Games won't registor correctly unless disabled
+    w_override_app_dlls origin.exe disabled gameux
+}
+
+#----------------------------------------------------------------
+
+w_metadata originupdater apps \
+    title="EA Origin (updater)" \
+    publisher="EA" \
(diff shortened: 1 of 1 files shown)


src/winetricks
```diff
@@ -16029,6 +16029,85 @@ load_openwatcom()
...
+}
+
+load_origin()
+{
+    # Need to force wine-6.0 as Origin doesn't run below WineCX21 (wine32on64)
+    if [ "$(uname -s)" = "Darwin" ] && w_wine_version_in ,6.0 ; then
+        w_die "${W_PACKAGE} requires wine version 6.0 (or newer)"
+    fi
+
+    if [ "${WINETRICKS_FORCE}" = 0 ] && w_workaround_wine_bug 44691 "Installer fails under wine, manually unpacking it instead" 6.7,; then
```
Gcenx, line 16054: As many custom wine packages include a patch that fixes this but this was tweaked to allow bypassing the unpacking in favour or running the actual installer.

src/winetricks
```diff
@@ -16029,6 +16029,85 @@ load_openwatcom()
...
+    fi
+
+    if w_workaround_wine_bug 44985 "Disabling libglesv2 to make Store and Library function correctly."; then
+        w_override_dlls disabled libglesv2
+    fi
+
+    # Origin requirements
+    w_call vcrun2010
+    w_call vcrun2015
+    w_call vcrun2019
```
Gcenx, line 16074: These should be shrunk down to only `vcrun2019` & `d3dcompiler_47` the other two were added due to OriginSetup also installing those during setup.

src/winetricks
```diff
@@ -16029,6 +16029,85 @@ load_openwatcom()
...
+
+    # Remove cached installer as the checksum changes too often that is even more
+    # critical for the updater function
+    rm -f "${W_CACHE}/origin/OriginSetup.exe"
+
+    helper_origin_dl
+
+    w_try_7z "${W_CACHE}"/origin "${W_CACHE}"/origin/OriginSetup.exe update
+    w_try_7z "${W_PROGRAMS_X86_UNIX}"/Origin "${W_CACHE}"/origin/update/OriginUpdate_*_*_*_*.zip -aoa
+    rm -f "${W_CACHE}"/origin/OriginSetup
```
austin987, line 16106: Missing w_try. Also looks wrong, is there really a file named 'OriginSetup'? Or was this meant to be 'OriginSetup.exe'? Or is a folder and should be 'rm -rf OriginSetup'?

src/winetricks
```diff
@@ -16029,6 +16029,85 @@ load_openwatcom()
...
+load_originupdater()
+{
+    # Need to force wine-6.0 as Origin doesn't run below WineCX21 (wine32on64)
+    if [ "$(uname -s)" = "Darwin" ] && w_wine_version_in ,6.0 ; then
+        w_die "${W_PACKAGE} requires wine version 6.0 (or newer)"
+    fi
+
+    # Remove cached installer as the checksum changes too often that is even more
+    # critical for the updater function
+    rm -f "${W_CACHE}/origin/OriginSetup.exe"
```
austin987, line 16100: Similar here, missing w_try. Though in theory this could break if someone is using a read-only cache, in practice I don't think that's a common setup, particularly for anyone using 'originupdater'.

src/winetricks
```diff
@@ -16029,6 +16029,85 @@ load_openwatcom()
...
+
+load_originupdater()
+{
+    # Need to force wine-6.0 as Origin doesn't run below WineCX21 (wine32on64)
+    if [ "$(uname -s)" = "Darwin" ] && w_wine_version_in ,6.0 ; then
+        w_die "${W_PACKAGE} requires wine version 6.0 (or newer)"
+    fi
+
+    # Remove cached installer as the checksum changes too often that is even more
+    # critical for the updater function
```
austin987, line 16099: Nitpick: please combine this into one line, it's not that long :)

src/winetricks
Gcenx: Changed this to `"${WINETRICKS_FORCE}" != 1`

src/winetricks
Gcenx: Yeah that shouldn't be OriginSetup anymore, it was OriginSetup.exe prior but should actually be `update` to clear out the extracted directory.

src/winetricks
Gcenx: Done

src/winetricks
Gcenx: Done

src/winetricks
Gcenx: Done

src/winetricks
Gcenx: Not sure why this doesn’t show outdated…

Removed vcrun2015 since vcrun2019 already overs this is forgotten….

PR #1951, opened 2022-08-22
Fix vc2010express

edmcman:
The old code only works with --no-clean.  When cleaning is turned on, the unpacked VCExpress directory will be blown away as soon as `w_call` is called.


austin987:
Nice catch, thanks!

PR #1961, opened 2022-09-24
Wow64 tweaks

Gcenx:
Upstreams wow64 system will use `wine64` for 32Bit binaries also, this is already being used within CrossOver-22.0.0

### DIFF
--- a/src/winetricks
+++ b/src/winetricks
@@ -820,7 +820,7 @@ w_try_regedit64()
 
 w_try_regsvr()
 {
-    w_try "${WINE}" regsvr32 ${W_OPT_UNATTENDED:+/S} "$@"
+    w_try "${WINE}" "${W_SYSTEM64_DLLS_WIN32}/regsvr32" ${W_OPT_UNATTENDED:+/S} "$@"
 }
 
 w_try_regsvr64()
--- a/src/winetricks
+++ b/src/winetricks
@@ -803,7 +803,7 @@ w_try_regedit32()
     esac
 
     # shellcheck disable=SC2086
-    w_try "${WINE_MULTI}" ${cmdc} regedit ${W_OPT_UNATTENDED:+/S} "$@"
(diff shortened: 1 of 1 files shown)


Gcenx:
This is wrong, do we even still need to run regedit twice anymore?

austin987:
AFAIK, yes. See https://learn.microsoft.com/en-us/windows/win32/sysinfo/32-bit-and-64-bit-application-data-in-the-registry

Unless something changed on the wine side that I missed (which is very possible, please let me know if I did).

Gcenx:
I’ll do some additional testing with my CX22 based build and see how WoW64 handles this (it’s probably still correct for most things)

At least the paths I’d tired to use were incorrect as the standard regedit is within c:\windows but a 32Bit version in a 64Bit prefix would be within c:\windows\syswow64.

austin987:
I can see how having explicit paths would be an improvement, so in that sense, I encourage this PR.

Remember that this needs to be tested in multiple configs. I.e., CX22 is a fine config to support, I'm more concerned with upstream wine/wine-staging, and also Linux/BSD, as well as OSX. That said, you're able to test OSX/CX22, which is also super helpful :). In any event, I appreciate the PR, and testing :).

In short though, this *must* pass 'make check' on linux/wine32 linux/wine64 (and ideally, osx/wine32 osx/wine64 bsd/wine32 bsd/wine64 cx/wine32 cx/wine64).

austin987:
Also, after testing, if this tweak is still needed please reopen the issue :)

Gcenx:
I’ll be testing against regular upstream along with WineCX21 (wine32on64) & WineCX22.

WineCX22 (wine-7.7) makes use of upstream still incomplete wow64 system meaning even 32Bit windows binaries will be launched via wine64, that’s the reason to have an exact PATH to the correct arch binary.

> Also, after testing, if this tweak is still needed please reopen the issue :)

Once I’m done with the expanded testing I’ll get back to this.

PR #1966, opened 2022-10-02
ucrtbase2019: new verb

Gcenx:
I remember a number of online games require this or they get un-synced.

GE-Proton bundled protonfixes uses a custom version of vcrun2019 that uses this installer as it's the final version that still bundled ucrtbase

### DIFF
--- a/src/winetricks
+++ b/src/winetricks
@@ -12889,7 +12889,7 @@ w_metadata vcrun2015 dlls \
     publisher="Microsoft" \
     year="2015" \
     media="download" \
-    conflicts="vcrun2017 vcrun2019" \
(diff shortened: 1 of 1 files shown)


src/winetricks
```diff
@@ -12968,7 +12968,7 @@ w_metadata vcrun2017 dlls \
     publisher="Microsoft" \
     year="2017" \
     media="download" \
-    conflicts="vcrun2015 vcrun2019" \
+    conflicts="vcrun2017 vcrun2019 ucrtbase2019" \
```
austin987, line 12971: Why did you remove vcrun2015, and why was vcrun2017 added? There's no need for it to conflict with itself.

Gcenx: My mistake I'll fix that

PR #1967, opened 2022-10-02
dxwnd: new verb

Gcenx:
The aim of this application is to hook games allowing even older games to support windowed mode or simply get working on modern versions of windows, this also helps getting some games working within wine.

Lutris has many scripts that make use of this and is also good for getting legacy titles working again on macOS Monterey that worked on prior macOS versions (using winecx/crossover)

### DIFF
--- a/src/winetricks
+++ b/src/winetricks
@@ -14842,6 +14842,24 @@ load_dxdiag()
 
 #----------------------------------------------------------------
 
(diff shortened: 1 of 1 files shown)


Gcenx:
A new version might added soon, I'll update the PR when it lands if it's not been merged before this.

PR #1968, opened 2022-10-02
origin: Connect overriders to Winehq bugs

Gcenx:
- workaround for the constant crashing from igoproxy(64).exe
- vcrun2013 as I'd forgotten this
- w_warn to ensure in-game origin gets disabled (many games crash when it's enabled)

Should wait for #1966 to be merged to ensure a new version of ucrtbase is usable.

### DIFF
--- a/src/winetricks
+++ b/src/winetricks
@@ -16066,21 +16066,48 @@ load_origin()
         w_call corefonts
     fi
 
+    if w_workaround_wine_bug 44258 "Origin crashes on start."; then
+        w_override_app_dlls igoproxy.exe disabled d3d10
+        w_override_app_dlls igoproxy.exe disabled d3d10_1
+        w_override_app_dlls igoproxy.exe disabled d3d10core
+        w_override_app_dlls igoproxy.exe disabled d3d11
+        w_override_app_dlls igoproxy.exe disabled d3d12
+        w_override_app_dlls igoproxy.exe disabled dxgi
+        w_override_app_dlls igoproxy.exe disabled vulkan-1
+        w_override_app_dlls igoproxy.exe disabled winevulkan
+
+        if [ "${W_ARCH}" = "win64" ]; then
+            w_override_app_dlls igoproxy64.exe disabled d3d10
+            w_override_app_dlls igoproxy64.exe disabled d3d10_1
+            w_override_app_dlls igoproxy64.exe disabled d3d10core
+            w_override_app_dlls igoproxy64.exe disabled d3d11
+            w_override_app_dlls igoproxy64.exe disabled d3d12
+            w_override_app_dlls igoproxy64.exe disabled dxgi
+            w_override_app_dlls igoproxy64.exe disabled vulkan-1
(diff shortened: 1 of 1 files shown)


src/winetricks
```diff
@@ -16066,21 +16066,48 @@ load_origin()
...
 
+    # Avoids "An unexpected error has occurred. Please try again in a few moments. Error: 327684:3"
+    # Games won't registor correctly unless disabled
+    if w_workaround_wine_bug 52781 "Origin does not notice games exiting, does not allow them to be relaunched."; then
+        w_override_app_dlls origin.exe disabled gameux
+    fi
+
     # Origin requirements
     w_call vcrun2010
+    w_call vcrun2013
```
austin987, line 16103: Is there a WineHQ bug for that (or the other vcrun verbs)?

src/winetricks
```diff
@@ -16066,21 +16066,48 @@ load_origin()
...
+            w_override_app_dlls igoproxy64.exe disabled winevulkan
+        fi
+    fi
+
     if w_workaround_wine_bug 44985 "Disabling libglesv2 to make Store and Library function correctly."; then
         w_override_dlls disabled libglesv2
     fi
 
+    # Avoids "An unexpected error has occurred. Please try again in a few moments. Error: 327684:3"
+    # Games won't registor correctly unless disabled
```
austin987, line 16096: This should be 'register'

src/winetricks
```diff
@@ -16066,21 +16066,48 @@ load_origin()
...
     w_call vcrun2019
 
     if w_wine_version_in ,6.3 ; then
         w_call d3dcompiler_47
     fi
 
-    # Avoids "An unexpected error has occurred. Please try again in a few moments. Error: 327684:3"
-    # Games won't registor correctly unless disabled
-    w_override_app_dlls origin.exe disabled gameux
+    w_warn "Origin In-game overlay must be disabled"
```
austin987, line 16110: Does disabling 'gameux' not work? Maybe the override is needed for a different executable name?

src/winetricks
Gcenx: I don't remember seeing any, however the origin client it's self installs vcrun2010/vcrun2013/vcrun2015, I'd swapped out vcrun1015 with vcrun2019 for the newer dlls.

I've not tested how well Origin runs with the builtin dlls but the Winehq appdatabase Origin section does say to install these verbs, different origin components load these relevant dlls.

src/winetricks
Gcenx: Disabling `gameux` is to get games to register correctly it was just moved under a `w_workaround_wine_bug` section above linked to the relevant bug.

In-game-origin is `igoproxy.exe` & `igoproxy64.exe` along with the connected dlls but disabling these caused Origin be think it's offline, theres a Proton patch that fixes in-game-origin that hasn't made it upstream yet.

src/winetricks
Gcenx: Fixed in the force push.

PR #1974, opened 2022-10-30
vcrun2022: new verb

Saancreed:
Inspired by `vcrun2019` but with a few changes:

1. Doesn't override various apiset DLLs that `vcrun2019` does.
    * I don't exactly understand the reasoning for overriding them in `vcrun2019` but neither `vcrun2019` nor `vcrun2022` ship any of them.
2. Mentions new MFC DLLs that are also available in `vcrun2019` nowadays but doesn't override them.
3. Uses `vcruntime140.dll` as `installed_file1` instead of `mfc140.dll` as the former is specific for Visual C++ Runtime packages from 2015 onward.
4. Has updated link to Microsoft documentation.
5. ~~Attempts to workaround bug 50894 only when Wine version is older than 6.7, according to upstream bug report this is the version where it got fixed.~~ Doesn't include workaround for bug 50894 because `vcrun2022` installs only when Windows version is Win7 or newer.
6. ~~When the workaround for 50894 is applied, it attempts to query and save current Windows version and restore it later.~~
    * ~~I couldn't verify if this works 100% correctly because I don't have any version of Wine that's old enough to check. Testing would be appreciated :slightly_smiling_face:~~
7. Comments with checksums as they change over time use ISO 8601 format for dates, include exact version number and provide direct download links for packages instead of checksums themselves. Second to last path fragment in the URL is the checksum itself.


Gcenx:
If you wanted to test the function you could set `WINETRICKS_WINE_VERSION` to something before 6.7.

If this works it would be better integrating this into the current `w_set_winver` function so the users currently set version can be restored over `default`

Saancreed:
@Gcenx Thanks for the tip. Actually, I actually have just tried removing known good version from arguments to `w_workaround_wine_bug` call. As it turns out, `vcrun2022` refuses to install if Windows version is set to anything older than Win7 so I've just removed the workaround altogether.

On the other hand, updating the workaround in `vcrun2019` _almost_ works, but I had to strip trailing `\r` from the output of `winecfg -v`, changing the command to restore version to `w_set_winver "${vcrun2019_restore_winver%"$(printf '\r')"}"`. Ugly, but I can't really think of anything better.

It's also unclear to me how could this be integrated into `w_set_winver` since we still need to preserve current Windows version in a (local) variable. Maybe just moving the reading logic and stripping trailing `\r` there would be nice but I'd consider it out of scope of this PR, even more so now that the workaround for `vcrun2022` is gone.

PR #1978, opened 2022-11-07
steam: disable libglesv2 only for steamwebhelper.exe

Gcenx:
Restore disabling `libglesv2` but make it targeted to `steamwebhelper.exe`, after expaneded testing across more wine/macOS versions sometimes this was required for the Libary/Store to not render black.

### DIFF
--- a/src/winetricks
+++ b/src/winetricks
@@ -16487,8 +16487,8 @@ load_steam()
         w_override_dlls disabled gameoverlayrenderer
     fi
 
-    if w_workaround_wine_bug 44985 "Disabling libglesv2 to make Store and Library function correctly." 7.0,; then
-        w_override_dlls disabled libglesv2
+    if w_workaround_wine_bug 44985 "Disabling libglesv2 to make Store and Library function correctly."; then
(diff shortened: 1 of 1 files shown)


Gcenx:
Want to test against wine-6.x to see if that can render retrologin or not first. Then tweak this Pr accordingly.

austin987:
Hi @Gcenx, I'm curious what happened here. What ended up being the issue/correct fix? Is there some other bug (tracker)I can look at?

Gcenx:
@austin987 it’s a weird one and hard to replicate with upstream wine due to some required hacks to make Steam work on macOS.

I’d kicked this up to CodeWeavers who’ve said they’ve been unable to replicate the issue, when using my own wine-crossover-22.0.1 based builds I usually end-up with a fully functional Steam client now.

Annoyingly on multiple Reddit subs and a number of discord channels I keep seeing users having issues with Steam installed via CrossOver crosstie, but it seems to be mostly fine when installed via winetricks as long as you occasionally clear Steams cache.

PR #2014, opened 2023-02-02
autohotkey: bump to 1.1.36.01 in w_ahk_do()

BowDown097:
The version in this method was still old, which caused breakages when installing components like allcodecs.

### DIFF
--- a/src/winetricks
+++ b/src/winetricks
@@ -1798,10 +1798,10 @@ _EOF_
 
 w_ahk_do()
 {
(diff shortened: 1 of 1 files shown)


austin987:
The older version works fine with allcodecs here (but so does this, so I'm fine with merging it).

Just curious what sort of breakage your seeing?

BowDown097:
Looking at it now I believe the downloading task in this method is a fallback for if the normal install method fails for whatever reason. The problem is the URL wasn't updated here, so if you do end up in this fallback it will never succeed because the URL is dead.

PR #2015, opened 2023-02-02
misc: remove games

austin987:
This code has been unmaintained for years (other than mass fixes).

I don't think anyone is using these games or cares, but I'm going to leave this PR open for a few weeks for comments before merging in case someone does.


SmallJoker:
:+1: I don't see a need for this in Winetricks either. Installing (mainly older) games manually is not too complicated (Wine AppDB to the rescue) and the internet archive contains a much larger list of demos and shareware applications anyway:

* https://archive.org/details/cdbbsarchive
* https://archive.org/details/classicpcgames

PR #2017, opened 2023-02-05
zenity: Limit description length

SmallJoker:
The description column is too wide. This commit attempts to shorten it a bit.

The problem looks as follows:
grafik

With this PR the year and publisher columns will not appear but the description column is shorter now.

### DIFF
--- a/src/winetricks
+++ b/src/winetricks
@@ -12899,7 +12899,7 @@ load_mfc140()
 #----------------------------------------------------------------
 
 w_metadata vcrun2017 dlls \
-    title="Visual C++ 2017 libraries (concrt140.dll,mfc140.dll,mfc140u.dll,mfcm140.dll,mfcm140u.dll,msvcp140.dll,msvcp140_1.dll,msvcp140_2.dll,msvcp140_atomic_wait.dll,vcamp140.dll,vccorlib140.dll,vcomp140.dll,vcruntime140.dll,vcruntime140_1.dll)" \
+    title="Visual C++ 2015-2017 redistributable (VC2015 + msvcp140_2.dll)" \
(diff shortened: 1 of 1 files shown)


austin987:
Hi @SmallJoker.

Thanks for your submission. And thanks in particular for looking at bugs in the GUI. While I agree this is a problem for the GUI, I don't think this is the right approach. This approach will 'fix' vcrun20xx for GUI users (but potentially confuse CLI users / break any scripts depending on the file lists), but more importantly, won't address the problem in the future if another verb is added that has a 'too long' description.

Instead, I'd suggest using `head -c ..` or some other way of making a general fix that will cut off long descriptions for the GUI.

src/winetricks
```diff
@@ -4033,6 +4033,10 @@ winetricks_showmenu()
                         installed=TRUE
                         echo "${code}" >> "${WINETRICKS_WORKDIR}"/installed.txt
                     fi
+                    if [ "${#title}" -gt 100 ]; then
+                        title=$(printf "%s" "${title}" | head -c 90)
```
austin987, line 4037: What's the reasoning behind choosing two different numbers here?

SmallJoker: It's a hysteresis to not replace the last few characters or last word by three dots.

austin987: Makes sense; could you please add a comment to that effect, for future reference?

austin987: I'd also suggest changing 90 > 95, but I'll leave that to your discretion since you're actually using the GUI ;).

SmallJoker: Alright. PR updated.

austin987:
Thanks!

Issue #2023, opened 2023-02-13
unable to install riched20 and gdiplus 


tannenfreund87:
When trying to install riched20, I get the following error. It doesn't matter if I try in a 32 bit or 64 bit wineprefix. I cleared the winetricks cache, so all the files were redownloaded. I also tried multiple times with cleared cache and an empty wineprefix each time. I'm running opensuse tumbleweed with up-to-date packages and winetricks 20230212-next

```
Executing load_riched20 
grep: warning: stray \ before /
Executing cabextract -q -d /home/tannenfreund/.wine/dosdevices/c:/windows/temp -L -F i386/riched20.dl_ /home/tannenfreund/.cache/winetricks/win2ksp4/W2KSP4_EN.EXE
/home/tannenfreund/.wine/dosdevices/c:/windows/temp/i386/riched20.dl_: No such file or directory
------------------------------------------------------
warning: Note: command cabextract -q -d /home/tannenfreund/.wine/dosdevices/c:/windows/temp -L -F i386/riched20.dl_ /home/tannenfreund/.cache/winetricks/win2ksp4/W2KSP4_EN.EXE returned status 1. Aborting.

```

Trying to install gdiplus gives me the same error:

```
> winetricks gdiplus
Using winetricks 20230212-next - sha256sum: 8f8e5595d76dd220fe71ba06cda65cd81223f2305cdc9f90560e17b75c01889a with wine-7.12 and WINEARCH=win32
Executing w_do_call gdiplus
Executing load_gdiplus 
grep: warning: stray \ before /
Executing cabextract -q -d /home/tannenfreund/.wine/dosdevices/c:/windows/temp -L -F x86_microsoft.windows.gdiplus_6595b64144ccf1df_1.1.7601.17514_none_72d18a4386696c80/gdiplus.dll /home/tannenfreund/.cache/winetricks/win7sp1/windows6.1-KB976932-X86.exe
/home/tannenfreund/.wine/dosdevices/c:/windows/temp/x86_microsoft.windows.gdiplus_6595b64144ccf1df_1.1.7601.17514_none_72d18a4386696c80/gdiplus.dll: No such file or directory
------------------------------------------------------
[...]

PR #2025, opened 2023-02-15
changing cabextract arguments due to its recent update

haldai:
`cabextract` has just been updated to 1.10 and breaks the installation of `riched20`, `gdiplus`, `quarks`, etc., see issue https://github.com/Winetricks/winetricks/issues/2023. Adding the `--keep-symlinks` argument can solve this problem.

### DIFF
--- a/src/winetricks
+++ b/src/winetricks
@@ -703,7 +703,11 @@ w_try_cabextract()
         w_die "Cannot find cabextract.  Please install it (e.g. 'sudo apt-get install cabextract' or 'sudo yum install cabextract')."
     fi
 
-    w_try cabextract -q --keep-symlinks "$@"
+    if cabextract -h 2>&1 | grep -q 'keep-symlinks'; then
+        w_try cabextract -q --keep-symlinks "$@"
+    else
+        w_try cabextract -q "$@"
+    fi
 }
 
 w_try_cd()
--- a/src/winetricks
+++ b/src/winetricks
@@ -703,7 +703,7 @@ w_try_cabextract()
         w_die "Cannot find cabextract.  Please install it (e.g. 'sudo apt-get install cabextract' or 'sudo yum install cabextract')."
     fi
 
-    w_try cabextract -q "$@"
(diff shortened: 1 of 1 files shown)


austin987:
That will break users on older cabextract, that doesn't support that option.

My suggestion would be to gate it behind a check like `cabextract --help 2>&1 | grep -q keep-symlink` so you only use the option if it's supported.

Ideally, someone using (older) Proton could test this, since that was the entire reason for cabextract adding symlink support.

haldai:
> That will break users on older cabextract, that doesn't support that option.
> 
> My suggestion would be to gate it behind a check like `cabextract --help 2>&1 | grep -q keep-symlink` so you only use the option if it's supported.
> 
> Ideally, someone using (older) Proton could test this, since that was the entire reason for cabextract adding symlink support.

Indeed, thanks for your suggestion! I have added a test for this and tried with the `cabextract-1.9.1-2` downgraded via Archlinux Package Archive.

austin987:
Thanks! FYI I'm going to do some testing/think about it before merging :).

austin987:
Looking into this more, I think it's a bug in cabextract (already fixed).

The bug was introduced by https://github.com/kyz/libmspack/commit/89ac9df371c2e9c2909549d6459ef35e0d90a7a5, and fixed by https://github.com/kyz/libmspack/commit/cc09dd3cd01d4293a77497352e79562d00118bc6. See https://github.com/kyz/libmspack/issues/42#issuecomment-1442858743

I don't think unconditionally adding `--keep-symlinks` is the right fix here. That will cause cabextract to overwrite the contents of the symlinked file (which I suspect wouldn't work for older proton, as the symlink targets are likely root owned) and may cause other issues.

Instead, this should be reported to distros so they can backport the fix (or, avoid the broken cabextract versions for now).

evren320:
seems cabextract 1.11 fixes the problem

haldai:
> seems cabextract 1.11 fixes the problem

That's great! I'll close this PR.

PR #2026, opened 2023-02-15
Fix w_store_winver when $WINE contains spaces

SmallJoker:
While at it, I tried to find and fix more issues with shellcheck... but the script gave no output; so I assume it's alright?

Tested and reproduced with the `dash` POSIX shell and `WINE="/path/with spaces/wine"`


austin987:
> While at it, I tried to find and fix more issues with shellcheck... but the script gave no output; so I assume it's alright?

Yeah; shellcheck is run as part of the Github Actions that run before changes are merged, so it should generally be clean (until there's a new shellcheck version, of couse ;)).

> Tested and reproduced with the `dash` POSIX shell and `WINE="/path/with spaces/wine"`

Thanks!

Issue #1983, opened 2022-11-13
Feature request : Add dxvk-nvapi


xPakrikx:
Add installation of dxvk-nvapi if its possible :) 
[https://github.com/jp7677/dxvk-nvapi](https://github.com/jp7677/dxvk-nvapi)

Very thanks

chaitan3:
Would be nice if this could be updated to the latest version like dxvk and vkd3d, maybe even have a mechanism to always use the latest version from github?

PR #2033, opened 2023-02-24
Issue #1983: Add a dxvk-nvapi verb

ThisNekoGuy:
Please let me know if I overlooked anything or otherwise did something wrong; I _**think**_ I did this right but I'm not sure.

dxvk-nvapi is (mostly) a replacement for the physx verb, because it supports physx calls, but I didn't want to touch the physx verb since I'd be unsure of the potential impact; though, it is pretty safe to say that the physx verb is otherwise deprecated (but do correct me if you know more than I do).

### DIFF
--- a/src/winetricks
+++ b/src/winetricks
@@ -6910,7 +6910,7 @@ helper_dxvk()
     _W_dll_overrides="$(echo "${_W_dll_overrides}" | sed 's/d3d10 /&d3d10_1 /')"
     _W_package_dir="${_W_package_archive%.tar.gz}"
     _W_package_version="${_W_package_dir#*-}"
-    w_warn "Please refer to ${_W_repository#*/} version ${_W_package_version} release notes... See: https://github.com/${_W_repository}/releases/tag/${_W_package_version}"
+    w_warn "Please refer to ${_W_repository#*/} version ${_W_package_version} release notes... See: https://github.com/${_W_repository}/releases/tag/v${_W_package_version}"
     w_warn "Please refer to current dxvk base graphics driver requirements... See: https://github.com/doitsujin/dxvk/wiki/Driver-support"
 
     if w_wine_version_in ",${_W_min_wine_version}" ; then
--- a/src/winetricks
+++ b/src/winetricks
@@ -6931,7 +6931,7 @@ helper_dxvk()
         w_try mv "${W_TMP}/${_W_package_dir}/x32/${_W_dll}.dll" "${W_SYSTEM32_DLLS}/"
         [ "${_W_dll}" = "d3d9" ] && _W_d3d9_support="1"
     done
-    
+
     if test "${W_ARCH}" = "win64"; then
         for _W_dll in ${_W_dll_overrides}; do
             w_try mv "${W_TMP}/${_W_package_dir}/x64/${_W_dll}.dll" "${W_SYSTEM64_DLLS}/"
@@ -8248,7 +8248,7 @@ helper_dxvk_nvapi()
     for _W_dll in ${_W_dll_overrides}; do
         w_try mv "${W_TMP}/${_W_package_dir}/x32/${_W_dll}.dll" "${W_SYSTEM32_DLLS}/"
     done
-    
+
     if test "${W_ARCH}" = "win64"; then
         for _W_dll in ${_W_dll_overrides}; do
             w_try mv "${W_TMP}/${_W_package_dir}/x64/${_W_dll}.dll" "${W_SYSTEM64_DLLS}/"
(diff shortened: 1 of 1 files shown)


src/winetricks
```diff
@@ -8215,6 +8215,69 @@ load_dxvk()
...
+
+    _W_supported_overrides="nvapi nvapi64"
+    _W_invalid_overrides="$(echo "${_W_dll_overrides}" | awk -vvalid_overrides_regex="$(echo "${_W_supported_overrides}" | sed 's/ /|/g')" '{ gsub(valid_overrides_regex,""); sub("[ ]*",""); print $0 }')"
+    if [ "${_W_invalid_overrides}" != "" ]; then
+        w_die "parameter (4) unsupported dll override: '${_W_invalid_overrides}' ; supported dll overrides: ${_W_supported_overrides}"
+    fi
+
+    _W_package_dir="${_W_package_archive%.tar.gz}"
+    _W_package_version="v${_W_package_dir#*-v}"
+    w_warn "Please refer to ${_W_repository#*/} version ${_W_package_version} release notes... See: https://github.com/${_W_repository}/releases/tag/${_W_package_version}"
```
austin987, line 8237: Why the difference from dxvk? They end up with the same output, but dxvk_nvapi sets the "v" prefix in the variable assignment, while dxvk sets it in the warning itself?

src/winetricks
```diff
@@ -8215,6 +8215,69 @@ load_dxvk()
...
+    if [ "${_W_package_archive##*.}" = "zip" ]; then
+        w_try_unzip "${W_TMP}" "${W_CACHE}/${W_PACKAGE}/${_W_package_archive}"
+    else
+        w_try_cd "${W_TMP}"
+        w_try tar -zxf "${W_CACHE}/${W_PACKAGE}/${_W_package_archive}"
+    fi
+
+    for _W_dll in ${_W_dll_overrides}; do
+        w_try mv "${W_TMP}/${_W_package_dir}/x32/${_W_dll}.dll" "${W_SYSTEM32_DLLS}/"
+    done
```
austin987, line 8249: Please add a newline here.

ThisNekoGuy:
@austin987 Adding your requested newline(s) seems to displease the shellcheck github action

austin987:
> @austin987 Adding your requested newline(s) seems to displease the shellcheck github action

It should be a new line without trailing spaces (currently there are 4 empty spaces).

austin987:
Hi @ThisNekoGuy,

Thanks for fixing the CI checks.

So, took a slightly closer look, now that it looks more stable. I'm unclear on some of the changes. I'll make more detailed comments inline.

src/winetricks
```diff
@@ -8207,14 +8208,78 @@ load_dxvk()
 {
     # https://github.com/doitsujin/dxvk
     _W_dxvk_version="$(w_get_github_latest_release doitsujin dxvk)"
-    _W_dxvk_version="${_W_dxvk_version#v}"
-    w_linkcheck_ignore=1 w_download "https://github.com/doitsujin/dxvk/releases/download/v${_W_dxvk_version}/dxvk-${_W_dxvk_version}.tar.gz"
+    _W_dxvk_version="v${_W_dxvk_version#v}"
+    w_linkcheck_ignore=1 w_download "https://github.com/doitsujin/dxvk/releases/download/${_W_dxvk_version}/dxvk-${_W_dxvk_version}.tar.gz"
```
austin987, line 8212: Hi again @ThisNekoGuy. I think the original change was fine. I didn't see the explanation from your commit message, just the diff itself. My apologies for misunderstanding.

Please adjust the commit message though, to something like:
"dxvk: fix link to release notes"

(doesn't have to be exactly that, feel free to tweak/explain, but commit message should clearly indicate dxvk, not just a line number).

src/winetricks
austin987: To clarify, it's okay if things differ. I wanted to know why. Ideally, the code should be shared as much as possible, but differences happen.

That said, changes to dxvk itself should be avoided (unless it's a bug fix, or a change to _share_ code).

src/winetricks
ThisNekoGuy: [Like this?](https://github.com/Winetricks/winetricks/pull/2033/commits/e5f977eeafa4fc8ec32d9ab684d407af88f2c8c8)

src/winetricks
ThisNekoGuy: The only change I made to DXVK was how the package version variable collected the version tag; the value is only used for referencing the release notes here, so... what's the issue?

Issue #2030, opened 2023-02-23
winetricks breaks if used with wine build with --enable-architectures=i386,x86_64 / winetricks_set_wineprefix()


austin987:
In this setup, the prefix is 64-bit:
```
$ ls .wine/drive_c/
 ProgramData  'Program Files'  'Program Files (x86)'   users   windows

$ ls .wine/drive_c/windows/
command       globalization  inf        Microsoft.NET  performance  rundll.exe  system.ini  temp          twain_64     winhlp32.exe
explorer.exe  help           Installer  mono           regedit.exe  system      syswow64    twain_32      twain.dll    win.ini
Fonts         hh.exe         logs       notepad.exe    resources    system32    tasks       twain_32.dll  winhelp.exe  winsxs
```

But there's no `wine64` binary:
```
$ file /opt/oldwownew/wine-8.0/bin/* | grep -v 'shell script'
/opt/oldwownew/wine-8.0/bin/wine:           ELF 64-bit LSB executable, x86-64, version 1 (SYSV), dynamically linked, interpreter /lib64/ld-linux-x86-64.so.2, BuildID[sha1]=2f7a2fb204c876e5afb2159a1fee1be79c3bb2eb, for GNU/Linux 3.2.0, with debug_info, not stripped
/opt/oldwownew/wine-8.0/bin/wine-preloader: ELF 64-bit LSB executable, x86-64, version 1 (SYSV), statically linked, BuildID[sha1]=f0ae5ff9d792141653276bf560fbd50d7fed5a8a, with debug_info, not stripped
/opt/oldwownew/wine-8.0/bin/wineserver:     ELF 64-bit LSB pie executable, x86-64, version 1 (SYSV), dynamically linked, interpreter /lib64/ld-linux-x86-64.so.2, BuildID[sha1]=4f79bb850598068f37fcc55ba1eac37e99c8e662, for GNU/Linux 3.2.0, with debug_info, not stripped
```

In this configuration, though, the wine binary is 64-bit, not 32-bit. wine-preloader/wineserver are also 64-bit, not 32-bit.

winetricks_set_wineprefix() needs to be adjusted for this case. Given that WINE can point to a shell script (for in tree builds), it's probably best to check the bitness of the wineserver binary (but this needs testing).

See also: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1031649

austin987:
[...]

PR #2035, opened 2023-02-25
Workaround missing /usr/bin/wine64.

jre-wine:
Workaround case where wine and related binaries are in /usr/bin, but wine64 is only in Wine's bindir (as in Debian wine 8.0~repack-4). Use /usr/bin/wineboot (which is a symlink in Debian packaging) to figure out the bindir, because /usr/bin/wine is a wrapper script in Debian.

See discussion in https://bugs.debian.org/1031649.

I'm not sure if this already should be added in Winetricks, probably this depends on the next set of Debian Wine packages. For now I applied this patch in Debian.

Tested with the deb packages from Debian wine 8.0\~repack-2 and 8.0\~repack-4, and Winehq winehq-staging 8.2~bookworm-1.

### DIFF
--- a/src/winetricks
+++ b/src/winetricks
@@ -5063,6 +5063,19 @@ winetricks_set_wineprefix()
         # Check the bitness of wineserver + wine binary, used later to determine if we're on a WOW setup (no wine64)
         # https://github.com/Winetricks/winetricks/issues/2030
         WINESERVER_BIN="$(which "${WINESERVER}")"
+
+        # wineboot often is a link pointing to wineapploader in Wine's bindir. If we don't find binaries we may look for them there later
+        if [ -n "${READLINK_F}" ]; then
+            true
+        elif [ "$(uname -s)" = "Darwin" ]; then
+            # readlink exists on MacOS, but does not support "-f" on MacOS < 12.3
+            # Use perl instead
+            READLINK_F="perl -MCwd=abs_path -le 'print abs_path readlink(shift);'"
+        else
+            READLINK_F="readlink -f"
+        fi
+        WINEBOOT_BIN="$(dirname "${WINESERVER_BIN}")/$(basename "${WINESERVER_BIN}"|sed 's,wineserver,wineboot,')"
+
         _W_wineserver_binary_arch="$(winetricks_get_file_arch "${WINESERVER_BIN}")"
         WINE_BIN="$(which "${WINE}")"
         _W_wine_binary_arch="$(winetricks_get_file_arch "${WINE_BIN}")"
--- a/src/winetricks
+++ b/src/winetricks
@@ -5102,10 +5102,18 @@ winetricks_set_wineprefix()
         elif command -v "${WINE}64" >/dev/null 2>&1; then
             WINE64="${WINE}64"
         else
-            # Handle case where wine binaries (or binary wrappers) have a suffix
-            WINE64="$(dirname "${WINE}")/"
-            [ "${WINE64}" = "./" ] && WINE64=""
-            WINE64="${WINE64}$(basename "${WINE}" | sed 's/^wine/wine64/')"
+            if [ -x "${WINEBOOT_BIN}" ]; then
+                WINE_BINDIR="$(dirname "$(${READLINK_F} "${WINEBOOT_BIN}" 2>/dev/null)" 2>/dev/null)"
+                if [ -x "${WINE_BINDIR}/wine64" ]; then
+                    # Workaround case where wine is in path, but wine64 is only in Wine's bindir
+                    WINE64="${WINE_BINDIR}/wine64"
+                fi
+            else
+                # Handle case where wine binaries (or binary wrappers) have a suffix
+                WINE64="$(dirname "${WINE}")/"
+                [ "${WINE64}" = "./" ] && WINE64=""
+                WINE64="${WINE64}$(basename "${WINE}" | sed 's/^wine/wine64/')"
+            fi
         fi
         WINE_ARCH="${WINE64}"
         WINE_MULTI="${WINE}"


austin987:
I appreciate the patch, but not everyone has wine installed in /usr/bin ;).

This will need a more general solution (see also #2030).

Gcenx:
Also only the gnu version of readlink supports -f, using that option would effectively break winetricks on platforms that don’t use gnu utility’s.

austin987:
> Also only the gnu version of readlink supports -f, using that option would effectively break winetricks on platforms that don’t use gnu utility’s.

FWIW, the version of readlink that I see on macOS 12.6 (gcc104 from https://cfarm.tetaneutral.net/machines/list/):
`macOS 12.6                                                       June 22, 2017                                                      macOS 12.6`

supports `readlink -f`. Do you happen to know if that is recent in OSX/in other versions? The binary there is from /usr/bin, and comes from BSD, so I don't think it's homebrew/etc. thing, but I could definitely be wrong.

Gcenx:
Hum `readlink -f` does work on macOS Ventura (13.x) but it doesn’t work on macOS Mojave (14.x)

So maybe macOS Monterey (12.x) and macOS Ventura (13.x) have a new enough version of `readlink`.

Might need to require `coreutils` package on macOS for `greadlink` for legacy versions of macOS.

jre-wine:
I just want to clarify that Winetricks won't be "broken" if readlink -f is not working. Only this specific workaround wouldn't work.
And of course this workaround is only a fix for a specific setup as in current Debian unstable, but it is no fix for a completely missing wine64 as we might see if wine is compiled the new way (--enable-archs=i386,x86_64).

austin987:
Does https://github.com/austin987/winetricks/commit/05c43b58d2191958064516ec9406d837d8cf54e8 work for the debian package?

jre-wine:
Unfortunately not, even with the -preloader packages installed. Their binaries are also only installed to Wine's bindir (I just had a quick glance at the commit, but I assume that's the issue).

```
$ ./src/winetricks
warning: Unknown wine-loader binary type ("cannot open `' (No such file or directory)"). Please open a bug and attach output of "./src/winetricks -v"
$ ./src/winetricks -v
[...]
+ latest_version=20230212
+ echo 20230212
+ grep -q -E [0-9]{8}
+ [ -z 20230212 ]
+ [ ! 20230212-next = 20230212 ]
+ [ ! 20230212-next = 20230212-next ]
+ W_ISO_MOUNT_ROOT=/mnt/winetricks
+ W_ISO_USER_MOUNT_ROOT=/home/jens/winetricks-iso
+ W_ISO_MOUNT_LETTER=i
+ WINE=wine
+ command -v wine-preloader
+ dirname wine
+ test -x 
+ test -x wine-preloader
+ test -x 
+ test -x ./loader/wine-preloader
+ file 
+ _W_file_output=cannot open `' (No such file or directory)
+ w_die Unknown wine-loader binary type ("cannot open `' (No such file or directory)"). Please open a bug and attach output of "./src/winetricks -v"
+ w_warn Unknown wine-loader binary type ("cannot open `' (No such file or directory)"). Please open a bug and attach output of "./src/winetricks -v"
+ [ -z  ]
+ printf %s\nwarning: %b\n%s\n ------------------------------------------------------ Unknown wine-loader binary type ("cannot open `' (No such file or directory)"). Please open a bug and attach output of "./src/winetricks -v" ------------------------------------------------------
warning: Unknown wine-loader binary type ("cannot open `' (No such file or directory)"). Please open a bug and attach output of "./src/winetricks -v"
+ test 
+ unset _W_timeout
+ exit 1
+ winetricks_cleanup
+ test 
+ set +e
+ test -f /tmp/winetricks.lohAWTaw/w.jens.103921/dd-pid
+ test 
+ [  = 1 ]
```

kenorb:
I've tested on the latest Debian Docker container and PR seems to fix the problems as per <https://github.com/Winetricks/winetricks/issues/2119#issuecomment-1902678526>.

melroy89:
I have the same problem with [WineGU](https://github.com/winegui/WineGUI/issues/21#issuecomment-1950570842), it impact various distros now. And you're indeed correct there is no wine64 binary, but only a `wine` wrapper script.

I hope a final solution can soon be released for Winetricks, solving this urgent issue once and for all.

melroy89 (approved):

jre-wine:
I updated this PR for both issues mentioned here before:

Instead of hardcoding /usr/bin/wineboot I use WINESERVER_BIN to figure out wineboot's path. If this doesn't work there's no regression, just a missed opportunity to figure out where wine64 might be after other attempts also failed.

readlink exists on MacOS, but does not support "-f" on MacOS < 12.3. So I use perl there now.
I use "perl" without checking if it is installed at all. (Even though it is installed on MacOS per default since a certain version that I forgot, it still might've been uninstalled). But perl is already used in other places of Winetricks without special checks, and it looks cleaner that way.

There are further readlink-related optimizations possible which I might implement with the help of some MacOS users if someone thinks it's worth the effort (I don't). See https://stackoverflow.com/questions/1055671/how-can-i-get-the-behavior-of-gnus-readlink-f-on-a-mac
- If we know the MacOS version, we might use "readlink -f" on current MacOS.
- If I know the exact syntax and filepath we might use greadlink (provided by a separate package)
- As fallback we might use "stat" - but the syntax I found on the web doesn't work on Linux.

melroy89 (approved):

jre-wine:
Updated to allow for parent dirs containing wineserver in their name (same as in PR 2189)
Now I use dirname for the dir, and basename+sed for the filename (to allow for suffixes in the filename.

melroy89:
Hopefully @austin987 now agrees with this.

jre-wine:
I moved the basics for determining Wine's bindir (readlink and wineboot) to a place where it can be used both for determining the wineserver and wine arch (https://github.com/Winetricks/winetricks/pull/2189), and the wine64-only-in-BINDIR workaround (this PR). These basics are in a seperate commit that is shared by both PRs.

austin987 (approved):
Thanks @Chiitoo, and sorry for the delay in reviewing!

PR #2037, opened 2023-02-27
Fix 'shell' and 'winecmd' commands

Matoking:
Two small fixes to `winecmd` and `shell` commands:

* Do not use `-e` parameter when launching terminal emulator if no command
was provided. This is the case for `shell` verb, which does not provide
any predefined command and should just launch the terminal.
  In addition, pass the `-e` argument as one quoted string. Some terminal emulators only accept one value, while others consume all parameters following the option. Using a single string works with both.

* Fix `winecmd` launch. `winecmd` seems to be the intended command here. `cmd` only installed an
application of the same name without launching it.


austin987:
Thanks!

PR #2044, opened 2023-03-10
galliumnine09: new verb

dhewg:
The new version was released because of binary compatibility with wine v8.3.
Extend the current galliumnine verbs to handle the incompatible combinations.

### DIFF
--- a/src/winetricks
+++ b/src/winetricks
@@ -10107,6 +10107,7 @@ load_galliumnine02()
 {
     w_package_broken "https://bugs.winehq.org/show_bug.cgi?id=49676" 5.13
     w_package_broken "https://github.com/iXit/wine-nine-standalone/issues/83" 5.14
+    w_package_broken "https://github.com/iXit/wine-nine-standalone/issues/149" 8.3
 
     w_download "https://github.com/iXit/wine-nine-standalone/releases/download/v0.2/gallium-nine-standalone-v0.2.tar.gz" 6818fe890e343aa32d3d53179bfeb63df40977797bd7b6263e85e2bb57559313
     helper_galliumnine "${file1}"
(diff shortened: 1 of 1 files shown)


dhewg:
The spellcheck failure is because of the result of w_get_github_latest_release in an URL, I don't see a way around that

src/winetricks
```diff
@@ -10212,23 +10220,49 @@ w_metadata galliumnine08 dlls \
...
     installed_file2="${W_SYSTEM32_DLLS_WIN}/ninewinecfg.exe" \
     homepage="https://github.com/iXit/wine-nine-standalone"
 
 load_galliumnine()
 {
-    _W_galliumnine_version="$(w_get_github_latest_release iXit wine-nine-standalone)"
-    w_linkcheck_ignore=1 w_download "https://github.com/iXit/wine-nine-standalone/releases/download/${_W_galliumnine_version}/gallium-nine-standalone-${_W_galliumnine_version}.tar.gz"
+    if w_wine_version_in ,5.7 ; then
+        _W_galliumnine_version="v0.8"
+        w_download "https://github.com/iXit/wine-nine-standalone/releases/download/${_W_galliumnine_version}/gallium-nine-standalone-${_W_galliumnine_version}.tar.gz" 8d73dcf78e4b5edf7a3aea8c339459b5138acd1c957c91c5c06432cb2fc51893
```
austin987, line 10260: The 'fix' would be to use 'w_linkcheck_ignore=1' before 'w_download', as the not old case does.

That said, since you're going to hardcode it for older wine (which makes sense), you can skip the variable and put that in the code directly, i.e.,:
`w_download "https://github.com/iXit/wine-nine-standalone/releases/download/v0.8/gallium-nine-standalone-v0.8.tar.gz" 8d73dcf78e4b5edf7a3aea8c339459b5138acd1c957c91c5c06432cb2fc51893`

austin987: edit: the variable will still be needed, when calling helper_galliumnine(), but you don't need to use the variable for w_download().

dhewg: D'oh, right.
Pushed another solution, which is even more obvious :)

Issue #2004, opened 2023-01-15
can you please add d8vk? 


stefson:
hello everyone, 

d8vk is a fork of dxvk, with a focus on directx8 and directx9. 

It offers `d3d8.dll` and `d3d9.dll`, and is certainly very helpfull for anyone stuck with low performance of d3d8 on modern hardware drivers. 

this is the homepage: https://github.com/AlpyneDreams/d8vk

and this is the release page: https://github.com/AlpyneDreams/d8vk/releases

current release is `0.10` 

I did a quick copy and paste of the dxvk-2.0 code, only thing I don't understand is the `helper_dxvk` line. And I don't know the real name of the person who publishes this. 

```

w_metadata d8vk0010 dlls \
    title="Vulkan-based D3D8/D3D9 implementation for Linux / Wine (0.10)" \
    publisher=" AlpyneDreams " \
    year="2022" \
    media="download" \
    file1="d8vk-v0.10.tar.gz" \
    installed_file1="${W_SYSTEM32_DLLS_WIN}/d3d8.dll" \
    installed_file2="${W_SYSTEM32_DLLS_WIN}/d3d9.dll" 

load_d8vk0010()
{
    # https://github.com/AlpyneDreams/d8vk/
    w_download "https://github.com/AlpyneDreams/d8vk/releases/download/d8vk-v0.10/d8vk-v0.10.tar.gz" 43e3d1051b31d91541fba1a6c59d0991208869203e66779bfab8a861eb545eb1
    helper_dxvk "${file1}" "7.1" "1.3.204" "dxgi,d3d9,d3d10core,d3d11"
}

```

austin987:
helper_dxvk() is used to validate the needed Wine/Vulkan API versions. It looks like d8vk needs similar? https://github.com/doitsujin/dxvk/wiki/Driver-support

I'd suggest looking at extending helper_dxvk() to also allow overriding d3d8 for d8vk.

FYI @bobwya

sonic2kk:
I don't think anyone was looking at this yet? If not, I poked around and got an implementation working and hope to get a PR out soon for review :-)

PR #2045, opened 2023-03-10
d8vk: new verb

sonic2kk:
Adds d8vk verb as requested in #2004.

Tested with Touhou 6 via Lutris, MangoHud showed the renderer changed from WINED3D to DXVK. The selection of D3D8 games I can test with is somewhat limited it seems so further testing is of course welcome.

This PR touches `helper_dxvk`, I was specifically concerned with the changes to the `tar` command which now uses `--one-top-level`, so I tested installing regular `dxvk` and it seemed to work fine. This extra flag is required because DXVK archives have a root folder in the archive, but D8VK does not.

One outstanding issue is with the URL in `helper_dxvk` that points to the release notes; this URL is incorrect for D8VK. This is because DXVK release tags are just the version, e.g. `v2.1`, but D8VK is `d8vk-vX.X`. The `_W_package_version` needs to be set differently and I wasn't sure of the best approach to take with this.

First time contribution and first time even looking at Winetricks code so any and all feedback is very welcome :-)

Thanks!

### DIFF
--- a/src/winetricks
+++ b/src/winetricks
@@ -8439,8 +8439,23 @@ load_d8vk()
     # https://github.com/AlpyneDreams/d8vk
     _W_dxvk_version="$(w_get_github_latest_release AlpyneDreams d8vk)"
     w_linkcheck_ignore=1 w_download "https://github.com/AlpyneDreams/d8vk/releases/download/${_W_dxvk_version}/${_W_dxvk_version}.tar.gz"
-    helper_dxvk "${_W_dxvk_version}.tar.gz" "7.1" "1.3.204" "d3d8,d3d9" "AlpyneDreams/d8vk"  # Not completely sure on min Vulkan version
+    helper_dxvk "${_W_dxvk_version}.tar.gz" "7.1" "1.3.204" "d3d8,d3d9" "AlpyneDreams/d8vk"
     unset _W_dxvk_version
+}
+
+w_metadata d8vk0010 dlls \
+    title="Direct3D 8 to Vulkan translation for DXVK" \
+    publisher="Jeffery Ellison" \
+    year="2023" \
+    media="download" \
+    file1="d8vk-v0.10.tar.gz" \
+    installed_file1="${W_SYSTEM32_DLLS_WIN}/d3d8.dll" \
+    installed_file2="${W_SYSTEM32_DLLS_WIN}/d3d9.dll" \
+
+load_d8vk0010()
+{
+    w_download "https://github.com/AlpyneDreams/d8vk/releases/download/d8vk-v0.10/d8vk-v0.10.tar.gz" 43e3d1051b31d91541fba1a6c59d0991208869203e66779bfab8a861eb545eb1
+    helper_dxvk "${file1}" "7.1" "1.3.204" "d3d8,d3d9" "AlpyneDreams/d8vk"
 }
 
 #----------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


src/winetricks
```diff
@@ -6912,14 +6912,14 @@ helper_dxvk()
...
-        w_try tar -zxf "${W_CACHE}/${W_PACKAGE}/${_W_package_archive}"
+        w_try tar -zxf "${W_CACHE}/${W_PACKAGE}/${_W_package_archive}" --one-top-level
     fi
 
     for _W_dll in ${_W_dll_overrides}; do
         w_try mv "${W_TMP}/${_W_package_dir}/x32/${_W_dll}.dll" "${W_SYSTEM32_DLLS}/"
         [ "${_W_dll}" = "d3d9" ] && _W_d3d9_support="1"
     done
-    if test "${W_ARCH}" = "win64"; then
+    if test "${W_ARCH}" = "win64" && -d "${W_TMP}/${_W_package_dir}/x64"; then
```
sonic2kk, line 6922: Direct3D 8 is only for 32bit applications, so the archive only has a `x32` directory, hence this check.

sonic2kk:
I think the checks should pass now

sonic2kk:
That was embarrassing, sorry... Guess I changed something since the last time I ran shellcheck or something.

Passes on my machine now, gave the same warning that the CI did so I guess it's fixed :slightly_smiling_face:

src/winetricks
```diff
@@ -6912,14 +6912,14 @@ helper_dxvk()
         w_try_unzip "${W_TMP}" "${W_CACHE}/${W_PACKAGE}/${_W_package_archive}"
     else
         w_try_cd "${W_TMP}"
-        w_try tar -zxf "${W_CACHE}/${W_PACKAGE}/${_W_package_archive}"
+        w_try tar -zxf "${W_CACHE}/${W_PACKAGE}/${_W_package_archive}" --one-top-level
```
austin987, line 6915: `--one-top-level` is a GNU extension; this will break on BSD tar.

sonic2kk: Thanks! After some light research it looks like `-C` might work for BSD tar but I haven't tested just yet.

sonic2kk: Updated to use `-C "${_W_package_dir}"`, Winetricks still fetches and installs d8vk correctly (tested with no cached archive and with a cached archive), only tested Touhou 6 at the moment and it still works too.

sonic2kk:
Since merging master the diff seems like it's gotten pretty big? It mentions removing `helper_dxvk_nvapi` which I saw when resolving the conflicts, but I just chose to accept the incoming changes which evidently removed it. This function is also not present on master it looks like from a Ctrl+F.

Not sure what went wrong there :frowning:

austin987:
There was a merge (dxvk_nvapi) that conflicted, so it needs a manual merge.

I did some initial fixes (﻿https://github.com/austin987/winetricks/commit/01c9188bc8683e483be40f6138e68cd65d464bc9), but didn't tested it with a real application.

Note that the release notes are still broken (should be fixed). Also, this will always point to the latest dxvk release; you probably want to add at least one real release so users have a stable option they can use.

sonic2kk:
Not sure why this is only working for some games and crashing others. Maybe the supported games list only refers to specific versions of games. Touhou 6 works for me, but Syberia II does not. The D3D9 DLL gets overwritten for all the games it looks like, but Touhou 6 works still. I tested different Wine-GE versions and Lutris Wine 7.2 and d8vk consistently installed properly. 

I noticed the install script will move the DLLs into the Proton lib folders for Steam games as well as into the games prefix, but as far as I can tell this is not something Winetricks does at all, so I don't think it'a step required here.

Maybe that limits the usefulness of including d8vk in Winetricks. I am not sure what would be going wrong with the code, as it mostly appears to mirror how DXVK is installed, and D8VK works for at least one game.

PR #2049, opened 2023-03-22
Adding Bulgarian Translation

RacerBG:
Hello there, this is the Bulgarian translation for Winetricks.

This PR is ready for review although I need help with one specific thing related to WinRAR. Currently I'm using the default English version of it because I have no idea how to see the checksums (?) for the Bulgarian version. Could someone give me a hand for this (write the checksums here)? The lines are 17055 and 17057. Thanks in advance!

Also, I would like to ask if this is everything I need to do, for example do I need to edit other files?

Regards,
Georgi

### DIFF
--- a/src/winetricks
+++ b/src/winetricks
@@ -77,7 +77,7 @@ WINETRICKS_VERSION=20230212-next
 #   Copyright (C) 2013-2017 Andrey Gusev <andrey.goosev!gmail.com>
 #   Copyright (C) 2013-2020 Hillwood Yang <hillwood!opensuse.org>
 #   Copyright (C) 2013,2016 André Hentschel <nerv!dawncrow.de>
-#   Copyright (C) 2023 Georgi Georgiev (RacerBG) <g.georgiev.shumen@gmail.com>
+#   Copyright (C) 2023 Georgi Georgiev (RacerBG) <g.georgiev.shumen!gmail.com>
 #
 # License:
 #   This program is free software; you can redistribute it and/or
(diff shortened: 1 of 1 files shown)


austin987:
> Hello there, this is the Bulgarian translation for Winetricks.

Great, thanks for working on this!
 
> This PR is ready for review although I need help with one specific thing related to WinRAR. Currently I'm using the default English version of it because I have no idea how to see the checksums (?) for the Bulgarian version. Could someone give me a hand for this (write the checksums here)? The lines are 17055 and 17057. Thanks in advance!

Looks like you figured it out, but it's a sha256sum. Generally, `sha256sum $filename` if you have GNU coreutils, other OSes may have a different binary (osx/bsd).

> Also, I would like to ask if this is everything I need to do, for example do I need to edit other files?

`src/winetricks` is the primary file to worry about. But your changes triggered some shellcheck warnings that will need to be fixed before this can be merged (looks like 'smart quotes' were used instead of regular quotes). See https://www.shellcheck.net/wiki/SC1111

RacerBG:
@austin987 Thank you very much for the explanation. Knowing that the checksums are SHA256 was all I needed to add the correct ones. I'm assuming that WinRAR version 6.11 is used by Winetricks here. Also I had fixed the quotes, Bulgarian uses different quotes than English but it is not a big deal really. This PR is probably done if you don't have further recommendations for me.

PS: I can't squash commits if asked, I'm doing this via the GitHub website.

Regards,
Georgi

austin987:
Thanks!

PR #2055, opened 2023-04-10
winetricks_set_wineprefix: warn user if running on OSX m1/m2 and Rosetta2 is not installed

austin987:
There's obviously a lot more room for improvement for winetricks/osx, but as a first step..

@Gcenx could I get your thoughts?

### DIFF
--- a/src/winetricks
+++ b/src/winetricks
@@ -4999,6 +4999,14 @@ winetricks_set_wineprefix()
     # Make sure the prefix is initialized:
     w_try winetricks_early_wine cmd /c "echo init" > /dev/null 2>&1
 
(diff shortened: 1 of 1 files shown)


Gcenx:
I’d need to run it to verify but it seems correct at least. It’s actually possible to check that Rosetta2 is installed.

<br>

Not sure I’d directly say m1/m2 maybe (M Series/Apple Silicon)

Apple no longer uses the term OSX it’s now macOS since Sierra (10.12 released September 20, 2016), it’s better sticking to the term “macOS” as the packages @gverm and myself are maintaining only support macOS High Sierra and later.

austin987:
Updated (and now GPG signed); not in a rush if you'd like to test it first.

Gcenx:
Seems fine

PR #2073, opened 2023-05-24
Add missing filename to `dotnet6` verb

yyny:
As it stands this code tries to execute the non-existant file `""` which fails with a confusing error, but the install works correctly when the filename is specified.


austin987:
Thanks!

Issue #2077, opened 2023-06-07
VKD3D-Proton verb broken


AlfredoRodda:
Hello,

VKD3D-Proton version 2.9 introduced a change regarding the bundled D3D12 DLLs. More information about the release and changes can be found here (https://github.com/HansKristian-Work/vkd3d-proton/releases/tag/v2.9). 

In summary, VKD3D-Proton's "d3d12core.dll" needs to be moved and overridden like how Winetricks currently handles the d3d12.dll. 

I am not able to create a pull request. If the following changes are made in the Winetricks script then VKD3D-Proton 2.9 is fine.

1.  

Before 
```
# $1 - vkd3d-proton archive name (required)
helper_vkd3d_proton()
{
_W_package_archive="${1}"

_W_dll_overrides="d3d12"
```

After
```
# $1 - vkd3d-proton archive name (required)
helper_vkd3d_proton()
{
_W_package_archive="${1}"

_W_dll_overrides="d3d12 d3d12core"
```

2.
Before 
```
w_metadata vkd3d dlls \
    title="Vulkan-based D3D12 implementation for Linux / Wine (latest)" \
    publisher="Hans-Kristian Arntzen " \
    year="2020" \
    media="download" \
    installed_file1="${W_SYSTEM32_DLLS_WIN}/d3d12.dll"
```

After
```
w_metadata vkd3d dlls \
    title="Vulkan-based D3D12 implementation for Linux / Wine (latest)" \
    publisher="Hans-Kristian Arntzen " \
    year="2020" \
    media="download" \
    installed_file1="${W_SYSTEM32_DLLS_WIN}/d3d12.dll" \
    installed_file2="${W_SYSTEM32_DLLS_WIN}/d3d12core.dll"
```

Thank you,

AlfredoRodda:
Resolved as of #2081, thank you all.

PR #2081, opened 2023-06-14
vkd3d: Add d3d12core

Alexander88207:
Version 2.9 introduced a change regarding the bundled D3D12 DLLs.

Reported by: @AlfredoRodda in #2077


austin987:
Thanks!

PR #2082, opened 2023-06-24
support new wow64 mode

theofficialgman:
wow64 mode only builds a `wine` binary. the previous checking would result in a soft error in `winetricks_set_wineprefix` which would eventually result in `cmd.exe /c echo '%AppData%' returned empty string`

### DIFF
--- a/src/winetricks
+++ b/src/winetricks
@@ -5013,8 +5013,8 @@ winetricks_set_wineprefix()
     if test -d "${W_DRIVE_C}/windows/syswow64"; then
         # Probably need fancier handling/checking, but for a basic start:
         # Note 'wine' may be named 'wine-stable'/'wine-staging'/etc.):
-        # WINE64 = wine64, available on 64-bit prefixes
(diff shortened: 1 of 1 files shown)


austin987:
I don't think this is the right approach. While it does solve the immediate case, I think we want to know if the user is using the new style wow64 (it doesn't work in all cases, and I suspect that we may need to take different code paths in the future based on if a user is using it or old style wow64.

See https://github.com/Winetricks/winetricks/issues/2030 for more discussion about this.

theofficialgman:
I personally haven't noticed any issues yet (except dotnet20 hangs during installation which as noted in that issue is likely a wine bug) with winetricks with the patch implemented. For bug reporting though I agree this should probably be distinguished.

austin987:
I've pushed a test patch, if you'd like to try it, see https://github.com/Winetricks/winetricks/issues/2030#issuecomment-1621095879

I'm going to close this PR, as I don't think this is the right way to approach it. But thanks for your submission, and I'd welcome your feedback/help in #2030.

PR #2086, opened 2023-07-01
Add name to workflows and update checkout version

QuintenQVD0:
- Add a name to the workflows, so GitHub does not just show the file.
- update the actions/checkout version from v2 to v3 so the actions warning goes away


austin987:
Thanks!

Issue #2094, opened 2023-07-07
Move Wiki to Github or Link to Wiki


kekkoudesu:
I just discovered that there is a rich wiki for working on Winetricks available at the old repository host: https://code.google.com/archive/p/winetricks/wikis

If I knew about this yesterday, my life would have been a lot easier! I was moving through the script in Vim, trying to figure out what each function and variable did and how they connected to each other, when I could have read these pages and gotten up to speed much more quickly:

* https://code.google.com/archive/p/winetricks/wikis/AddingNewVerbs.wiki
* https://code.google.com/archive/p/winetricks/wikis/Functions.wiki
* https://code.google.com/archive/p/winetricks/wikis/Variables.wiki

Some of this information is likely to be outdated or incomplete, but it's a very good start!

I think the README should either link to this wiki, or it should be moved to Github's wiki. That would make it a lot easier to onboard new contributors.

Chiitoo:
I could add them to https://wiki.winehq.org/Winetricks if that makes sense, at least the information that is still true.

kekkoudesu:
The entire Adding New Verbs article was very useful to me and is great at outlining the process. While the Functions/Variables articles might be incomplete, they're a good place to start. The [Autohotkey article](https://code.google.com/archive/p/winetricks/wikis/AutoHotkey.wiki) was also just the thing I was looking for when trying to write an AHK script before I found the Unattended Mode section in Adding New Verbs. There are a few other miscellaneous articles that have useful, timeless information like Commadnline Tips.

Moving that information to the WineHQ wiki is a better idea than moving it to the Github wiki as it's a more permanent home, although maybe some of these articles should be on separate, dedicated pages for better readability (the article is currently a user's guide rather than a contributor's guide)?

austin987:
I don't think it makes sense to move them to the Wine wiki. winetrick is a separate project, and its documentation should be on GitHub alongside the project/code.

The wiki page on WineHQ is intended for Wine users.

austin987:
I've re-enabled the wiki; it will need some cleanup. Maybe you could help with that @kekkoudesu ?

kekkoudesu:
Sure, I can make a start.

I think the WineHQ wiki page should link to this Github wiki. I don't think I can create an account on the WineHQ Wiki—could either of you make this edit to the page?

kekkoudesu:
I've been trying to push to the Github wiki with git, but have encountered some trouble:

```
[...]

PR #2095, opened 2023-07-10
Update manpage and help with latest options

kekkoudesu:
New options were added to winetricks over time, but the manpage hasn't been updated. This commit brings it up-to-date.

`games list` is removed because there are no longer any games verbs. Also references Github Wiki discussed in #2094  as Winetricks documentation.


austin987:
Thanks!

PR #2101, opened 2023-07-20
Fix remove_mono breaks dotnet40 on subsequent dotnet* installs

### LINKED ISSUE #2072: remove_mono breaks dotnet40 on subsequent dotnet* installs
On every call to `remove_mono`, the script checks to see whether Mono is actually installed. If it's not, the function is supposed to bail out before the end where it removes the registry keys and `mscoree.dll`. However, that check has a subtle bug - if dotnet40 is installed (which installs its own `mscoree.dll`) and then, for example `dotnet48` is installed, the `elif` is skipped and the `dotnet40` installation gets broken. Moving the registry key and file removals inside the `if` block fixes the problem for me.

deriamis:
We should not be removing mscoree.dll in all dotnet* installs because it could have been placed by the dotnet40 verb, which would cause subsequent dotnet* verbs to break dotnet40. The deletion should instead only happen when the native dotnet is detected. We continue to remove the registry keys in any case, though, since they are only relevant to the native DLL.

Closes #2072

### DIFF
--- a/src/winetricks
+++ b/src/winetricks
@@ -18758,18 +18758,15 @@ load_remove_mono()
         for uuid in ${mono_uuid}; do
             "${WINE_ARCH}" uninstaller --remove "${uuid}"
         done
+        "${WINE_ARCH}" reg delete "HKLM\\Software\\Microsoft\\NET Framework Setup\\NDP\\v3.5" /f || true
+        "${WINE_ARCH}" reg delete "HKLM\\Software\\Microsoft\\NET Framework Setup\\NDP\\v4" /f || true
+
+        w_try rm -f "${W_WINDIR_UNIX}/system32/mscoree.dll"
+        if [ "${W_ARCH}" = "win64" ]; then
+            w_try rm -f "${W_WINDIR_UNIX}/syswow64/mscoree.dll"
+        fi
     elif [ -z "$1" ] || [ "$1" != "internal" ]; then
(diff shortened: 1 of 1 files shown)


deriamis:
> Hi. I have the following suggestion, leaving it outside of the `if`.
> 
> ```shell
>     if grep --quiet --text "WINE_MONO_OVERRIDES" ${W_WINDIR_UNIX}/system32/mscoree.dll; then
>         w_try rm -f "${W_WINDIR_UNIX}/system32/mscoree.dll"
>     fi
>     if grep --quiet --text "WINE_MONO_OVERRIDES" ${W_WINDIR_UNIX}/syswow64/mscoree.dll; then
>         w_try rm -f "${W_WINDIR_UNIX}/syswow64/mscoree.dll"
>     fi
> ```
> 
> I arrived at this because I think I witnessed it still being removed while trying to debug things, but I can't say right now if it's really a thing or not. In any case, this should make it idempotent in a way.

I agree that this is better. I'll update the PR when I get off work. Thanks!

austin987:
Please fix the quoting errors that shellcheck pointed out; the current code will break for prefixes that contain special characters.

deriamis:
@austin987 I've addressed the `shellcheck` issues. Sorry about that!

qwertychouskie:
@austin987 Any chance of getting this merged?

austin987:
Yes, sorry for the delay, and thanks @deriamis for the patch!

PR #2110, opened 2023-08-23
Update mfc140 to load the same version of VC2015 as vcrun2015

SurenIV:
mfc140 download the different version of VC2015 than vcrun2015. Changed URL and checksum.


austin987:
Thanks!

PR #2112, opened 2023-08-29
Update winetricks

pi3141:
update link for windowscodecs exe to avoid a 404 errror


austin987:
I don't think this makes sense to do. This effectively reverts bfe1cf676674fe7cef718a088119a4a36c16d98d (because IIRC microsoft took this download offline). While archive.org sometimes has issues, it's working now.

Moreover, this only made the change to 32-bit, not 64-bit.

PR #2117, opened 2023-09-13
d3dcompiler_47: Download from mozila github repo

Gcenx:
Directly download d3dcompiler_47 dlls instead of extracting them from Firefox installers.


austin987:
I noticed that the shasums of the binaries are different (and doesn't seem to match any of the binaries in the history).

That said, it's mozilla and GPG signed now, and the DLL has had a lot of updates, so probably fine. Just noting it in case there's a regression from this :).

PR #2124, opened 2023-09-20
Fix CoreFonts Broken Mirrors Issue

rayanmargham:
Changed Mirror from mirrors.kernel to github.com/pushcx/corefonts


austin987:
Thanks for the patch, but I went with #2127 which uses the same websites and just fixes the URLs.

Issue #2118, opened 2023-09-16
Stuff such for like fonts are gone from mirrors.kernel.org/gentoo


Alexander88207:
Hello,

i just wanted to install corefonts or arial fonts then saw that there is a mismatch.

When you browse the url then you get an 404 message.

Some mirror suggestion: https://github.com/pushcx/corefonts

rayanmargham:
Yea this is a big issue I experienced this as well

rayanmargham:
cannot install corefonts

rayanmargham:
@austin987 please look into this issue! outdated broken mirror

rayanmargham:
I fixed it and purposed a pull request #2124

Chiitoo:
The files are still there, but organised under different directories since 2019 or so, and the compatibility symlinks were recently removed [1], and for example, 'andale32.exe' can be found under '7d' [2] now.

1. https://bugs.gentoo.org/914092
2. https://mirrors.kernel.org/gentoo/distfiles/7d/

illegalcoding:
Yep, just experienced this. I could fix it by downloading them from the wayback machine and quietly swapping out the executable while it was waiting for a response on if it should continue with an invalid hash
It was getting an invalid hash since it downloaded a 404 page and put it into the exe file

austin987:
https://github.com/Winetricks/winetricks/commit/afd1782abe2f2ac7c480292d73049667faebb77f

PR #2127, opened 2023-09-26
Update corefont urls to new locations on mirror.

calheb:
Fixes issue #2118, while still pulling fonts from the original mirror (at their new URLs).

### DIFF
--- a/src/winetricks
+++ b/src/winetricks
@@ -14191,7 +14191,7 @@ w_metadata andale fonts \
 
 load_andale()
 {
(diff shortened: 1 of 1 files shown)


fsworld009:
Thanks, I patched winetricks my flatpak lutris with this branch and now `winetricks corefonts` is successful.

Note: Use `which winetricks` in Bash terminal to find out where is it installed, for me it is `~/.var/app/net.lutris.Lutris/data/lutris/runtime/winetricks/winetricks`

austin987:
Thanks!

PR #2129, opened 2023-10-03
Update winetricks MS fonts urls

cooltg66:
i used the following commands to find the hash:
echo -n "andale32.exe" | b2sum
echo -n "arial32.exe" | b2sum
etc


cooltg66:
closed because i didn't saw #2127

PR #2134, opened 2023-10-06
dotnet7: bump to latest minor

xlanor:
Updated dotnet7 to latest version. computed sha256 hashes because sha512 is not supported and MSFT site only shows sha512.


austin987:
Thanks for the PR, but I committed https://github.com/Winetricks/winetricks/commit/90d1c6915cec6b2c6113d6187fda60d43c443b7b instead (as it was a new version).

PR #2153, opened 2023-11-16
crypt32: Install both 32 and 64 bit DLLs

qwertychouskie:
Use win7sp1 instead of winxpsp3, it easily provides both 32 bit and 64 bit DLLs, with the added benefit of likely being newer.

### DIFF
--- a/src/winetricks
+++ b/src/winetricks
@@ -6163,6 +6163,28 @@ load_crypt32()
 
 #----------------------------------------------------------------
 
+w_metadata crypt32_winxp dlls \
+    title="MS crypt32" \
+    publisher="Microsoft" \
+    year="2004" \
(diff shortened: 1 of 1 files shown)


austin987:
Could you rename the old crypt32 to crypt32_winxp, then add yours? That way if any users depend on the older version, they can continue to use it, if needed (a la gdiplus).

qwertychouskie:
> Could you rename the old crypt32 to crypt32_winxp, then add yours? That way if any users depend on the older version, they can continue to use it, if needed (a la gdiplus).

Will do

qwertychouskie:
Update pushed, ready to merge.

austin987:
Thanks!

Issue #2163, opened 2023-12-27
dotnet7 & dotnetdesktop7 verbs using unsupported shasums


RunnerFaith:
When attempting to install either dotnet7 or dotnet7desktop (32 or 64-bit) using winetricks, both verbs produce the following output which prevents their installation:
 ```
------------------------------------------------------
Using winetricks 20230212-next - sha256sum: 71fbd7803111173073e2c95c31616b36042de0608e5e3e89b8130e67967e1ad4 with wine-9.0-rc3 (Staging) and WINEARCH=win64
Executing w_do_call dotnetdesktop7
------------------------------------------------------
warning: You are using a 64-bit WINEPREFIX. Note that many verbs only install 32-bit versions of packages. If you encounter problems, please retest in a clean 32-bit WINEPREFIX before reporting a bug.
------------------------------------------------------
Executing load_dotnetdesktop7 
------------------------------------------------------
warning: unsupported shasum..bug
------------------------------------------------------
```

This seems to stem from the checksums provided in #2149, which use the official checksums provided when downloading the runtime on the .NET [website.](https://dotnet.microsoft.com/en-us/download/dotnet/7.0) From what I can gather, winetricks does not support validation of SHA512 checksums, which would explain the "unsupported shasum" message.

PR #2164, opened 2023-12-27
Fix dotnet7

QuintenQVD0:
Closes: #2163 

Now the sum is a sha256 sum, if you go to the URL you actually get a file and not an error.


RunnerFaith:
Thanks for this! I've tested both verbs locally and they seem to be working as expected now.

austin987 (approved):

Issue #2148, opened 2023-11-09
msxml3: Only installs 32-bit version


qwertychouskie:
`msxml3` only installs the 32 bit version, and it's also an older version than what's available.  Currently SP7 is installed.

SP10? 64-bit: https://catalog.s.download.windowsupdate.com/msdownload/update/software/secu/2010/07/windowsserver2003.windowsxp-kb2079403-x64-enu_2b9a05b19d3d62079ed2ed042d87ef315f6cc9b7.exe
SP10? 32-bit: https://catalog.s.download.windowsupdate.com/msdownload/update/software/secu/2010/07/windowsxp-kb2079403-x86-enu_bc5640c1e48194b88a0cd9635392c3ff3012f1d1.exe

@itsmattkc ran into an issue where installing `msxml3` via Winetricks wouldn't work for Adobe Premier to run, but manually copying `msxml3` from a Windows install worked.  After some investigation, I suspect the missing 64 bit DLL (and perhaps also the outdated version) is causing the issue.

I found the newer version by way of http://www.stuartgillibrand.co.uk/2011/08/list-of-msxml-versions.html

EDIT: found slightly newer version: https://www.microsoft.com/download/details.aspx?familyid=d147f865-6a56-4db2-8d78-14f2bee6da18 from https://learn.microsoft.com/en-us/security-updates/SecurityBulletins/2013/ms13-002?redirectedfrom=MSDN#security-update-deployment

EDIT again: even newer: https://learn.microsoft.com/en-us/security-updates/SecurityBulletins/2014/ms14-033?redirectedfrom=MSDN

PR #2176, opened 2024-01-26
msxml3: Update to newer version; add 64-bit support

qwertychouskie:
Closes https://github.com/Winetricks/winetricks/issues/2148


qwertychouskie:
Hmm, it seems win7sp1 has a newer version, and that version is needed for Premier.  I'll update this PR

qwertychouskie:
Working locally, should be ready for review.

Issue #2178, opened 2024-01-26
winetricks dotnetdesktop8 not work please solve


amidevous:
winetricks dotnetdesktop8

not work on fedora and ubuntu for install net framework 8

work online on archlinux

tested wine wine and winehd-stable package

please solve problem

```
winetricks dotnetdesktop8
```

```
------------------------------------------------------
warning: You are using a 64-bit WINEPREFIX. Note that many verbs only install 32-bit versions of packages. If you encounter problems, please retest in a clean 32-bit WINEPREFIX before reporting a bug.
------------------------------------------------------
Using winetricks 20240105-next - sha256sum: cd868b5357fc4d22a0c811687280759fb4575ab7e3687da5ab972c66c9614d76 with wine-9.0 and WINEARCH=win64
Unknown arg dotnetdesktop8
Usage: /usr/bin/winetricks [options] [command|verb|path-to-verb] ...
Executes given verbs.  Each verb installs an application or changes a setting.

Options:
    --country=CC      Set country code to CC and don't detect your IP address
-f, --force           Don't check whether packages were already installed
    --gui             Show gui diagnostics even when driven by commandline
    --gui=OPT         Set OPT to kdialog or zenity to override GUI engine
    --isolate         Install each app or game in its own bottle (WINEPREFIX)
    --self-update     Update this application to the last version
    --update-rollback Rollback the last self update
-k, --keep_isos       Cache isos (allows later installation without disc)
    --no-clean        Don't delete temp directories (useful during debugging)
    --optin           Opt in to reporting which verbs you use to the Winetricks maintainers
    --optout          Opt out of reporting which verbs you use to the Winetricks maintainers
-q, --unattended      Don't ask any questions, just install automatically
-r, --ddrescue        Retry hard when caching scratched discs
-t  --torify          Run downloads under torify, if available
    --verify          Run (automated) GUI tests for verbs, if available
-v, --verbose         Echo all commands as they are executed
-h, --help            Display this message and exit
-V, --version         Display version and exit

Commands:
list                  list categories
list-all              list all categories and their verbs
apps list             list verbs in category 'applications'
benchmarks list       list verbs in category 'benchmarks'
dlls list             list verbs in category 'dlls'
fonts list            list verbs in category 'fonts'
settings list         list verbs in category 'settings'
list-cached           list cached-and-ready-to-install verbs
list-download         list verbs which download automatically
list-manual-download  list verbs which download with some help from the user
list-installed        list already-installed verbs
[...]

PR #2179, opened 2024-01-26
winetricks dotnetdesktop8 not work please solve

amidevous:
winetricks dotnetdesktop8 not work please solve

solve issues

https://github.com/Winetricks/winetricks/issues/2178

require emergengie update for all linux distribution

for add support netframework 8.0 support

### DIFF
--- a/src/winetricks
+++ b/src/winetricks
@@ -10155,6 +10155,54 @@ load_dotnetdesktop7()
     fi
 }
 
+w_metadata dotnet8 dlls \
+    title="MS .NET Runtime 8.0 LTS" \
+    publisher="Microsoft" \
+    year="2023" \
+    media="download" \
+    file1="dotnet-runtime-7.0.14-win-x86.exe" \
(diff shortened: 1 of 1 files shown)


amidevous:
update include all version of net 7 and net 8

it is necessary for net framework 7 and 8

to make available all versions

because when a program requests version 7.0.10

7.0.0 or 7.0.15 does not work

please rapid update on this git for add update work for all with

sudo winetricks --self-update

austin987:
This PR needs work:
* You have 5 commits just called 'Update winetricks', commits should specify exactly what they do, e.g., `dotnet715: new verb`
* I don't love having to add several .net versions, but having dealt with this recently at my day job, I'm aware that it's not as backwards compatible as expected. That said, the generic `dotnet7` shouldn't hardcode a version, but instead call the latest version (i.e., add dotnet715, then have dotnet7 call dotnet715).
* Please break the commits into smaller changes, following the above guidelines.

PR #2180, opened 2024-01-27
Add otvdm verb

DarkShadow44:
This adds a verb for https://github.com/otya128/winevdm
This is a fork of winevdm (using an emulator) with way better compatibility. Since it needs a bunch of overrides, I do think this could be useful.

### DIFF
--- a/src/winetricks
+++ b/src/winetricks
@@ -11945,6 +11945,141 @@ load_openal()
     w_try "${WINE}" "${W_TMP}/oalinst.exe" /silent
 }
 
+
+#----------------------------------------------------------------
+
+# $1 - otvdm archive name (required)
+helper_otvdm()
+{
+    _W_package_archive="${1}"
+    _W_package_dir="${_W_package_archive%.zip}"
+
+    w_try_unzip "${W_TMP}" "${W_CACHE}/${W_PACKAGE}/${_W_package_archive}"
+    w_try_cp_dll "${W_TMP}/${_W_package_dir}/otvdm.exe" "${W_SYSTEM32_DLLS}/otvdm.exe"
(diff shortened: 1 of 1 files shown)


src/winetricks
```diff
@@ -11945,6 +11945,141 @@ load_openal()
...
+    w_override_dlls native,builtin win87em.dll16
+    w_override_dlls native,builtin wing.dll16
+    w_override_dlls native,builtin winnls.dll16
+    w_override_dlls native,builtin winoldap.mod16
+    w_override_dlls native,builtin winsock.dll16
+    w_override_dlls native,builtin winspool.drv16
+    w_override_dlls native,builtin wow32
+}
+
+w_metadata otvdm_0_9_0 dlls \
```
austin987, line 12059: Can we use otvdm090 to match what we do for dotnet/dxvk/etc.?

austin987:
Thanks for working on this; I was thinking about adding this myself and hadn't gotten around to it yet :)

src/winetricks
DarkShadow44: I guess we could, but doesn't that risk duplicate names?

austin987: Possibly if they change their naming scheme? If there's a future 0.90.0, that would be otvdm0900, rather than otvdm090 (granted, still confusing). Or am I missing something?

DarkShadow44: Yeah I only added those underlines because I was afraid of duplicate or weird names like that. If you think that's not a big deal I can rework that?

austin987: Yes, please. It should only be a problem if they change their versioning scheme (and have a lot more releases).

DarkShadow44: Sure, updated

PR #2181, opened 2024-01-28
Determine if there is an internet connection first using getent befor…

GloriousEggroll:
…e attempting to download anything.

This method uses getent which is part of glibc, a pretty much hard requirement for any linux system.

Since winetricks is capable of both downloading and installing software as well as performing library overrides or registry edits, this is useful when you run a script that performs multiple winetricks with a mixture of all types (such as a lutris install script or protontricks) and don't have an internet connection. This way anything that does not require an internet connection is not forced to wait on those that do to fail before continuing.

### DIFF
--- a/src/winetricks
+++ b/src/winetricks
@@ -2950,10 +2950,26 @@ _EOF_
 
 #---- Private Functions ----
 
+export INTERNET="no"
+internet_check() {
+  hasinternet=$(getent hosts cloudflare.com)
+  # Check for internet connection
+  if [ "${hasinternet}" != '' ]; then
+    export INTERNET="yes"
+  fi
+}
+
 # Determines downloader to use, etc.
 # I.e., things common to w_download_to(), winetricks_download_to_stdout(), and winetricks_stats_report())
 winetricks_download_setup()
 {
+    if [ "${INTERNET}" = "no" ] ; then
+        internet_check
+        if [ "${INTERNET}" = "no" ] ; then
+            w_die "No internet connection found, aborting download attempt."
+        fi
+    fi
+
     # shellcheck disable=SC2104
     case "${WINETRICKS_DOWNLOADER}" in
         aria2c|curl|wget|fetch) : ;;


austin987:
I have a few concerns with this:
* getent is available on most Linux systems, but not macOS (or possibly other OSes)
* querying cloudlare.com for every winetricks usage raises privacy concerns
* the variable naming don't match winetricks conventions

That aside, this just seems like a dirty hack. If you don't have internet and don't expect to, you could prepopulate the download cache to avoid errors.

Winetricks already has WINETRICKS_DOWNLOADER_TIMEOUT which defaults to 15 seconds. Why not set that to 1?

GloriousEggroll:
I think expecting people to cache the entire thing in advance is a a bit of a stretch honestly. As for WINETRICKS_DOWNLOADER_TIMEOUT it still has to hit the timeout for every single download. With the getent method proposed it exports the envvar once so that it's set for the rest of the time the script runs in the same shell, so you would not in fact have it pinging cloudflare multiple times.  

That being said I understand the concern of randomly pinging cloudflare to begin with, as well as the compatibility issue with no glibc on mac. Kind of stuck between a rock and a hard place here. If theres a similar mechanism on mac we could simply do a check to see which one it's being used on and use that mechanism. As for where it checks for DNS resolution -- if there's a safer option I don't see why it couldn't be changed.

austin987:
> I think expecting people to cache the entire thing in advance is a bit of a stretch honestly.

Well, yes. But I also don't think that users should expect things that require an internet connection to work if they don't have one. If they know they don't have internet, they can:
* cache the files in advance
* set WINETRICKS_DOWNLOADER_TIMEOUT to a reasonable value

> * As for WINETRICKS_DOWNLOADER_TIMEOUT it still has to hit the timeout for every single download.

Surely setting it to 1 second isn't *that* long of a wait. Especially for a corner case that isn't expected to work anyway.

GloriousEggroll:
> Well, yes. But I also don't think that users should expect things that require an internet connection to work if they don't have one. 

That's part of the problem, winetricks isn't ONLY for downloading and installing fixes. It also performs registry edits, and overrides as well. It's not uncommon to be running some script that does both.

Regardless it's clear this isn't a wanted approach. I'll go ahead and close the PR.

austin987:
I understand where you're coming from. However it's clearly an edge case, and I don't think it makes sense to make _every_ user ping a random website when they download something to accommodate this case, when there is already a workaround.

If this is a widespread problem that I'm not aware of, then we can revisit it.

Issue #1730, opened 2021-03-30
warning: You are using a 64-bit WINEPREFIX


DocMAX:
is there a way to disable this warning? nowadays 64bit gets more and more standard.

austin987:
There's not.

While it is more common/standard, a lot of verbs only install 32-bit dlls (and a lot of things don't work on 64-bit) so I think the warning is justified.

phibbs7:
Installing verb dotnet472 in a 64bit prefix requires clearing 18 of such warnings before the first installer dialog appears. I'd agree that such nagging is excessive.

If anything such numerous warnings just encourages users to either not bother with winetricks at all, or to edit the script to disable the warnings. It also gives users a poor impression of winetricks and wine in general because most consumer applications are 64bit nowadays. The tools coming up and declaring that 64bit support is half-baked at best, and to not even bother reporting bugs against it just leaves a bad taste in people's mouths. Let alone the constant stream of error messages making the tool itself seem to be bugged when in fact it is working as intended.

If such nagging is actually needed, maybe winetricks' 64bit support should better indicate in the selection menu what verbs won't work or will have issues in a 64bit prefix instead of annoying the users constantly. If support really is so hopeless maybe winetricks should just refuse to run in a 64bit prefix, unless a known working verb is given on the command line. (E.g. Disable the selection GUI in 64bit prefixes.)

TheKodeToad:
This is not a solution, but I forked this project to only display warnings in the terminal as a temporary workaround.

github12101:
Installing "all fonts" position require to click is about 100 times, this is madness!!! This warning should be suppressed to only one per session.

Chiitoo:
I think using the CLI is a rather painless "workaround" for things like "requiring about 100 clicks" (there's also '-q' or '--unattended' which makes things like 'dotnet' installs a lot more bearable).

Is the original comment here about these GUI clicks, or the warning in general (including the CLI)?

github12101:
> I think using the CLI is a rather painless "workaround" for things like "requiring about 100 clicks" (there's also '-q' or '--unattended' which makes things like 'dotnet' installs a lot more bearable).
> 
> Is the original comment here about these GUI clicks, or the warning in general (including the CLI)?

How I supposed to know what command install "all fonts" package? I launch winetricks directly from Lutris.

Chiitoo:
[...]

PR #2184, opened 2024-02-05
minor update to win64 warning messages to support toggling via env var

zpangwin:
This adds support to allow users to `export W_NO_WIN64_WARNINGS=1` for toggling off win64 warning messages (this new way is in addition to the file-based logic from [an earlier commit](https://github.com/austin987/winetricks/commit/01de762dfa656471cee0df3f6508b82a07de9117) and so should not break any existing setups).

This proposal and the advantages were previously discussed in the comments of issue #1730 but for convenience, those were:

> 1. Allows users to _completely_ suppress the message (including echo statement from touched file) if they _really_ want to
>
> 2. Allows to set an environment variable for a terminal session but not forever (or alternately could use `.bashrc` or equivalent for all terminal sessions)
> 
> 3. Reduced number of I/O hits checking for a file (I know it is an _extremely_ minor I/O hit but it's still _technically_ a plus)
> 
> 4. This also allows a user to temporarily turn the warning back on if they have created the file but wish to restore the warning temporarily such as for a single debugging session.

I think the actual changes should be pretty straight forward and keeping with the same coding style. But if there are any issues / questions / comments / etc, please let me know.

### DIFF
--- a/src/winetricks
+++ b/src/winetricks
@@ -5123,19 +5123,24 @@ winetricks_set_wineprefix()
         W_COMMONFILES_WIN="$(w_expand_env CommonProgramW6432)"
         W_COMMONFILES="$(w_pathconv -u "${W_COMMONFILES_WIN}")"
 
-        # 64-bit prefixes still have plenty of issues/a lot of verbs only install 32-bit libraries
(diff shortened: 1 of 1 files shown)


austin987:
Thanks!

PR #2185, opened 2024-02-11
Update vcrun2022 sha256 sum

QuintenQVD0:
In the latest release I got a sha256 mismatch error, so I downloaded the files and updated the sums

(ofc there lowercase here, stupid windows)

afbeelding


austin987:
Thanks!

Issue #2143, opened 2023-10-31
winetricks_get_file_arch path - od: wine: No such file or directory / Unknown file arch


randallpittman:
From the change made in #2030, the call

```bash
_W_wine_binary_arch="$(winetricks_get_file_arch "${WINE}")"
```

on line [5057](https://github.com/Winetricks/winetricks/blob/07c6d38913433c6318c4d5358ea75987b9a419f1/src/winetricks#L5057) doesn't work because `WINE==wine`, so in `winetricks_get_file_arch`, `od` fails with

```text
od: wine: No such file or directory
```

I'm not sure it's a sane workaround, but for me I just changed line 5057 to

```bash
_W_wine_binary_arch="$(winetricks_get_file_arch $(which "${WINE}"))"
```

ETA:
@theofficialgman posted perhaps a more robust solution on Issue 2023: <https://github.com/Winetricks/winetricks/issues/2030#issuecomment-1705755913>

GuillaumeCisco:
I can confirm too this patch fixes the issue.

kenorb:
Same errors:

```
#14 13.04 wine: configuration in L"/home/ubuntu/.wine" has been updated.
#14 13.05 Installing winhttp...
#14 13.25 od: wine: No such file or directory
#14 14.22 od: wine: No such file or directory
...
#14 23.35 Installing .NET...
#14 23.54 od: wine: No such file or directory
#14 24.50 od: wine: No such file or directory
```

Fuckingnameless:
same here, how can i fix it?

seisdr:
Jesus

DropTable69:
Nothin' more exciting than three-hundred and forty five popups while installing fonts.

Chiitoo:
> same here, how can i fix it?

You could try the changes from this pull request:

https://github.com/Winetricks/winetricks/pull/2147

> Nothin' more exciting than three-hundred and forty five popups while installing fonts.

You can avoid those by not using the GUI.

> Winetricks still broken

Yes, there are issues, though at least for me things still work even if I do get this error message as well.

I know not for how long things may remain as they are.  I have not seen any of the usual activity from the project owner, nor do I know of anyone else who knows their status.

I just hope they are okay... or better.

While I try to help around, I will not be merging any changes to 'master' or create releases, since we never agreed on that.

At most I might create my own branch where I fix things that I can, but a fork would lilkely be better still...

ActuallyTheOwner:
Still broken?
[...]

PR #2191, opened 2024-02-20
Rewrite new style wow64 support

vitlav:
This rewrite done in continue to #2030 and #2143 and pretent to be better improvement than #2189.

The first commit removed wine binary arch detection (these binaries can be wrapper scripts) and add new style wow64 detection by missed wine64 command.
The second commit fix environment expand macro to get correct %ProgramFiles% path.
And w_try_regsvr is renamed to w_try_regsvr32 to be more consisency with other 32/64 functions. Now some code is already using w_try_regsvr32 and this code is broken without this commit.

### DIFF
--- a/src/winetricks
+++ b/src/winetricks
@@ -1064,6 +1064,16 @@ w_expand_env()
     winetricks_early_wine_arch cmd.exe /c echo "%$1%"
 }
 
+w_expand_env32()
+{
+    winetricks_early_wine "${W_SYSTEM32_DLLS_WIN}\\cmd.exe" /c echo "%$1%"
+}
+
+w_expand_env64()
+{
+    WINE=${WINE64} winetricks_early_wine "${W_SYSTEM64_DLLS_WIN64}\\cmd.exe" /c echo "%$1%"
+}
+
 # Get the latest tagged release from github.com API
 w_get_github_latest_release()
 {
@@ -5063,9 +5073,6 @@ winetricks_set_wineprefix()
         WINE_MULTI="${WINE}"
         W_ARCH=win64
 
-        W_PROGRAMS_WIN="$(w_expand_env ProgramFiles)"
-        W_PROGRAMS_UNIX="$(w_pathconv -u "${W_PROGRAMS_WIN}")"
-
         # Common variable for 32-bit dlls on win32/win64:
         W_SYSTEM32_DLLS="${W_WINDIR_UNIX}/syswow64"
         W_SYSTEM32_DLLS_WIN="C:\\windows\\syswow64"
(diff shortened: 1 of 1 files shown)


src/winetricks
```diff
@@ -5093,14 +5059,20 @@ winetricks_set_wineprefix()
             WINE64="$(dirname "${WINE}")/"
             [ "${WINE64}" = "./" ] && WINE64=""
             WINE64="${WINE64}$(basename "${WINE}" | sed 's/^wine/wine64/')"
+            test -x "${WINE64}" || WINE64="${WINE}"
+        fi
+
+        # if we can't detect wine64 command, it is the new wow64 mode
```
Matoking, line 5065: The `wine-wow64` package on Arch AUR creates a `wine64` -> `wine` symlink, so this doesn't work in that situation:
https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=wine-wow64#n100

No idea if other platforms do this, though. I imagine they might do it for backwards compatibility.

Maybe we could check if `WINE64` is a symlink to `WINE`?

vitlav: > The `wine-wow64` package on Arch AUR creates a `wine64` -> `wine` symlink, so this doesn't work in that situation: https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=wine-wow64#n100
> 
> No idea if other platforms do this, though. I imagine they might do it for backwards compatibility.
> 
> Maybe we could check if `WINE64` is a symlink to `WINE`?

I would prefer they remove wine64 link, but we can add workaround too (check if WINE64 is a symlink to WINE).

Matoking: I'm not sure how common this scenario is. Most distros don't seem to ship the new WoW64 yet (as in, the `--enable-archs=x86_64,i386` is set in the build script), so I have no idea how common this symlink is. Fedora, Ubuntu and Arch don't seem to ship the new WoW 64 build.

Chiitoo: On Gentoo Linux, we allow for multiple Wine versions to be installed at the same time, including different flavours (Vanilla, Staging, and Proton for example), so in general 'wine' will be a symlink to '/etc/eselect/wine/bin/wine' which will be a link to the currently selected 'main' provider.

We also have a USE-flag to toggle the new style 'wow64' build, which is not handled differently with regards to the links, and as such, this change will not help Winetricks to detect either.

May be tricksy to catch them all.  : ]

vitlav: This check needed only for text message. I can just remove it. Our near future is new WoW64 in any case.

vitlav: Well, I have added a patch to remove wine64 checking and a bit obsoleted warning message too.

vitlav:
I've been testing these changes for a month now, and everything is working. Can someone else report the results of their testing?

jre-wine:
I tested it on Debian testing (wine 9.0~repack-4, old style WoW64 in /usr/lib/wine/, with /usr/bin/wine being a wrapper script and /usr/bin/wine64 not existing):

32-bit and 64-bit prefixes are detected correctly.

Your implementation removed the distinction between classic and new style WoW64. I don't know how important that is for debugging, but I prefer not knowing this over a non-working Winetricks.

Unfortunately the Debian-specific missing /usr/bin/wine64 still leads to
`warning: wine cmd.exe /c echo '%AppData%' returned empty string, error message "" `

vitlav:
Any suggestions?

austin987:
I don't think this is the right approach, while its simpler, it doesn't allow us to detect the different wine setups.

PR #2197, opened 2024-03-06
Added new verb: cades

eteriviv:


austin987:
What is this? Could you explain why it would be useful to Wine users/developers?

eteriviv:
It implements CAdES-X. Used in https://chrome.google.com/webstore/detail/cryptopro-extension-for-c/iifchhfnnmpdbibifmljnfjhpififfog and other cryptography products (generally, with CryptoPro). It seems this software is specific to Russian users. If this is not a common case, let's not accept it into the repository.

PR #2203, opened 2024-03-26
dotnet35sp1: Workaround for error if the username is "guest"

SurenIV:
Dotnet35sp1 will not install if your username is a guest.
Same issue with dotnet35, 30sp1, 30, 20sp1, 20sp2.

### DIFF
--- a/src/winetricks
+++ b/src/winetricks
@@ -1332,6 +1332,36 @@ w_dotnet_verify()
     w_info ".Net Verifier returned ${dn_status}"
 }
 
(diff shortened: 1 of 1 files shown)


austin987:
This seems too hacky to me.. Wine doesn't support multiple users (https://bugs.winehq.org/show_bug.cgi?id=11112), and mixing up users with in single WINEPREFIX opens up a large can of worms.

I also don't think it's a very common case to have a username of 'guest', but if so, I think the appropriate fix is to not name your user that if you need .NET. If that's a hard requirement, then you should carry this patch locally.

PR #2204, opened 2024-03-27
Refactored link checks

Root-Core:
This fixes some issues with the link checker.
The commit 2bc95a12954f591acc463602bb45fcef791048c0 is reverted, but I can't find a reason why it should be a problem. Feedback is appreciated. 

- Check w_download_to
- Check only URL, not whole line
- Find lines that don't start with w_download
- Allow escaped $ in URL
- Escape '\$' to '%24'

CC #2199

### DIFF
--- a/src/linkcheck.sh
+++ b/src/linkcheck.sh
@@ -1,5 +1,4 @@
 #!/bin/sh
-# shellcheck disable=SC2317
 
 # Link checker for winetricks.
 #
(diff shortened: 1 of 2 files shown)


src/linkcheck.sh
```diff
@@ -1,4 +1,6 @@
 #!/bin/sh
+# shellcheck disable=SC2317
```
austin987, line 2: Please keep it local to the affected functions instead of making it global.

Root-Core: Done. It was global, as it applies to multiple lines.. but this should be fine now.

Root-Core:
@austin987 Is there anything else that needs to be changed?

PR #2206, opened 2024-04-03
New verb "graphics", for enabling Wayland driver

gardenappl:
Hi, it's my first PR here. Added a verb for easily enabling the experimental Wayland driver in Wine 9.0+.

- Also usable for the same purposes as macdriver; deprecated macdriver
- Has Ukrainian title and (new language) Russian title
- Fixed settings with value =default being interpreted as DLLs
That last one was also affecting another setting, `videomemorysize=default`.

This setting seems important to me, so I put it at the top of the "desktop" category. Maybe it should be renamed to "graphics"? Or maybe it doesn't matter much, looks like those categories are just comments.

Note that the "wayland" option actually sets the driver to "wayland,x11", so Wine will try Wayland first, but should fallback to X11 if it's not available.

### DIFF
--- a/src/winetricks
+++ b/src/winetricks
@@ -3941,6 +3941,12 @@ winetricks_settings_menu()
                             *) title="${title_bg}";;
                         esac
                     ;;
+                    ru*)
+                        case "${title_ru}" in
+                            "") ;;
+                            *) title="${title_ru}";;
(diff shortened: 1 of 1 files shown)


austin987:
Thanks!

PR #2220, opened 2024-05-09
winetricks_set_wineprefix: replace "which" with "command -v"

jre-wine:


austin987:
Thanks; I enabled the shellcheck directive for this in 383292899865c41d7f9e5e77c3f09b73e72a0fde, to prevent future incidents.

PR #2228, opened 2024-05-22
winetricks: Update msvcrt40 description.

mohamadaljaf:


austin987:
You should update winetricks itself (those files are autoupdated on release, but updating them as part of the PR is fine too).

austin987:
Thanks!

PR #2231, opened 2024-05-30
msxml3: added downloading dll with resources for 64-bit

SurenIV:
Added downloading and copying msxml3r.dll in system32 for 64-bit prefixes.


src/winetricks
```diff
@@ -12002,7 +12002,9 @@ load_msxml3()
 
     if [ "${W_ARCH}" = "win64" ]; then
         helper_win7sp1_x64 amd64_microsoft-windows-msxml30_31bf3856ad364e35_6.1.7601.17514_none_e6944609ad75ac7d/msxml3.dll
+        w_download https://cloudflare-ipfs.com/ipfs/QmanVf9v1NDfBRieGmuHLNkZyoxndijitbTJxs9MyGVpCc?filename=msxml3r.dll 3efa5c3a8ee14c2de9f9ea5507a9950a0d991184dd723c79aca786e0b5fd7cf5 msxml3r.dll
```
austin987, line 12005: This doesn't seem like a reputable source (and also makes the verb out of sync with the 32-bit version).

austin987: That doesn't address the core issue. Taking a dll from windows and putting it on the internet violates Microsoft's copyright. We need a publicly redistributable binary.

SurenIV: An article from https://mskb.pkisolutions.com/kb/973687 states that resources can be found in update KB973687.
This matches msxml 3.0 SP10 (for Vista SP1) https://learn.microsoft.com/en-us/previous-versions/troubleshoot/msxml/list-of-xml-parser-versions. The win7sp1 update contains a newer version of msxml3 and probably does not require the msxml3r.dll resources.

qwertychouskie:
FWIW, this is the entire contents of the 64-bit msxml3r.dll from Windows 7:
image
It seems to just be a stub for compatibility or something like that.  It could probably be "recreated" trivially into a byte-for-byte matching version by just creating a DLL of the correct size and adding the relevant strings.  Though TBH, just hosting a copy of the file might be just fine, as it's IMHO pretty unlikely that it could even be considered copyrightable at all.

EDIT: And this is the 32-bit (syswow64) version for comparison:
image

PR #2261, opened 2024-09-01
Fix multiple verbs getting latest github release

arbruijn:
With multiple verbs on the command line that retrieve the latest github release, the version from the first verb would be used for the second verb.

Delete the temporary version information file before fetching it to avoid using the previous contents.

### DIFF
--- a/src/winetricks
+++ b/src/winetricks
@@ -1113,6 +1113,9 @@ w_get_github_latest_release()
     org="$1"
     repo="$2"
 
+    # release.json might still exists from the previous verb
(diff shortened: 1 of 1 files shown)


src/winetricks
```diff
@@ -1113,6 +1113,9 @@ w_get_github_latest_release()
     org="$1"
     repo="$2"
 
+    # release.json might still exists from the previous verb
+    rm -f "${W_TMP_EARLY}/release.json"
```
austin987, line 1117: Please add `w_try` before this command.

arbruijn: Done.

PR #2277, opened 2024-10-01
redo opensymbol download, remove w_try_ar

eteriviv:
w_try_ar was needed for only one component, so I redid it and now this function is not needed - fewer dependencies
opensymbol font now downloaded from openoffice open github repo.


austin987:
Nice, thanks!

PR #2278, opened 2024-10-08
Fix Unicode $USER

davispuh:
My `$USER` is `Dāvis` which contains Unicode and that causes `%AppData%` to be `C:\users\Dāvis\AppData\Roaming`
Currently `winetricks` doesn't work at all for me and fails with:
```
grep: (standard input): binary file matches
WINEPREFIX INFO:
Drive C: total 12
[...]

Registry info:
/mnt/Wine/system.reg:#arch=win64
/mnt/Wine/userdef.reg:#arch=win64
/mnt/Wine/user.reg:#arch=win64
warning: wine cmd.exe /c echo '%AppData%' returned empty string, error message ""
```

Reason is that because when `cmd.exe` pipes output it will use Windows Code Page which won't be recognized by Linux/grep (it thinks it's binary output). So this PR fixes it by enforcing that output is in UTF-8 instead which works correctly.

### DIFF
--- a/src/winetricks
+++ b/src/winetricks
@@ -1054,7 +1054,7 @@ w_pathconv()
 # Expand an environment variable and print it to stdout
 w_expand_env()
 {
-    winetricks_early_wine_arch cmd.exe /c echo "%$1%"
+    winetricks_early_wine_arch cmd.exe /c 'chcp 65001 > nul && echo "%$1%"'
 }
 
 # Determine what architecture a binary file is built for, silently continue in case of failure.
--- a/src/winetricks
+++ b/src/winetricks
@@ -1054,7 +1054,7 @@ w_pathconv()
 # Expand an environment variable and print it to stdout
 w_expand_env()
 {
-    winetricks_early_wine_arch cmd.exe /c 'chcp 65001 > nul && echo "%$1%"'
+    winetricks_early_wine_arch cmd.exe /c "chcp 65001 > nul && echo %$1%"
 }
 
 # Determine what architecture a binary file is built for, silently continue in case of failure.


davispuh:
You can verify it like this
```
$ wine64 cmd.exe /c echo Dāvis | hexdump -C
00000000  44 83 76 69 73 0d 0a                              |D.vis..|
$ wine64 cmd.exe /c "chcp 65001 > nul && echo Dāvis" | hexdump -C
00000000  44 c4 81 76 69 73 0d 0a                           |D..vis..|
```

As you can see in Windows Code Page `ā` is only single char 0x83 while in UTF-8 it's `0xc4 0x81`

austin987:
Hi @davispuh,

Thanks for the PR. While I'm inclined to improve Unicode support, I can't reproduce this.

I've added a user to my machine with a unicode name (`tëstuser`).

Then, I try to start wine, but wine can't every seem to install mono (or gecko):
`wineserver -k ; rm -rf ~/.wine ; wine notepad.exe`

Fails to install mono (it downloads, but then fails the checksum check. Similar for wine-gecko/wine-9.19.

Does this work for you? What wine version? I've never used a unicode username before, is there something else needed? I tried pre-prepopulating ~/.cache/wine with the .msi installers, but no change.

davispuh:
Are you saying you can't even get Wine to work with such username?

I'm on ArchLinux with wine-9.19 and Mono I have installed with Arch Linux package `wine-mono`
I also tried removing  `wine-mono` package but it doesn't look like Notepad even needs it...

I recommend just using cmd.exe and not Notepad.

It's fully reproducible for me
```
$ sudo useradd --create-home --badname tëstuser
$ sudo su -l tëstuser
$ env | grep -E "^HOME|^LANG"
HOME=/home/tëstuser
LANG=lv_LV.UTF-8
$ wine cmd.exe /c echo "%AppData%"
C:\users\tëstuser\AppData\Roaming
$ wine cmd.exe /c echo "%AppData%" | hexdump -C
00000000  43 3a 5c 75 73 65 72 73  5c 74 3f 73 74 75 73 65  |C:\users\t?stuse|
00000010  72 5c 41 70 70 44 61 74  61 5c 52 6f 61 6d 69 6e  |r\AppData\Roamin|
00000020  67 0d 0a                                          |g..|
00000023
```

austin987:
> Are you saying you can't even get Wine to work with such username?
> 
> I'm on ArchLinux with wine-9.19 and Mono I have installed with Arch Linux package `wine-mono` I also tried removing `wine-mono` package but it doesn't look like Notepad even needs it...

Notepad itself will run without it. I was attempting to run the test suite with this change, to make sure it doesn't break anything else, when I noticed the mono/gecko issue.

Trying again today, I'm not getting a checksum error (seems like others were, so maybe that was a temporary winehq issue, https://bugs.winehq.org/show_bug.cgi?id=57331). I do notice that it does install, but it doesn't cache the installer (as it does with my normal user).

I think the reason you haven't seen this is the Arch wine-mono package installed. If I put the gecko/mono installers in `/usr/local/share/wine/`, then it works as expected.

I opened a Wine bug for the installer not caching: https://bugs.winehq.org/show_bug.cgi?id=57338

I'd like to make sure this is something Wine supports before adding a workaround to Winetricks.

rmi1974:
This can't work properly on Wine releases < 7.7 since it also requires UTF8 support/changes in Wine code.

See my comment https://github.com/Winetricks/winetricks/issues/2409#issuecomment-3536784393

Since people are reverting this change in their winetricks forks, I suggest to put a fallback (old impl) for all older Wine releases.

Regards

PR #2280, opened 2024-10-11
Update URL for ucrtbase2019 to remove hardcoded archive.org prefix

MadByteDE:
It seems like the `archive.org` prefix is automatically added in the download function (on second try), meaning there shouldn't be a need for hardcoding the `archive.org` URL prefix.

I've tested a quick game installation using the downloaded EXE from the Microsoft servers & everything seemed to run fine.

### Note
Theoretically there could be another reason for the permanent  URL prefix - e.g. the setup file on the Microsoft servers changed at some point, no longer containing a working `ucrtbase.dll`. Play-testing Tarkov didn't reveal any issues with the dll..
edit: since the file is checked via `sha256sum` - this shouldn't be a concern either.

### DIFF
--- a/src/winetricks
+++ b/src/winetricks
@@ -13964,7 +13964,7 @@ load_ucrtbase2019()
     w_override_dlls native,builtin ucrtbase
 
     # Microsoft download no longer containts ucrtbase so get the last known version from archive.org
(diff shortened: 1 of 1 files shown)


MadByteDE:
To clarify: `ucrtbase.dll` is provided in the microsoft download:
grafik

austin987:
Thanks!

PR #2302, opened 2024-11-20
Updated load_7zip to Install the Latest Version (24.08) and Added load_7zip64

ThisNekoGuy:
`Updated load_7zip to install latest version and added load_7zip64`

- Brings the 32-bit version up to date
- Adds the 64-bit version as an installation option

### DIFF
--- a/src/winetricks
+++ b/src/winetricks
@@ -15726,17 +15726,31 @@ load_3m_library()
 
 #----------------------------------------------------------------
 
-w_metadata 7zip apps \
-    title="7-Zip 24.09" \
-    publisher="Igor Pavlov" \
-    year="2024" \
-    media="download" \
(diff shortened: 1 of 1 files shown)


austin987:
A) Please rebase to latest (7-zip was recently updated to 24.09)
B) I don't think a separate verb here makes sense. Instead, it should detect if the WINEPREFIX is 64-bit, and install the 64-bit version instead of the 32-bit one.

ThisNekoGuy:
Accidentally closed this, but I'm trying to address the issues you mentioned @austin987

ThisNekoGuy:
I actually came across this a few moments ago :sweat_smile: 
Originally thought there wasn't a variable for it and thought of using grep, then did some thinking afterwards and found it did actually exist after some searching

PR #2303, opened 2024-11-20
IE8 page loading is slow

zmtong1988:
native wininet , ie8 page loading is slow. wininet Change to builtin.


austin987:
Thanks for the patch, but I'm not sure it makes sense. While improving ie8 performance is a nice benefit, I suspect that most people installing native ie8 are doing so to help with other applications, and using builtin wininet wouldn't make sense there.

Instead, I would suggest adding a w_warn message that the user may want to use builtin wininet for better performance.

PR #2307, opened 2024-12-01
Dxvk cleanup

Blisto91:
This PR does 3 things

1. Removes any pre dxvk 1.0 versions
It is unlikely that a regular user would ever need to use the very original pre release versions and so i thought it would make a nice cleanup to remove them.

2. Removes the installation of d3d10.dll and d3d10_1.dll
Even before their removal in dxvk 2.0 they were not installed by default in the old dxvk install script since version 1.5.5.
Their implementations were not complete and it was basically nearly always recommended to use either the Wine or Windows implementations. Using the dxvk ones could break some games.
The actual d3d10 implementation is in d3d10core while the two above are just helper libraries.

3. Removed the automatic installation of the d3dx9 and d3dcompiler libraries
Dxvk does not have a direct dependency on these and so doing a implicit install of them doesn't seem quite right and it has lead to confusion in the past. Bug reports have been made where a user think there was a dxvk regression because installing a older dxvk via winetricks fixed a bug even though it was because of the verb installing these libraies.
Imo it is better to just install these explicitly via their own verbs and not as a side effect of the dxvk one.


austin987:
Please rebase on latest git.

austin987:
Thanks!

PR #2310, opened 2024-12-17
Updated 7zip to 24.09 and added 7zip arch bits conditional installation

ThisNekoGuy:
Updated 7zip to 24.09 and added arch bits conditional installation for win32 and win64

Better implementation of my attempt of https://github.com/Winetricks/winetricks/pull/2302 that doesn't require a new verb.

### DIFF
--- a/src/winetricks
+++ b/src/winetricks
@@ -15726,33 +15726,25 @@ load_3m_library()
 
 #----------------------------------------------------------------
 
-if [ "${W_ARCH}" = "win32" ]; then
-    w_metadata 7zip apps \
-        title="7-Zip 24.09" \
-        publisher="Igor Pavlov" \
-        year="2024" \
-        media="download" \
-        file1="7z2409.exe" \
-        installed_exe1="${W_PROGRAMS_X86_WIN}/7-Zip/7zFM.exe"
-elif [ "${W_ARCH}" = "win64" ]; then
-    w_metadata 7zip apps \
-        title="7-Zip 24.09" \
-        publisher="Igor Pavlov" \
-        year="2024" \
-        media="download" \
-        file1="7z2409-x64.exe" \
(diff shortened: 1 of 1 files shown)


ThisNekoGuy:
Huh?? Shellcheck didn't say anything about an issue on my system-- Oh, I see...
Weird that variables in the URL are a problem... I'll fix that.

src/winetricks
```diff
@@ -15726,17 +15726,31 @@ load_3m_library()
...
 #----------------------------------------------------------------
 
-w_metadata 7zip apps \
-    title="7-Zip 24.09" \
-    publisher="Igor Pavlov" \
-    year="2024" \
-    media="download" \
-    file1="7z2408.exe" \
-    installed_exe1="${W_PROGRAMS_X86_WIN}/7-Zip/7zFM.exe"
+if [ "${W_ARCH}" = "win32" ]; then
```
austin987, line 15729: w_metadata should be the same for both architectures.

ThisNekoGuy: The problem with that is leaving it the same would create the issue of placing the 64-bit version in the 32-bit programs folder...
That's why I separated the metadata conditionally per architecture with otherwise minimal changes besides also targeting the correct executable per architecture.

Functionally, this would never get in the way of normal usage because winetricks only ever gets run against one prefix per invocation...
I know it's not as clean as just having __**one**__ `w_metadata`, but I don't know well enough how `installed_exe2` would get treated when attempting installation. :/

austin987: Yes, see `winetricks_set_wineprefix()`

ThisNekoGuy: Looks a bit vaguely defined to me when both of them are set to this:
`W_PROGRAMS_WIN="$(w_expand_env ProgramFiles)"`
with no clear distinction even if I follow where `w_expand_env()` is defined, but I'll take your word for it

PR #2318, opened 2025-01-12
Allow {WINE,WINESERVER}_BIN env vars

Matoking:
Make it possible to set `WINE_BIN` and `WINESERVER_BIN` using environment variables outside Winetricks.

Winetricks performs a Wine binary bitness detection to print a warning. In some environments, such as Protontricks and Nix, the Wine executable is a wrapper that calls a binary that's located elsewhere. In those situations we want to provide the path to the binary that's only used for the check to avoid printing unnecessary warnings.

Closes: #2183


austin987:
Thanks!

Issue #2319, opened 2025-01-13
winetricks does not support environment variable WINEPREFIX for installing dll overrides in user-defined wine prefixes


rhabacker:
With actual `winetricks`, when installing a dll in a custom wine prefix by using `WINEPREFIX=<custom-wine-prefix> /usr/bin/winetricks ole32`, I got a message box 

`regedit: The file `C:\Windows\Temp\override-dll.reg` not found
`

and in the command line i see

```
Using native,builtin override for following DLLs: ole32
++ WINEPREFIX=/home/<user>/.wine64
++ /usr/bin/wine64 'C:\windows\regedit.exe' 'C:\windows\Temp\override-dll.reg'
```

which means that the specified wine prefix has been ignored.

```
/usr/bin/winetricks --version
Executing cd /usr/bin
20250102 - sha256sum: 657e0c4260424723476aea7668414a10ed859baf4660f96626694df9b2398460
```

rhabacker:
A subsequent analysis showed that this is not a problem of winetricks, but that a wrapper with the name `wine64` was included in a local search path that specified the wine prefix.

Due to the error output
```
++ WINEPREFIX=/home/<user>/.wine64
++ /usr/bin/wine64 'C:\windows\regedit.exe' 'C:\windows\Temp\override-dll.reg'
```
was unfortunately not visible that an executable file with the name `wine` was actually called from a different path, which could perhaps be improved.

rhabacker:
> ++ /usr/bin/wine64 'C:\windows\regedit.exe' 'C:\windows\Temp\override-dll.reg'

Starting with Wine 11, this loader has been removed https://gitlab.winehq.org/wine/wine/-/releases/wine-11.0, so this issue may no longer occur.

PR #2320, opened 2025-01-13
Setup internal used environment variable WINE with absolute wine path

rhabacker:
In the error or verbosity case this helps to see the actual wine executable used.

Closes: #2319


jre-wine (commented):
I haven't really tested it, sorry, but if this does what you need it for, it sounds good. Just two thoughts:

Somewhere else in the code there's already
`WINE_BIN="$(command -v "${WINE}")"`
Doing the same thing twice seems unnecessary, could this be simplified?


AFAICT this  is unnecessary or wrong here:
`&& test -f "${_abswine}"`

To avoid the unnecessary -f use
`if test -x "${_abswine}" ; then`

Or if there is a legit case where WINE is not executable but is a file use
`if test -x "${_abswine}" || test -f "${_abswine}"; then`

src/winetricks.1
```diff
@@ -159,7 +159,7 @@ winetricks 7zip
 .SH ENVIRONMENT VARIABLES
 Wine checks several environment variables on startup:
 .TP
-.I WINE
+.I WINE, WINE_BIN
```
austin987, line 162: The manual should explain what each of them does / when to use them / etc.

Issue #2172, opened 2024-01-15
64-bit warning on Winetricks using -q


melroy89:
You will keep seeing the warning when you use `--gui` parameter together with the `-q` parameter on 64-bit bottles:

![image](https://github.com/Winetricks/winetricks/assets/628926/ee6b1a95-8bc2-4a39-b2b9-3126d85b9183)

Despite you saying it's already behind the `-q` flag: https://github.com/Winetricks/winetricks/issues/1730#issuecomment-1891612104. This is not true, since when I use winetricks with both `-q` and `--gui` I will get this warning pop-up.

Version: 20240105-next - sha256sum: 73f475faf6c041a9b04e033d4db74b8cc5fae20ddc8229ab494f8f938f96e73f

t0stiman:
Did you find a solution for this?

austin987:
It doesn't HAVE to be solved by me, someone else could send a pull request for review ;)

melroy89:
> It doesn't HAVE to be solved by me, someone else could send a pull request for review ;)

True statement!

melroy89:
Could you point me to the code line where this pop-up is triggered? I'm trying to search on (part of) the error message string. But unable to get a result.

mirh:
`W_NO_WIN64_WARNINGS` is around line 4550 atm.

melroy89:
ow I see thanks. It seems that GitHub doesn't index this repo correctly/fully: https://github.com/search?q=repo%3AWinetricks%2Fwinetricks%20W_NO_WIN64_WARNINGS&type=code

melroy89:
Fixed: https://github.com/Winetricks/winetricks/pull/2334

PR #2334, opened 2025-01-29
Do not show Win64 warnings anymore in unattended mode. Eg. -q

melroy89:
- Stop showing this annoying Windows 64-bit bottle warning, when using a 64-bit bottle in unattended mode (eg. `-q` flag is set).

Fixes: #2172

### DIFF
--- a/src/winetricks
+++ b/src/winetricks
@@ -4549,6 +4549,11 @@ winetricks_set_wineprefix()
             else
                 W_NO_WIN64_WARNINGS=0
             fi
+
+            # In unattended mode, do NOT show any Win64 warnings
+            if [ -n "${W_OPT_UNATTENDED}" ]; then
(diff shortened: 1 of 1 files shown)


melroy89:
It seems to work a bit too good. It also not show the warning message now, even if `-q` is not provided. That is strange.

melroy89:
Works as expected now. You still get the warning with `--gui`. But not if you also set unattended: `--gui -q`

fabiojb:
Tested this and I can confirm that work as expected

melroy89:
I was also hoping this was merged by now..

miniluz:
Would be very useful!

melroy89:
I still think its useful to merge this PR.

melroy89:
Thanks!

PR #2358, opened 2025-03-30
Fixes VC2022 SHA mismatch

KotaruS:
Just a simple change to fix error due to SHA256 mismatch. Thanks microsoft.


QuintenQVD0:
@KotaruS Please update those already don't match anymore

austin987:
Thanks!

PR #2359, opened 2025-04-03
Fix sha mismatch

SurenIV:
Fix sha mismatch for:
adobe_diged4
dotnet11sp1
dxsdk_jun2010
nook
procexp
sketchup
steam
utorrent3
vcrun2022
vstools2019


austin987:
Thanks!

PR #2369, opened 2025-04-27
Added ImgBurn as Installable App Verb

ThisNekoGuy:
As of Wine Vanilla 10.5, ImgBurn itself doesn't seem to currently work anymore but the installer does. For user convenience, this commit adds the `imgburn` app verb for installation.


austin987:
I don't think this belongs in winetricks (it's not something that a lot of Wine users would need, particularly if it's broken).

Issue #2361, opened 2025-04-14
New verbs suggestion


PlayNeth:
Hello, I'm wondering if it's possible to include [dinputto8](https://github.com/elishacloud/dinputto8) and [DSOAL](https://github.com/ThreeDeeJay/dsoal) as new verbs, as they have been super useful to me on Linux and probably will for other people as well. 

Former translates dinput to dinput8 and improves gamepad/controller support, some older games I've got won't detect my dualshock 4 at all without it. Latter emulates EAX support in older games and potentially higher quality playback in general in games that use directsound. Thanks.

danielfikko:
# DSOAL

I tried to make a verb for DSOAL, but I don't have enough shell scripting skills and knowledge of how winetricks works, so I gave up. However, I'll leave the first stage here, with some additional comments that I discovered while researching, in case anyone wants to give this verb a go:

```sh
# Download the latest release
curl -L --output-dir ~/Downloads -O https://github.com/kcat/dsoal/releases/download/latest-master/DSOAL.zip

# Extract the release
7z x -oDownloads ~/Downloads/DSOAL.zip

# Extract again
7z x -oDownloads ~/Downloads/DSOAL_r658.zip

# Copy the dsoal-aldrv.dll and dsound.dll 32-bit DLLs to syswow64
cp ~/Downloads/DSOAL/Win32/{dsoal-aldrv,dsound}.dll $WINEPREFIX/drive_c/windows/syswow64

# Copy the dsoal-aldrv.dll and dsound.dll 64-bit DLLs to system32
cp ~/Downloads/DSOAL/Win64/{dsoal-aldrv,dsound}.dll $WINEPREFIX/drive_c/windows/system32

# Override dsoal-aldrv to use the native DLL
wine reg add "HKCU\Software\Wine\DllOverrides" /v "dsoal-aldrv" /d "native" /f

# Override dsound to use the native DLL
wine reg add "HKCU\Software\Wine\DllOverrides" /v "dsound" /d "native" /f
```

[...]

PR #2389, opened 2025-06-13
dsoal: New verb

danielfrrrr:
Add a new verb for https://github.com/kcat/dsoal. ~~It's a draft because...~~

1. ~~Is it safe and can I use the `*` in `file2="DSOAL_r*.zip"`? I did this because after extracting `DSOAL.zip`, you get a new file to extract: `DSOAL_r658.zip`. This time the file name contains the version of the DLL. The version number is the repository's commit count. If the `*` is not used, the verb will need to be edited manually with each new version, which breaks the flexibility a little.~~
1.1. **This is no longer necessary, since there was a change in the upstream that removed this second file: https://github.com/kcat/dsoal/pull/136!**
2. Including the list of installed files and `winetricks` storing `DSOAL.zip` in `~/.cache/winetricks/$W_PACKAGE` makes it impossible to re-install a more up-to-date version of the DLL. That's because `DSOAL.zip` doesn't change its name, it's a static name that gives you the latest version, ~~the name only changes the second time you extract it. There's also https://github.com/kcat/dsoal/releases/tag/archive, which deletes the `DSOAL.zip` file and gives you `DSOAL_r658.zip` (or the latest release) file directly. You'll have the same problem as with 1, which is having to manually edit the verb with each new release (or add some helper that gets the commit count from the repository)~~. I don't know what to do here.
2.2. **The strikethrough is no longer relevant since https://github.com/kcat/dsoal/pull/136, but it's still a problem. Maybe a warning saying to delete the cache if you want a more up-to-date version?**
3. ~~This DLL comes in two versions: `DSOAL` and `DSOAL+HRTF`. The latter comes with a configuration file that enables HRTF: `alsoft.ini`. This configuration file must be installed in `%APPDATA%/alsoft.ini`. If you install the version with HRTF disabled, you can later enable it with this configuration file and vice versa. Anyway, I have two questions: which version should be installed by default and should this configuration file be installed in both cases?~~
3.3. **Currently this verb installs DSOAL without HRTF, I'll keep it that way. HRTF can be enabled by copying the `./DSOAL+HRTF/Win32/alsoft.ini` file to `$WINEPREFIX/drive_c/users/$USERNAME/AppData/Roaming/alsoft.ini`.**
4. ~~I don't know if I'm installing the 32-bit and 64-bit DLLs correctly. I feel that you need to respect whether the Wine prefix is fully 32-bit.~~
4.4. **This has already been fixed.**

I know little about shell scripting and how `winetricks` works, help here would really be appreciated (or maybe someone else should take responsibility for that verb)!

Partially closes: https://github.com/Winetricks/winetricks/issues/2361


danielfikko:
https://github.com/kcat/dsoal/pull/136 was a change that made things a lot easier, I modified the verb to match this change.

Only option 2 remains a problem. Is it acceptable to make a warning saying to delete the cache if you want a newer version? Apart from that, I believe everything is ready.

danielfikko:
Hi, @austin987. Could you take a look at this PR and https://github.com/Winetricks/winetricks/pull/2390?

austin987:
> Only option 2 remains a problem. Is it acceptable to make a warning saying to delete the cache if you want a newer version? 

Sure.

danielfikko:
I removed the redundant arguments and added a warning!

Well, I completely messed up the commits due to my lack of experience with git. If needed, I can open another PR.

mirh:
Dsoal should also include the registry changes from https://github.com/ThreeDeeJay/DSWRP

PR #2394, opened 2025-06-19
winmetadata: download from windows-rs github

ElementalWarrior:


ElementalWarrior:
The wine devs didn't have an appetite to support this. I have another idea for a verb though. Standby :)

PR #2395, opened 2025-06-20
affinity: new verb

ElementalWarrior:


austin987:
I don't know what affinity is. What is it / why should it be in winetricks?

ElementalWarrior:
> I don't know what affinity is. What is it / why should it be in winetricks?

Affinity is the name for the software suite by Serif of their line of graphics applications.

To run it, it requires winrt functionality. Which is what the verb is adding. And a DLL to load the windows-rs packaged Windows.winmd file.

The winmd file contains the typing information used by winrt/dotnet applications.

austin987:
I don't think this belongs in winetricks, as this is pretty application specific. It would be better to put this on the AppDB for the application.

PR #2403, opened 2025-07-21
New verb: msls31_x64

eskay993:
Added a new verb `msls31_x64` which installs both a 32-bit and 64-bit version of the dll as needed, depending on the arch. The existing `msls31` verb only installs a 32-bit version, which causes a crash with 64-bit apps using rich text with `msftedit` and extended Unicode characters. 

The official IE8 installer has the needed dlls, so this adapts the `ie8` verb and strips it down to just extracting both versions of msls31.dll.

I created a new verb since `msls31` is called by other functions, so did not want to introduce anything unexpected. The version of the dll included with the IE8 installer is slightly newer than the one grabbed by `msls31`.

Also updated the title of `msls31` to make it clear it's 32-bit only.

Thank you.

### DIFF
--- a/src/winetricks
+++ b/src/winetricks
@@ -11092,6 +11092,32 @@ load_msls31()
 
 #----------------------------------------------------------------
 
+w_metadata msls31_x64 dlls \
+    title="MS Line Services (both 32-bit and 64-bit versions)" \
+    publisher="Microsoft" \
+    year="2001" \
+    media="download" \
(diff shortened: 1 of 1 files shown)


eskay993:
Apologies, just noticed an issue with the cache download path. It's still using ` "${W_CACHE}"/ie8` instead of `"${W_CACHE}"/msls31_x64`. It was still working for me since I have the ie8 download cached. Will change and resubmit.

PR #2404, opened 2025-07-22
Add verb: msls31_x64

eskay993:
Added a new verb `msls31_x64` which installs both a 32-bit and 64-bit version of the dll as needed, depending on the arch. The existing `msls31` verb only installs a 32-bit version, which causes a crash with 64-bit apps using rich text with `msftedit` and extended Unicode characters.

The official IE8 installer has the needed dlls, so this PR adapts the `ie8` verb and strips it down to just handling both versions of msls31.dll.

I created a new verb since `msls31` is called by other functions, so did not want to introduce anything unexpected. The version of the dll included with the IE8 installer is slightly newer than the one grabbed by `msls31`.

Also updated the title of `msls31` to make it clear it's 32-bit only.

Thank you.

### DIFF
--- a/src/winetricks
+++ b/src/winetricks
@@ -11072,7 +11072,7 @@ load_msscript()
 #----------------------------------------------------------------
 
 w_metadata msls31 dlls \
-    title="MS Line Services" \
+    title="MS Line Services (32-bit only)" \
     publisher="Microsoft" \
     year="2001" \
     media="download" \
--- a/src/winetricks
+++ b/src/winetricks
@@ -11092,6 +11092,32 @@ load_msls31()
 
 #----------------------------------------------------------------
 
+w_metadata msls31_x64 dlls \
+    title="MS Line Services (both 32-bit and 64-bit versions)" \
+    publisher="Microsoft" \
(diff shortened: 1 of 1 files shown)


austin987:
Thanks for the PR. I think in this case it makes sense to rename the old verb and have the new one be the default. I'd suggest `msls_nt4` for the rename.

eskay993:
Sorry for the late reply. Sounds good, I'll switch the names round and send an update shortly. Many thanks.

austin987:
Please fix the shellcheck issues (and preferably, squash the commits).

Something else I realized: existing references to msls31 should be checked/possibly changed to use the nt4 version. I can do that after merging this if you can't.

eskay993:
No worries. Tidied up trailing spaces and squashed to one commit.

Checking others that call the old verb is a good shout. One of them is being ie8 itself. From my early testing, after it calls the old verb, it then overwrites msls31.dll with one from the ie8 package anyway.  Not sure if this is a redundant call and not needed, or if there is a reason for it. 

The other two are riched20 and ie6 which I haven't played around with.

PR #2418, opened 2025-08-26
xact: Don't try to register files that are not DLLs

alexhenrie:
If the symbols have been separated from the libraries (for example, if the Wine bottle is in a Proton build that was compiled with `make install UNSTRIPPED_BUILD=1`), there will be a .dll.debug file next to each .dll file, and running regsvr32.exe on those files will fail.


alexhenrie:
Thanks Austin!

PR #2417, opened 2025-08-25
Addition of the French language

alrito28:
I'm working on it, wait before merging it.

PR #2421, opened 2025-08-28
Create manual (command man) in French

alrito28:


austin987:
There appears to be overlap with #2417 (and conflicts as well). It would be best to keep this PR to just the manpage changes.

alrito28:
Hi,

I think that's fine. The conflicts are mainly due to overlapping branches. There's one that does both. Try that, and if not, I'll do two pulls again, but this time on the master branch.

alrito28:
I think you can go ahead, it'll work out.

austin987:
Changes with conflicts can't be merged. You'll need to resolve it.

alrito28:
The changes have been made, and if there is a conflict, then we will modify it, but for me, it's a correction.
If you have any doubts, the code works perfectly with French and other translation points, particularly in the command line.

:)

Issue #2433, opened 2025-10-13
`winetricks --help` broken since commit `65c4a139b3abc92fcc5bf96d3702fb787d5deb78`


R1kaB3rN:
### Summary

A recent change introduced in commit 65c4a139b3abc92fcc5bf96d3702fb787d5deb78 appears to have broken `winetricks --help`. Attempting to run `winetricks --help` currently results in output:
```
Executing cd ./src
```

### Steps to reproduce

1. `git clone https://github.com/Winetricks/winetricks.git`
2. `cd winetricks`
3. `git checkout 65c4a139b3abc92fcc5bf96d3702fb787d5deb78`
4. `./src/winetricks --help`

### Expected result

```
Usage: ./src/winetricks [options] [command|verb|path-to-verb] ...
Executes given verbs.  Each verb installs an application or changes a setting.

Options:
    --country=CC      Set country code to CC and don't detect your IP address
-f, --force           Don't check whether packages were already installed
    --gui             Show gui diagnostics even when driven by commandline
    --gui=OPT         Set OPT to kdialog or zenity to override GUI engine
    --isolate         Install each app or game in its own bottle (WINEPREFIX)
    --self-update     Update this application to the last version
    --update-rollback Rollback the last self update
    --no-clean        Don't delete temp directories (useful during debugging)
    --optin           Opt in to reporting which verbs you use to the Winetricks maintainers
    --optout          Opt out of reporting which verbs you use to the Winetricks maintainers
-q, --unattended      Don't ask any questions, just install automatically
-t  --torify          Run downloads under torify, if available
    --verify          Run (automated) GUI tests for verbs, if available
-v, --verbose         Echo all commands as they are executed
-h, --help            Display this message and exit
-V, --version         Display version and exit

Commands:
list                  list categories
list-all              list all categories and their verbs
apps list             list verbs in category 'applications'
benchmarks list       list verbs in category 'benchmarks'
dlls list             list verbs in category 'dlls'
fonts list            list verbs in category 'fonts'
settings list         list verbs in category 'settings'
list-cached           list cached-and-ready-to-install verbs
list-download         list verbs which download automatically
list-manual-download  list verbs which download with some help from the user
list-installed        list already-installed verbs
arch=32|64            create wineprefix with 32 or 64 bit, this option must be
                      given before prefix=foobar and will not work in case of
                      the default wineprefix.
prefix=foobar         select WINEPREFIX=/home/foo/.local/share/wineprefixes/foobar
[...]

PR #2434, opened 2025-10-14
fix: temporarily revert translation patchset to list --help output

R1kaB3rN:
Closes https://github.com/Winetricks/winetricks/issues/2433. Reverts 1e0bf92...7511823 until the OP fixes the bug introduced in the Simplified Chinese patchset.


Chiitoo:
I had noticed the eluside '--help', but didn't get to looking into it, and quite forgot about it... apologies for that!

The following change fixes things for me as far as I can tell:

https://github.com/Winetricks/winetricks/pull/2438

PR #2437, opened 2025-10-20
dsound: Allow builtin fallback for dsound.dll

alexhenrie:
The dsound verb only installs 32-bit dsound, which means that setting dsound.dll to native-only breaks 64-bit dsound.


alexhenrie:
Thank you!

PR #2438, opened 2025-10-20
winetricks_usage: add a missing '_EOF_'

Chiitoo:
This fixes regression introduced in 65c4a139b3a [1], which causes 'winetricks --help' to only print 'Executing cd /usr/bin' in any language other than 'zh_CN', for which the English usage would also be printed.

1. https://github.com/Winetricks/winetricks/commit/65c4a139b3a

Closes: https://github.com/Winetricks/winetricks/pull/2434


austin987:
Thanks!

PR #2457, opened 2026-01-02
Do not give a warning about unknown file arch (when -q is set)

melroy89:
Under Debian Trixie (just an example), I still get "Unknown file arch for /usr/bin/wine" despite the `-q` option (`--unattended`).

image

This is **very** **annoying**, especially since I'm using winetricks with the `-q` option. This PR should fix that.


melroy89:
Thanks!

Issue #2464, opened 2026-01-22
vcrun2022 - cabextract doesn't do what's intended with multi-arch installers


gbdrbob:
Hi

The vcrun2022 verb attempts to work around [wine bug 57518](https://bugs.winehq.org/show_bug.cgi?id=57518) by directly extracting msvcp140.dll from the vc_redist.x86.exe ([see here](https://github.com/Winetricks/winetricks/blob/81a692e2ee6b311367f18a17a9a61a0ef1c0ad18/src/winetricks#L13615-L13618)) and vc_redist.x64.exe ([and here](https://github.com/Winetricks/winetricks/blob/81a692e2ee6b311367f18a17a9a61a0ef1c0ad18/src/winetricks#L13641-L13643)) packages.

However the filenames in the both the x86 and x64 packages currently supplied by MS are suffixed with the intended architechture (_x86 or _amd64), so the filter currently used in winetricks to extract the dll does not match on the file:

```
$ sha256sum vc_redist.x86.exe
0c09f2611660441084ce0df425c51c11e147e6447963c3690f97e0b25c55ed64  vc_redist.x86.exe
$ cabextract vc_redist.x86.exe -F 'a10'
Extracting cabinet: vc_redist.x86.exe
  extracting a10

All done, no errors.
$ cabextract -l a10
a10: WARNING; possible 10368 extra bytes at end of file.
Viewing cabinet: a10
 File size | Date       Time     | Name
-----------+---------------------+-------------
    259688 | 11.06.2025 05:25:36 | concrt140.dll_x86
    447616 | 11.06.2025 05:25:36 | msvcp140.dll_x86
     33896 | 11.06.2025 05:25:36 | msvcp140_1.dll_x86
    294992 | 11.06.2025 05:25:36 | msvcp140_2.dll_x86
     47240 | 11.06.2025 05:25:36 | msvcp140_atomic_wait.dll_x86
     30824 | 11.06.2025 05:25:38 | msvcp140_codecvt_ids.dll_x86
    383616 | 11.06.2025 05:25:38 | vcamp140.dll_x86
    287360 | 11.06.2025 05:25:38 | vccorlib140.dll_x86
    163488 | 11.06.2025 05:25:38 | vcomp140.dll_x86
     91216 | 11.06.2025 05:25:38 | vcruntime140.dll_x86
     36000 | 11.06.2025 05:25:38 | vcruntime140_threads.dll_x86

All done, no errors.
[...]

PR #2465, opened 2026-01-22
Fix extraction of msvcp140.dll in vcrun2022

gbdrbob:
This pull is intened to fix the bug described in issue #2464 - where msvcp140.dll is not extracted from vc_redist.x86.exe and vc_redist.x64.exe due to suffixes on the filenames in the archive.


src/winetricks
```diff
@@ -13640,7 +13641,8 @@ load_vcrun2022()
 
             # Also replace 64-bit msvcp140.dll
             w_try_cabextract --directory="${W_TMP}/win64"  "${W_CACHE}"/"${W_PACKAGE}"/vc_redist.x64.exe -F 'a12'
-            w_try_cabextract --directory="${W_SYSTEM64_DLLS}" "${W_TMP}/win64/a12" -F 'msvcp140.dll'
+            w_try_cabextract --directory="${W_TMP}/win64" "${W_TMP}/win64/a12" -F 'msvcp140.dll_amd64'
+            w_try mv "${W_TMP}/win64/msvcp140.dll_amd64" "${W_SYSTEM64_DLLS}/msvcp140.dll"
```
Gcenx: Same thing here

Issue #2436, opened 2025-10-15
Verb request [webview2]


Twig6943:
Download link: https://developer.microsoft.com/en-us/microsoft-edge/webview2

marioroy:
Steps for Microsoft Edge WebView2 Runtime installation

https://github.com/ryzendew/AffinityOnLinux/issues/29

Twig6943:
> Steps for Microsoft Edge WebView2 Runtime installation
> 
> [ryzendew/AffinityOnLinux#29](https://github.com/ryzendew/AffinityOnLinux/issues/29)

I already knew that.

Issue #2226, opened 2024-05-21
Verb request: Microsoft Edge WebView2 Runtime


qwertychouskie:
Download: https://go.microsoft.com/fwlink/p/?LinkId=2124703

Apps like the Adobe Creative Cloud installer will use WebView2 if available instead of mshtml.dll.  There is one workaround needed though, `msedgewebview2.exe` needs to be set to Windows 7 for apps using WebView2 to work (at least the Creative Cloud installer anyways).

st-jude:
Thanks for this trick! I was having UI issues with Universal Audio vst plugs and this did fix the issue, unfortunately creating another one. The UI now finally appears and am able to move around the knobs etc but now once i try to remove the plugin or close it my DAW crashes. This happened both in Ardour and Reaper.

austin987:
@qwertychouskie were you testing this in a 32 or 64 bit prefix? What wine version?

For me, it installs fine in a 32bit prefix, but on 64bit there's a crash. Setting the windows version to win7 didn't help.

Didn't try using it with a real app yet..

zmtong1988:
Can it be installed successfully?

flleeppyy:
> Can it be installed successfully?

Wine 9.22 staging
```
export WINEARCH=win32
export WINEPREFIX=~/.wineprefixes/webview2test
wine wineboot
winetricks vcrun2017 corefonts
wine ~/Downloads/MicrosoftEdgeWebView2RuntimeInstallerX86.exe # Latest x86 from ms website, but try https://web.archive.org/web/20241127085924/https://msedge.sf.dl.delivery.mp.microsoft.com/filestreamingservice/files/c234a7e5-8ebb-49dc-b21c-880622eb365b/MicrosoftEdgeWebView2RuntimeInstallerX86.exe
winetricks --force vcrun2015 # ctrl-c it when it stucks at waiting for wine server. Its a required step. Yes, really
winecfg -v win10
```
I was able to get webview2 installed with a fresh wineprefix, opening edge works pretty okay, it chugs, but it flashes a LOT. As for software that uses Webview2, I have no clue, I only have one program build with webview2 but it's Byond 516 which is a private beta, so, not entirely allowed to show screenshots.


Uploading simplescreenrecorder-2024-11-27_01.16.00.mp4…

Issue #1986, opened 2022-11-27
Easy install of webview2


qarmin:
Some apps require webview2 and I found that there is no way to do it via winetricks

Download link - https://developer.microsoft.com/en-us/microsoft-edge/webview2/#download-section

austin987:
Do you have a sample application that needs this? Even better if you have a sample application to test and a PR to add the dll ;)

Gubbins:
I was able to get a certain way through the installation process by using winetricks to install .Net 472, but webview2 installation still failed.

The application I'm attempting to run is Zwift (free download and the issue can be reproduced without needing to open an account). The first thing it does when launched is use webview2 to open a login window. Sorry it's not a simple sample application.

Barracuda72:
Another example of such an app is a launcher of an upcoming game "Snowbreak: Containment Zone":
https://snowbreak-content.amazingseasuncdn.com/ob202307/launcher/seasun/setup/Snow_Setup.exe

In fact, many new games (and their launchers) require WebView2. WineDB lists it as a "Platinum" for some reason, but I wasn't able to even install it, sadly.

Barracuda72:
I was able to install version 91.0.864.59 from the Web Archive:

https://web.archive.org/web/20210626091814/https://msedge.sf.dl.delivery.mp.microsoft.com/filestreamingservice/files/ee4e97c1-89a3-456f-b9f3-f29651316b7e/MicrosoftEdgeWebView2RuntimeInstallerX64.exe

However, it's too old to run the game in question. I wasn't able to find a newer build besides the latest one (114.x.x.x), which doesn't work.

I've tried to substitute files from "FixedBuild" version archive and the latest version I was able to use was 105.0.1343.25 (https://msedge.sf.dl.delivery.mp.microsoft.com/filestreamingservice/files/fbf928f5-dd43-4f7f-94ed-75cc22fc8141/Microsoft.WebView2.FixedVersionRuntime.105.0.1343.25.x64.cab). Newer ones either crash the game instantly or don't show the window displaying in the console "WebVie2w process crashed". This approach requires modifying a registry but allows the game launcher to run. 

Sadly, only sound works, nothing is dispalyed...

markus1189:
@Barracuda72 thank you for the tip from the Web Archive!  This finally allowed me to find an installable version of Edge WebView2 and run Zwift :tada:

Nick768:
[...]

PR #2467, opened 2026-01-24
webview2: New verb

qwertychouskie:
Closes https://github.com/Winetricks/winetricks/issues/2436
Closes https://github.com/Winetricks/winetricks/issues/2226
Closes https://github.com/Winetricks/winetricks/issues/1986

~~Can `installed_file1` be a directory rather than a file?  Would be nice to avoid the workaround.~~ DONE: Checking for `${W_PROGRAMS_X86_WIN}/Microsoft/EdgeUpdate/MicrosoftEdgeUpdate.exe` now.

~~Also, I'd like there to be some sort of workaround for https://bugs.winehq.org/show_bug.cgi?id=53925 so the prefix doesn't stay running indefinitely, as that can cause hiccups when combining with other verbs that expect the prefix to auto-shutdown when done running.  I think `MicrosoftEdgeUpdate.exe` is registered as a service, is there a way to disable the service?~~ DONE

### DIFF
--- a/src/winetricks
+++ b/src/winetricks
@@ -13767,6 +13767,48 @@ load_webio()
     w_override_dlls native,builtin webio
 }
 
+#----------------------------------------------------------------
+
+w_metadata webview2 dlls \
+    title="Microsoft Edge WebView2 Evergreen Runtime" \
+    publisher="Microsoft" \
+    year="2020" \
+    media="download" \
+    file1="MicrosoftEdgeWebview2Setup.exe" \
+    installed_file1="${W_PROGRAMS_X86_WIN}/Microsoft/EdgeUpdate/MicrosoftEdgeUpdate.exe"
+
+load_webview2()
+{
+    # https://developer.microsoft.com/en-us/microsoft-edge/webview2/
+    # Evergreen Bootstrapper - downloads and installs appropriate architecture
+    # https://go.microsoft.com/fwlink/p/?LinkId=2124703
+    # Skip checksum verification since this file updates frequently
+    w_download https://go.microsoft.com/fwlink/p/?LinkId=2124703 "" MicrosoftEdgeWebview2Setup.exe
+
+    w_try_cd "${W_CACHE}"/"${W_PACKAGE}"
+    w_try_ms_installer "${WINE}" MicrosoftEdgeWebview2Setup.exe ${W_OPT_UNATTENDED:+/silent /install}
+
+    if w_workaround_wine_bug 53925 "Setting edgeupdate service to manual startup"; then
+        cat > "${W_TMP}"/edgeupdate-service.reg <<_EOF_
+REGEDIT4
+
+[HKEY_LOCAL_MACHINE\\System\\CurrentControlSet\\Services\\edgeupdate]
+"Start"=dword:00000003
+
+_EOF_
+        w_try_regedit "${W_TMP_WIN}"\\edgeupdate-service.reg
+
+        # Kill the running service since it was started during installation
+        # Use pgrep -f to match full command line since process name is >15 chars
+        # shellcheck disable=SC2046,SC2086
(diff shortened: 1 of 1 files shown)


qwertychouskie:
@austin987 I think this should be fully ready for review now.

src/winetricks
```diff
@@ -13767,6 +13767,48 @@ load_webio()
...
+    file1="MicrosoftEdgeWebview2Setup.exe" \
+    installed_file1="${W_PROGRAMS_X86_WIN}/Microsoft/EdgeUpdate/MicrosoftEdgeUpdate.exe"
+
+load_webview2()
+{
+    # https://developer.microsoft.com/en-us/microsoft-edge/webview2/
+    # Evergreen Bootstrapper - downloads and installs appropriate architecture
+    # https://go.microsoft.com/fwlink/p/?LinkId=2124703
+    # 2026/01/19: 6a2f2492ea642e3965ce2421e257da53ff18c5dee08df33c84c061ae47a8a302
+    w_download https://go.microsoft.com/fwlink/p/?LinkId=2124703 6a2f2492ea642e3965ce2421e257da53ff18c5dee08df33c84c061ae47a8a302 MicrosoftEdgeWebview2Setup.exe
```
austin987: I get d75e4fb20d4e8ab50ab77d43c139aa0f654ecca9c504552761efb435917a01e5, with both win32 and win64 prefixes? Did the sha change, or is it inconsistent? If it's inconsistent, then winetricks shouldn't check the value.

qwertychouskie: How do I omit the checksum, but keep the filename parameter?  Removing just the checksum makes Winetricks try to interpret the filename as the checksum, and omitting the filename makes the downloaded filename be `?LinkId=2124703`

h-banii:
I'm not sure if this is useful, but it might also be possible to use the fixed version runtime (.cab file) instead of the evergreen installer

with something like this:

```sh
w_metadata webview2 apps \
    title="Microsoft Edge WebView2" \
    publisher="Microsoft" \
    media="download" \
    file1="Microsoft.WebView2.FixedVersionRuntime.143.0.3650.139.x64.cab"

load_webview2()
{
    w_download "https://msedge.sf.dl.delivery.mp.microsoft.com/filestreamingservice/files/ea30811e-a216-4d55-89f3-c1099862c8fc/Microsoft.WebView2.FixedVersionRuntime.143.0.3650.139.x64.cab" 685b0eecc96c50eacd9384b25ba2b8aa5e7a302d914f5035af0c1e82bd9d334c
    w_try_cabextract -d "${W_PROGRAMS_UNIX}" "${W_CACHE}"/webview2/"${file1}"

    cat > "${W_TMP}"/set-webview2-path.reg <<_EOF_
REGEDIT4

[HKEY_LOCAL_MACHINE\SOFTWARE\Policies\Microsoft\Edge\WebView2\BrowserExecutableFolder]
"*"="${W_PROGRAMS_WIN//\\/\\\\}\\\\${file1%????}"

_EOF_

    w_try_regedit "${W_TMP}"/set-webview2-path.reg

    w_set_app_winver msedgewebview2.exe win7
}
```

I don't know if it works on *any* app, but it does work for clip studio paint

I couldn't find a more permanent download link though, I just used the one from here:
https://developer.microsoft.com/en-us/microsoft-edge/webview2/
(so it'll probably not work in a few months or so)

qwertychouskie:
I would expect apps to need updated features of webview2 over time.  Unless there is a known regression in newer versions of webview2, pinning to an older version is not the right play IMHO.

qwertychouskie:
@austin987 Any chance you can take a look at this and https://github.com/Winetricks/winetricks/pull/2466?  Currently I'm keeping a modified copy at https://github.com/qwertychouskie/AdobeOnLinux but I'd like to have these merged so I can open a draft PR for the adobe_cc (Adobe Creative Cloud) verb here.

PR #2502, opened 2026-04-09
Improve Chinese translation.

narukeu:
### 2. Resolved the inconsistent translation of the term `wineprefix`. This is a key change in this PR.

In the original translation, `wineprefix` was inconsistently rendered as "容器", "前缀", or simply left in English. I believe this inconsistency causes ambiguity.

- "Wine 前缀" (Wine Prefix): While this is a literal translation, I find it obscure and difficult for average users to grasp the specific nature of a wineprefix.
- "容器" (Container/Bottle) The term "Bottle" appears only once in the source text; however, `wineprefix` was translated as "容器" in several other places. I think this disrupts the original context and needs to be corrected.

Given the technical specificity of `wineprefix`, I have ultimately chosen to retain the original English term and treat it as a proper noun throughout the text.

### DIFF
--- a/src/winetricks
+++ b/src/winetricks
@@ -432,10 +432,10 @@ _w_get_broken_messages()
             broken_no_version_known_no_mingw="此软件包（${W_PACKAGE}）在 wine 未使用 mingw 构建时已损坏。更多信息请参见 ${bug_link}。如仍要尝试，请使用 --force。"
 
             # win64 broken messages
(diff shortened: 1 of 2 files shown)


narukeu:
@austin987 Any updates on this? Ready to adjust if needed.

austin987:
Thanks, and sorry for the delay!

PR #2508, opened 2026-07-18
Let wine64 choose the correct regedit.exe

KenSharp:
See https://bugs.winehq.org/show_bug.cgi?id=60032

wine64 knows which regedit.exe it needs

### DIFF
--- a/src/winetricks
+++ b/src/winetricks
@@ -951,8 +951,9 @@ w_try_regedit64()
         *) unset cmdc ;;
     esac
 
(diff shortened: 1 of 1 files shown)


austin987:
Upstream bug was closed invalid.

KenSharp:
The Wine bug is invalid, the winetricks one clearly isn't. It shouldn't be making assumptions about wine64.
https://github.com/Winetricks/winetricks/issues/2509

It was winetricks causing the issues in the Wine bug.

PR #2511, opened 2026-08-01
dpi: new verb

alexhenrie:
When running Wine on a high-DPI display, it is very difficult to navigate the tiny winecfg interface to increase the DPI to a reasonable value. Add a winetrick to set the DPI from the command line. The possible values are 96, 120, 144, 168, 192, 216, 240, 288, 336, 384, 432, and 480, which is the same set of values that winecfg allows.

### DIFF
--- a/src/winetricks
+++ b/src/winetricks
@@ -18407,6 +18407,77 @@ _EOF_
 }
 
 #----------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


alexhenrie:
Thank you!

PR #2513, opened 2026-08-10
Complete the Russian translation

wehrwolfmann:
Russian was already partially translated, but a number of user-visible strings had no `ru*` branch and fell back to English. This fills the remaining gaps so that a Russian user no longer sees a mix of two languages in the same session.

What is added:

* `w_askpermission` cancellation message ("Operation cancelled, quitting.")
* All 16 "package is broken" messages (default / mingw / no-mingw / win64)
* The two wow64 warnings printed at startup. These had no `case ${LANG}` block at all, so they were always English even for locales that had everything else translated; they are now wrapped the same way as the neighbouring 64-bit WINEPREFIX warning.
* The WPF infinite-loop warning in the ten dotnet verbs
* "Select \<prefix\>" entry of the zenity prefix menu
* The "create new wineprefix" dialog (title / name / architecture)
* The `annihilate` confirmation
* The full `--help` output
* `title_ru` for the 128 settings verbs that did not have one yet (four of them already did)

One existing Russian string was reworded: the 64-bit WINEPREFIX warning translated "verb" as "ветка" (branch), which does not mean anything in this context. It now uses the same wording as the rest of the Russian text.

Tested on wine-11.15 with new-style wow64, both zenity and kdialog frontends:

* `LANG=ru_RU.UTF-8 winetricks --help` and `--gui=zenity settings` render in Russian
* `LANG=en_US.UTF-8`, `de_DE.UTF-8` and `zh_CN.UTF-8` are unchanged
* the `tests/shell-checks` formatting rules pass and `shellcheck -s sh src/winetricks` is clean (shellcheck 0.11.0)


austin987:
Спасибо!

